PR #1, opened 2018-03-13
moved weeks 4,5,6 from the JavaScript repo into JavaScript2

mkruijt:


README.md
```diff
@@ -6,15 +6,15 @@ Here you can find course content and homework for the JavaScript2 module
...
-|2.|• Functions + JSON/Arrays<br>• [Array Manipulations](fundamentals/array_manipulation.md)<br>• JSON<br>• [Map and filter](fundamentals/map_filter.md)<br>• Arrow functions |[Reading Week 5](/Week2/README.md)|[Homework Week 5](/Week2/MAKEME.md)|
-|3.|• [Closures](fundamentals/scope_closures_this.md) <br>• Callbacks|[Reading Week 6](/Week3/README.md)|[Homework Week 6](/Week3/MAKEME.md)|
+|1.|• Capturing user input <br>• Events<br>• [Basic DOM manipulations (img src, innerHTML)](../../../fundamentals/blob/master/fundamentals/DOM_manipulation.md)<br>• Code debugging using the browser <br>• [Code commenting](../../../fundamentals/blob/master/fundamentals/code_commenting.md)<br>• Structuring code files<br>• [Code formatting](../../../fundamentals/blob/master/fundamentals/code_formatting.md) |[Reading Week 4](/Week1/README.md)|[Homework Week 4](/Week1/MAKEME.md)|
```
GeorgeSapkin: Same comment as for the other repo. I think the links should be absolute when pointing to another repo.

mkruijt: Why do you think links need do be absolute when pointing at another repo? I did this because there are three other schools using and forking our repo's.

GeorgeSapkin: If I fork or clone this repo the links will be broken. I think relative links shouldn't extend outside of current repository.

mkruijt: yes, the idea is that together with this repo you will also clone the fundamentals repo

remarcmij: If you clone the links to markdown files will not work on you local machine, but that is not different from how it was. If you fork, you must fork both the JavaScript repos and the fundamentals repo. These relative links will not win a beauty contest, but Maartje's point is I think valid.

mkruijt: I know it is a suboptimal solution, @ChrisOwen101 what do you think? This concerns you as well

mkruijt: yes the other option is absolute links, and that the other chapters will have to change all the links, which I know easily to fix. not sure what is the best solution though.

ChrisOwen101: I'm obviously biased here but it a lot better for us for them to be relative links. It makes pulling down changes from HYF a lot easier (fewer conflicts) and reduces confusion if I happen to miss one of the links when changing them. Once our students get over the the HYF side it can get a little confusing for them and they find it hard to get back!

I understand your concerns, but since this is a living document it would be really beneficial for us to make it easy as possible to keep up to date.

remarcmij: Reading Week 4 -> Reading Week 1
same for other columns

mkruijt: thanks Jim, fixed it

mkruijt:
@GeorgeSapkin what if we put a message in the top of the readme file that if you fork this repo you also need to fork the fundamentals repo in order to make the links work?

PR #2, opened 2018-03-19
move readings to corresponding weeks

mkruijt:


Week3/MAKEME.md
```diff
@@ -7,7 +7,7 @@ Topics discussed this week:
 • Callbacks
 ```
 
->[Here](/Week3/README.md) you find the readings you have to complete before the seventh lecture.
+>[Here](https://github.com/HackYourFuture/JavaScript3/tree/master/Week1) you find the readings you have to complete before the first JavaScript3 lecture.
```
remarcmij: I think we can get rid of step 2 now and renumber the steps.

PR #3, opened 2018-04-05
added link to pr

mkruijt:


GeorgeSapkin:
The description doesn't match the diff. Is something missing?

mkruijt:
ah no, sorry I was doing this for 5 repo's at the same time. I only added a link to make a pr. the cc copyright was already in this readme sorry for the confusion. So it's complete only my description is not correct.

PR #4, opened 2018-04-30
Enable handing in homework via PR

remarcmij:
Adds starter templates for the homework assignments
Adds instructions how to hand in homework via PR


remarcmij:
Hi @mkruijt, I thought this was obsolete stuff that could be taken out (as I moved the example code out of JS3). I think these examples were last used with Timir. However, I can put them back in if you want to keep them.

PR #11, opened 2018-05-12
finished homework

AleOrtizA:
Hello, I have problems with images not loading.

### DIFF
--- a/Week1/homework/app.js
+++ b/Week1/homework/app.js
@@ -1,11 +1,172 @@
 'use strict';
-{
-  const bookTitles = [
-    // Replace with your own book titles
-    'harry_potter_chamber_secrets'
-  ];
+//https://github.com/HackYourFuture/JavaScript2/blob/master/Week1/MAKEME.md
+
+let textOfH1 = document.getElementById("this-is-the-header");
+console.log(textOfH1);
+textOfH1.textContent = "Hello From The Book Space";
+document.getElementById("this-is-the-header").style.fontFamily = "Impact,Charcoal,sans-serif";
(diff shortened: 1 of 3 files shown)


harditsingh:
Hey Alejandra! 
Good work on the assignment! Here are a few things that I noticed in your assignment:

- **booksListImg(), generateUI(), app.js:** These are not pure functions, you are making references to external objects! Always pass the variables as arguments instead of making external reference.

- Your code should be neatly written inside one (or more) function(s). You have written many statements out of functions which is not a nice design choice.

- You have used both arrow functions (foo, bar => {}) and normal functions (function(foo, bar) {}). You should always stick to one style.

- **Line 126, app.js:** You used 'var' to declare a variable! You should always use either let or const to do so.

- There are a lot of console.log() statements in you app.js. This was not a requirement as per the a assignment and hence you should not put them there.

- **myFaveBookCovers, app.js:** You tried to find a remote resource in the img folder(local directory), img/<link-to-image> does not work. You need to refer to the img folder only if you have local resources, if you have just remote resources, just put the link there.

- You have not named the variables very well, there are a lot of spellling mistakes. Also, on Line 168, you have used 'x' as a name for a variable. This is not a nice name and hence not advisable.

- You have use JS for adding styles to the elements, you should always try to avoid that. You can achieve the same thing by adding 'classes' to your newly created HTML tags and then stylize those classes in your CSS Stylesheet. Another point to note here is that you have directly styled the HTML tags instead of using classes. This can become problematic when your website grows in size!

- **bookData, myFaveBookCovers, app.js:** While it is not incorrect to put the keys of literal objects in double-quotes, you should always refrain from doing so.

Cheers!

PR #12, opened 2018-05-12
Homework week1

neveenatik:
I couldn't yet figure how to publish the website so the link to the website included in homework/README.md
doesn't work yet.


harditsingh:
Hey Neveen!

Excellent work on the assignment, your code looks very nice! I have a few remarks though:

- You were supposed to put the key in the books object instead of book1, book2... This is how it should have looked:
```
const books = {
    forty_rules_of_love: {
      bookTitle: 'Forty rules of love',
      language: 'Turkish',
      author: 'Elif Shabak'
    },
    azazel: { ...
...
};
```

- Try to avoid adding style from JS, use CSS to do that, will be easier to debug when things go wrong. 

- **Line 75, 80, app.js:** The variables divWrapper and wrapper sound very similar, try to be a bit more specific while naming.
 
- **ImageIdEncapsulation(), app.js:** This is not a pure function as it is referring to an external object (books). Also, the name of the function has its first letter capitalized, which is not in accordance with the naming convention.
 
- **Line 113-114, app.js:** I am not quite sure what you tried to do in these line, could you maybe explain these statements a bit?
 
- **style.css:** You should try to avoid styling the HTML tags directly, using classes instead is a better idea.

Keep up the great work!

neveenatik:
Hi Hardit,

Thanks a lot for your review, I will work on these points.
I have a question over naming of an object constructor isn't that suppose to be capital or maybe I misunderstood constructor 😆 
Thanks again for your time and effort!

PR #13, opened 2018-05-12
Homework Week1

ThreePalmTrees:


ThreePalmTrees:
Hey Hardit,
Thank you for your review, I will work on the homework right now and fix the bugs.
Just one quick question.. 
Is it okay to keep using the tag names in the css file considering that we're probably not gonna grow this site in the future? (or are we gonna comeback later to it and add more shiny things? :smile: ).
Have a nice evening.

PR #14, opened 2018-05-12
Added javascript2 homework week1

Dumie1:
I have made some changes and added Javascript2 homework week1.


harditsingh:
Hey Dumie!

Your website looks good, but the code behind it has some issues:

- **app.js:** You have implemented the problem 2.4 using instances of a constructor function. This complicates things a lot. First, it would have been a lot easier to solve this problem using Object Literals. Second, your code now has static references to your data. This makes your program very rigid and you have to manually rewrite the code if you want to add another book for example. Using Object Literal would have been helpful as data can be directly loaded from JSON files and you don't have to change your code as your data varies.

- **index.html:** You have added a fix number of <div> and  for storing the data for the books. That actually defeats the purpose of the assignment, in this assignment we aimed to load the data dynamically, we cannot prepare the HTML beforehand because we don't know how much data we have.

- **app.js:** There a few more issues:
  - You code is not very well structured. 
  - You haven't automated the adding of data to the HTML page using a loop, you have done that manually instead. This makes the code very rigid.
  - Same goes for the loading of images. You haven't stored the images in an array either
  - You just added an event listener to the main() function which is just adding the sources of the pictures. Your code might give an error if the rest of the JS runs before the HTML is loaded.

If you need any help regarding understanding the assignment properly, we can organize a Hangout session, just drop me a message on Slack. This would be a good time to go back to this assignment and correct your error because you don't have a class this week!

Best of luck, will wait for your message!

PR #16, opened 2018-05-13
Homework week1

PKnar:


PKnar:
Dear reviewer as I mentioned in the class , I misunderstood the last instruction.I thought we had to create  different list for images only. This is why you see 2 different lists , one for books, second one for images.

harditsingh:
Hey Knarik!

Great job on the assignment! The double list is okay for now, you can correct that when you have time, that is also probably causing the misalignment of images and the book descriptions. I'd like to mention a couple of things though:
- **app.js:** You should stick to one of the following styles for functions, you have used both in your code:
`function createList(obj) {} (Line 19)
or
const createList = (obj) => {} (Line 106, 143)`
- **app.js:** You should try to avoid variable names like 'obj1' (Line 143)

Everything else is great, keep up the good work!

PKnar:
Hi   @harditsingh thanks for the feedback :+1: 

yea I wanted to practice both functions, as you see ' function createList(obj) {}' was commented out, I used that for the first part only. We were supposed to delete or change it after we're done with it, but I  left it there and commented it out for the review, I wanted to make sure I got both function styles right.

Thanks again for the feedback and yes I will stop using 'obj1' type of function names :smile:

PR #17, opened 2018-05-13
Home work week1 JS2

AllouhM:


silksil:
Hi Mahmoud, well done. In the browser everything looks clean and the functionality is there. Nonetheless, there is always room for improvement. Thus, some feedback: 

- **first part:** I see you left in the first part of the exercise, which you don't use it in the actual application.  It's good that you left it there, because it allows us to review your work, nonetheless companies that potentially will hire and look to your Github won't know this. So I personally would advice to mention this in a comment or put it in separate file. 

- **css:** in general well structured, but try to be consistent with the spaces and tabs. 

- **images:** i see you also have a image folder with images, besides URL's. Delete the things that you don't use, which in this is case is the image folder. 

- **semicolon:** here and there you've forgot to put a `; `after your functions  (e.g. line 132)

- **line 104:** in the createBookList you loop through the bookDetails object, nonetheless you do this without passing it as an argument. Although it works, it's not good practice. For example, due to this you would not be able to reuse the createBookList function for a different object than Bookdetails. So next time do something like this:
```javascript
function createContentPage(parentObject) { 
      //loop
}
createContentPage(bookDetails)
```

```

- **line 150:** the comment above also applies to the function content

- **line 150:** you could be more descriptive with naming the function > e.g. handleImageDisplay

- **line 150:** the functions works well, nonetheless you are relying on two objects (bookDetails and covers), whereas you could have looped through only the object covers in order to have the same results. This would have make the function easier to understand and less buggy, as it is relying on less external functionality. 

- How you organized the for loops was perfectly fine, but you could also consider to organize the loops a bit different next time as advised in the exercise using Object.keys and Object.values. You can find an example below;

``` javascript
function loopNestedObjec(parentObject) {
  for (const childObject in parentObject) {
    for (let i = 0; i < Object.keys(parentObject[childObject]).length; i++) {
      console.log(Object.keys(parentObject[childObject])[i] + ' :\t' + Object.values(parentObject[childObject])[i])
    }
  }
}
const books = {
   book1: {
     bookTitle: 'Forty laws of love',
     language: 'Arabic',
     author: 'Elif Şafak'
   },
   book2: {
     bookTitle: 'Alchemist',
     language: 'Arabic',
     author: 'Paulo Coelhho'
   }
}

loopNestedObjec(books)

  /*
prints:
  bookTitle :     Forty laws of love
  language :      Arabic
  author :        Elif Şafak
  bookTitle :     Alchemist
  language :      Arabic
  author :        Paulo Coelhho
*/


```

AllouhM:
Dear Sil, Thank you very much for the efforts you did in reviewing my homework. I will revisit my homework and try to apply the points you illustrated. For the semicolon after functions, I thought that no need for it after functions. It seems that I am mistaken. 
Thank you again.

PR #19, opened 2018-05-14
Added homework for week1

karam-ajaj:


silksil:
Hi Karam, good job! Looks all nice and clean. p.s.  I also like your book choice -- Siddharta is probably my favourite book I've ever read.

Here are some comments

**- line 6:** I see you left in the first part of the exercise, which you don't use it in the actual application. It's good that you left it there, because it allows us to review your work, nonetheless companies that potentially will hire and look to your Github won't know this. So you could delete it know so you make sure you keep your CV / Github clean. 

- line 79: you did will by passing the books object, nonetheless you did not passed the images object. Instead you try to access them through global variables, which is not good practice. So:

```javascript
function createLists(books, bookImages) {

````

karam-ajaj:
Thank you for the feedback, I will take these points in consideration.

PR #22, opened 2018-05-16
Homework week2

Meazer:


Week2/homework/maartjes_work.js
```diff
@@ -44,4 +44,8 @@ const tuesday = [
 
 const tasks = monday.concat(tuesday);
 
-// Add your code here
+const earned = tasks.map(task => task.duration / 60)
+  .filter(task => task >= 2)
+  .reduce((sum, task) => sum += task * 16, 0);
```
silksil: I see you include the hourly salarly(16) in the reduce method. It works and it uses little code, but it's a bit hard to interpret where the 16 is referring to.  I personally would have put it either the 16 in a variable with a clear name or would put in a .map function to make it more clear what is happening, example:

.map(perHourSalary => (perHourSalary * 16))

Week2/homework/maartjes_work.js
```diff
@@ -44,4 +44,8 @@ const tuesday = [
 
 const tasks = monday.concat(tuesday);
 
-// Add your code here
+const earned = tasks.map(task => task.duration / 60)
+  .filter(task => task >= 2)
```
silksil: you could have be a bit more descriptive in what is happneing, e.g. filterTasks

Week2/homework/maartjes_work.js
```diff
@@ -44,4 +44,8 @@ const tuesday = [
...
+  .reduce((sum, task) => sum += task * 16, 0);
+
+console.log('Maartje has earned €' + Math.round(earned * 100) / 100);
```
silksil: I personally I think it would be a bit more clean if you would have put the method in the function and only console log earned.

Also, you could also consider .toFixed(2) instead of doing Math.round which is probably a bit less code and a bit more descriptive.

PR #29, opened 2018-05-21
added homework-week2

Wael-Alhomsi:


Week2/homework/maartjes_work.js
```diff
@@ -45,3 +45,16 @@ const tuesday = [
 const tasks = monday.concat(tuesday);
 
 // Add your code here
+function calculateEarnings(arrayOfTasks, hourlyWage) {
```
silksil: I like the fact that you passed on two arguments, instead of using global variables.

One comment: you could have passed the argument also directly into the arrow function:

 const calculateEarnings = (arrayOfTasks, hourlyWage) =>
    arrayOfTasks.map((task) => task.duration / 60)
    .filter((hoursAmount) => hoursAmount >= 2)
    .map((paidHoursAmount) => paidHoursAmount * hourlyWage)
    .reduce((sum, amount) => sum + amount, 0)
    .toFixed(2);

PR #34, opened 2018-05-27
Added homework week2

neveenatik:
Re-added homework after fixing branching from branch homework-week1 instead of master.

### DIFF
--- a/Week2/homework/maartjes_work.js
+++ b/Week2/homework/maartjes_work.js
@@ -45,3 +45,9 @@ const tuesday = [
 const tasks = monday.concat(tuesday);
 
 // Add your code here
+const totalEarnings = tasks
+  .filter((hour) => hour.duration >= 120)
+  .map((earning) => (earning.duration / 60) * 50)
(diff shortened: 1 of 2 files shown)


Week2/homework/maartjes_work.js
```diff
@@ -43,5 +43,12 @@ const tuesday = [
 ];
 
 const tasks = monday.concat(tuesday);
+const hourlyRate = 50;
 
 // Add your code here
+const totalEarnings = tasks
```
silksil: Great job. For the next time you could also consider creating an arrow functions to which you pass on arguments. Something like this: 

const calculateEarnings = (arrayOfTasks, hourlyWage) =>
arrayOfTasks.map((task) => task.duration / 60)
.filter((hoursAmount) => hoursAmount >= 2)
.map((paidHoursAmount) => paidHoursAmount * hourlyWage)
.reduce((sum, amount) => sum + amount, 0)
.toFixed(2);

console.log(calculateEarnings(tasks, hourlyRate))

PR #35, opened 2018-05-30
Knar Poghosyan - Week 3 Homework 

PKnar:


Week3/homework/1-step3.js
```diff
@@ -1,11 +1,13 @@
 'use strict';
 
 function foo(func) {
-  // What to do here? 
+  // What to do here?
+  func();
 }
```
rohni: It is generally a good idea to put the func call in a return statement.  You don't know if it is going to return a value that should be returned from your function.  In things like filter, map and reduce it is.
`return func();`
Would be a better idea.

Week3/homework/4-step3.js
```diff
@@ -1,2 +1,73 @@
...
+console.log(theMotorBike);
+
+//instruction 2 -> Give myCar a nickname property with a string value.
+
+const car = function () {
+    this.wheels = 4;
+    this.engines = 1;
+    this.seats = 5;
+};
+
```
rohni: For your constructor functions, i.e. motorBike() and car(), because they are designed to be called to create a new object, convention is that they should be capitalised so:
MotorBike
Car
Would be considered better names.  Otherwise the code is excellent.

Week3/homework/6-step3.js
```diff
@@ -4,3 +4,13 @@ const arr2d = [[1, 2], [3, 4], [5, 6]];
...
+
+const printAllItems = (arr) =>
+    arr.map((item, index) => {
+        return index + ". " + item;
+    }).join('\n');
+
+console.log(printAllItems(arr2d));
+console.log(printAllItems(arr3d));
+
+
```
rohni: Interesting solution, this approach did not even occur to me.

Does your solution also work with something like this?
```
```
```
> console.log(printAllItems(arrNd))
0. 1,2,3,4,5,6,7,8,9,10,11,12
1. 13,14
```
Yes it does.  Fair enough.  Well done.
I think what was being looked for was creating a recursive function.  Are you able to print the elements out (index is not needed) using recursion?

Week3/homework/7-step3.js
```diff
@@ -23,3 +23,27 @@ console.log(y);
...
+                    -> PASS BY REFERENCE 
+While when we pass a reference value as a function argument the original value changes too.
+
+And this could be compared with an Editor.
+When we edit a document and save it. It replaces the old version with the new version.
+This is why when we console.log the reference value,
+after changing it inside the function, we get the new version of it.
+
+*/
+
```
rohni: Wow, really excellent explanation.
Very clear.  Well done!

Week3/homework/step4-bonus.js
```diff
@@ -4,7 +4,15 @@ const values = ['a', 'b', 'c', 'd', 'a', 'e', 'f', 'c'];
...
+}
 
+const uniqueValues = storeUniqueValues(values);
 console.log(uniqueValues);
+
+//another example with numbers
+const numbers = [5, 4, 7, 13, 11, 26, 5, 7, 13, 30, 25, 17, 28, 21];
+
+const uniqueNumbers = storeUniqueValues(numbers);
+console.log(uniqueNumbers);
```
rohni: Ha Ha, well done.  This was my solution as well. 😄

rohni:
Excellent work!  
I only had a couple of small comments.  Really nicely done.

Week3/homework/step4-bonus.js
PKnar: Good to know :smiley:

PKnar:
Hi @rohni ! Thanks so much for your detailed feedback !! :smiley:

Week3/homework/4-step3.js
PKnar: I see, thanks for the information :+1: ( I fixed that after your review)

Week3/homework/6-step3.js
PKnar: Since in the homework instruction it was not specified that we need to use a recursive function. I tried to come up with something else. :sweat_smile:  Since you and Jim already posted your solutions on Slack, I will just learn from it, because I don't think I can come up with a better solution than yours  at this moment :smile:

PR #36, opened 2018-05-30
Homework Week 3

karam-ajaj:


Week3/homework/1-step3.js
```diff
@@ -1,7 +1,7 @@
 'use strict';
 
 function foo(func) {
-  // What to do here? 
+  func();
 }
```
rohni: Hi Karam, 
It is generally a good practice to put the call to func in a return statement, so the value is returned from your function.  In filter, map and reduce it is the return value that is important.
```
return func();
```
Otherwise, nice answer.

Week3/homework/3-step3.js
```diff
@@ -2,23 +2,55 @@
...
+  else {
+    let temp = '';
+    do {
+      temp = temp + str;
+      num--;
+    } while (num > 0);
+    str = temp;
+  }
   return str;
 }
```
rohni: Really excellent that you tested num for bad values first.

Week3/homework/5-step3.js
```diff
@@ -1,2 +1,14 @@
...
+  for (let i = 0; i < arr.length; i++) {
+    for (let j = 0; j < arr[i].length; j++) {
+      product = product * arr[i][j];
+    }
+  }
+  return product;
+}
+
+const product = multiplyAll([[1, 2], [3, 4], [5, 6, 7]]);
+console.log('The product is: ' + product);
```
rohni: Nice and clean.

Week3/homework/6-step3.js
```diff
@@ -3,4 +3,25 @@
...
+  for (let j = 0; j < arr3d[i].length; j++) {
+    arrText3 += arr3d[i][j] + '   ';
+  }
+  console.log(arrText3);
+  arrText3 = '';
+}
+
+/* There are many ways to print out a multiple dimension arrays depending on the purpose of this procedure */
+console.log(arr2d);
+console.log(arr3d);
```
rohni: The idea here is that you want each item to be printed on its own.  What you are doing is for each element in the loop, checking if it is also an array, and if so, descending into that array, and printing its elements, if any of them is an array, descend. Rinse and repeat until you have traversed the whole tree (nested arrays like this are essentially trees).

As an example here is how I tackled this problem, and a rather nasty mess of nested arrays to test it on.

``` 

const runArray = (arr) => {
  arr.map((elem) => {
    if (Array.isArray(elem)) {
      return runArray(elem);
    } else {
      console.log(elem);
    }
  });
};

// runArray(arr2d);
// runArray(arr3d);
runArray(arrNd);
```
I hope this helps illustrate (one) way to solve this.  The important part is detecting an array, and then recalling the function with this array.

Week3/homework/7-step3.js
```diff
@@ -21,5 +21,6 @@ f2(y);
 
 console.log(y);
 
-// Add your explanation as a comment here
-
+/*
+f1() is a pointer and f2() is a reference
+*/
```
rohni: Nope.  f1() is passed an atomic value (num, boolean etc.) which are passed by value, meaning that they are copied.  The second f2() is a compound object and is passed by reference.

karam-ajaj:
@rohni Thank you Rohan for the feedback, I will take these points in consideration.

PR #42, opened 2018-06-01
added homework-week3

Dumie1:
I have added javascript2 homework-week3.

### DIFF
--- a/Week3/homework/1-step3.js
+++ b/Week3/homework/1-step3.js
@@ -1,9 +1,11 @@
 'use strict';
 
 function foo(func) {
-  // What to do here? 
-}
+  const x = 12;
 
+  console.log('My address is :' + x);
+  func();
+}
 function bar() {
   console.log('Hello, I am bar!');
 }
--- a/Week3/homework/2-step3.js
+++ b/Week3/homework/2-step3.js
@@ -1,8 +1,28 @@
 'use strict';
 
+function sayThree(threeCallback) {
+  console.log(threeCallback + ' is divisible by three');
+}
+
+function sayFive(fiveCallback) {
+  console.log(fiveCallback + ' is divisible by five');
+}
+
 function threeFive(startIndex, stopIndex, threeCallback, fiveCallback) {
-  const values = [];
-  // Add your code here
+
+  const values = [];  //function generating an array containing values from start value to end value
+  for (let i = startIndex; i <= stopIndex; i++) {
+    values.push(i);  
+
+    if (i % 3 === 0) {
+      threeCallback(i);
+    }
+
+    if (i % 5 === 0) {
+      fiveCallback(i);
+    }
+  }
+  console.log(values);
 }
 
 threeFive(10, 15, sayThree, sayFive);
--- a/Week3/homework/3-step3.js
+++ b/Week3/homework/3-step3.js
@@ -1,25 +1,44 @@
 'use strict';
 
-// use a 'for' loop
 function repeatStringNumTimesWithFor(str, num) {
-  // add your code here
-  return str;
+
+  let strRepeated = '';
+
+  for (let i = 0; i < num; i++) {
+    strRepeated += str;
+  }
+
+  return strRepeated;
 }
 
 console.log('for', repeatStringNumTimesWithFor('abc', 3));
 
 // use a 'while' loop
 function repeatStringNumTimesWithWhile(str, num) {
-  // add your code here
-  return str;
+  let accumulatedStr = '';
+
+  while (num > 0) {
+    accumulatedStr += str;
+    num--;
+  }
+
+  return accumulatedStr;
 }
 
 console.log('while', repeatStringNumTimesWithWhile('abc', 3));
 
 // use a 'do...while' loop
 function repeatStringNumTimesWithDoWhile(str, num) {
-  // add your code here
-  return str;
+
+  let accumulatedStr = '';
+
+  do {
+    accumulatedStr += str;
+    num--;
+  }
+  while (num > 0);
+
+  return accumulatedStr;
 }
 
 console.log('while', repeatStringNumTimesWithDoWhile('abc', 3));
--- a/Week3/homework/4-step3.js
+++ b/Week3/homework/4-step3.js
@@ -1,2 +1,15 @@
 'use strict';
-// paste your freeCodeCamp solutions in here
+
+const Car = function () {
+  this.wheels = 4;
+  this.engines = 1;
+  this.seats = 5;
+};
+
+// Only change code below this line.
+
+const MotorBike = function () {
+  this.wheels = 2;
+  this.engines = 1;
+  this.seats = 1;
+};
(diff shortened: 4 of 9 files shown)


Week3/homework/1-step3.js
```diff
@@ -1,9 +1,11 @@
 'use strict';
 
 function foo(func) {
-  // What to do here? 
-}
+  const x = 12;
 
+  console.log('My address is :' + x);
+  func();
```
rohni: The value that is returned from calling `func()` should be returned from the function. e.g.
```
return func()
```
`foo` should know nothing about the function it is calling, and so can't make the assumption that the return value is not needed somewhere else.

Week3/homework/1-step3.js
```diff
@@ -1,9 +1,11 @@
 'use strict';
 
 function foo(func) {
-  // What to do here? 
-}
+  const x = 12;
```
rohni: I'm guessing these two lines were testing something else out. 😉 
They don't really have a place in this exercise.

Week3/homework/2-step3.js
```diff
@@ -1,8 +1,28 @@
...
+    if (i % 3 === 0) {
+      threeCallback(i);
+    }
+
+    if (i % 5 === 0) {
+      fiveCallback(i);
+    }
+  }
+  console.log(values);
 }
```
rohni: ✅ Yup, excellent.

Week3/homework/3-step3.js
```diff
@@ -1,25 +1,44 @@
...
-  return str;
+
+  let strRepeated = '';
+
+  for (let i = 0; i < num; i++) {
+    strRepeated += str;
+  }
+
+  return strRepeated;
 }
```

Week3/homework/3-step3.js
```diff
@@ -1,25 +1,44 @@
...
-  return str;
+  let accumulatedStr = '';
+
+  while (num > 0) {
+    accumulatedStr += str;
+    num--;
+  }
+
+  return accumulatedStr;
 }
```

Week3/homework/3-step3.js
```diff
@@ -1,25 +1,44 @@
...
+  let accumulatedStr = '';
+
+  do {
+    accumulatedStr += str;
+    num--;
+  }
+  while (num > 0);
+
+  return accumulatedStr;
 }
```

Week3/homework/4-step3.js
```diff
@@ -1,2 +1,15 @@
...
+  this.seats = 5;
+};
+
+// Only change code below this line.
+
+const MotorBike = function () {
+  this.wheels = 2;
+  this.engines = 1;
+  this.seats = 1;
+};
```

Week3/homework/5-step3.js
```diff
@@ -1,2 +1,17 @@
...
+    for (let x = 0; x < arr[i].length; x++) {
+
+      product = product * arr[i][x];
+    }
+  }
+  
+  return product;
+}
+
+multiplyAll([[1, 2], [3, 4], [5, 6, 7]]);
```

Week3/homework/7-step3.js
```diff
@@ -10,6 +10,11 @@ f1(x);
 
 console.log(x);
 
+//Explanation for first code-----
+/* When you pass a primitive value into a function, as an argument, you pass 
+a copy of it and original copy is not affected. Hence we get the result 9.
+Nine is a primitive value.
+-----------------------------------------------------------------------------*/
```

Week3/homework/step4-bonus.js
```diff
@@ -1,10 +1,16 @@
...
+    a[b.name] = (a[b.name] || 0) + b.count
+    return a;
+  }, {});
 
-// Replace `yourFunction` with the name of the function you just created
-const uniqueValues = yourFunction(values);
+const sorted = Object.keys(uniq).sort((a, b) => uniq[a] < uniq[b])
 
-console.log(uniqueValues);
+console.log(sorted);
```
rohni: ✅ Wow, I love this solution it is the most creative solution to this problem I've seen.  And sorted. Nice 👍 
I was worried about efficiency, but actually, the typical answer that uses `array.includes` does far more traversals.
If you just want to end up with your object in one traversal, although with out the nice count, you can skip the map and just do the reduce.  I'll keep getting the keys as a separate expression for readability.
```
const uniqObj = names.reduce((acc, item) => {
  return {...acc,
          ...{[item]: null}};
}, {});

const uniques = Object.keys(uniqObj);
```
So, very nice work especially using higher-order functions to such effect. 😄

Week3/homework/step4.js
```diff
@@ -1,8 +1,13 @@
 'use strict';
 
-// Add your code here
+const addSix = function (x) {
+  return function createBase() {
+    return x + 6;
+  };
+};
```
rohni: ✅ This does what is asked, however, I would be inclined to create a more generic version.  Imagine you were then asked to create an `add8` function.  You would have to copy a lot of code.
```
const createBase = function(x) {
  return function(y) {
    return x + y;
};
};
const addSix = createBase(6);
const addEight = createBase(8);
console.log(addSix(12)); // ==> 18
console.log(addEight(12)); // ==> 20
```
You can see that if `createBase` is able to receive a parameter, it will create an `add` anything function for you.  This makes for much better reuse, which means a lot less typing in the long run. 😉

Week3/homework/6-step3.js
```diff
@@ -1,6 +1,18 @@
...
+  const rowSizes = []
+
+  for (let i = 0; i < rowCount; i++) {
+    rowSizes.push(arr2d[i].length)
+  }
+  return [rowCount, Math.min.apply(null, rowSizes)]
+}
+
+arr3d([[[1, 2], [3, 4]], [[5, 6], [7, 8]]]);
+console.log('The length of array is : ' + (arr2d.length));
```
rohni: I think you might have missed the point of this exercise.
The idea is to go into each of the arrays and its "sub" arrays find the values (i.e. not arrays) and do something with them (i.e. print the value out).

So, your function is printing each row, but not finding the rows within the rows and printing the values from them.  The best way to tackle this is using a technique called `recursion`.  What that does, is test for some condition, and if it is true, the function recalls itself with a subset of the original value. It is best seen in code:
```

  let values = [];
    if (Array.isArray(arr[i])) {
    } else {
    }
  }
  return values;
};

console.log(printAllTheItems(arr2d));
console.log(printAllTheItems(arr3d));
console.log(printAllTheItems(arrNd));
```
Actually this flattens the array (using for loops) and then prints the results of flattening each of the arr2d, arr3d, and arr4d.
As I've seen from your last question you are not afraid of reduce, so here is a reduce based version:
```
const flatten = (arr, result=[]) => {
  return arr.reduce((acc, elem) => {
    if (Array.isArray(elem)) {
      return acc.concat(flatten(elem, result));
    } else {
      return acc.concat([elem]);
    }
  }, result);
}

flatten(arrNd);
```
The important point in both of these is that there is a test to see whether the current item is an Array.  If it is, we know we have to go into the array, so we call the function again, and pass the current item (an array) into the function.  We then `concat` the return of that call onto our `acc` (accumulator or results array), and keep going.

The nice thing with recursion, and what makes it so important, is that it won't care how deep the arrays are nested in arrays, it will just keeps going collecting values.

I hope this gave some insights. 😄

PR #43, opened 2018-06-01
Homework week3

amirno1:


Week3/homework/1-step3.js
```diff
@@ -1,7 +1,7 @@
 'use strict';
 
 function foo(func) {
-  // What to do here? 
+  func();
 }
```
rohni: I would suggest returning the results of calling `func()`.
`foo` does not know anything about func, just that it needs to call it. 
So, it should be using (i.e. returning) the results of the function call.

Week3/homework/2-step3.js
```diff
@@ -2,7 +2,23 @@
...
+      threeCallback(i);
+    }
+  }
+}
+function sayThree(i) {
+  console.log(i + ' is divisible by 3');
+}
+function sayFive(i) {
+  console.log(i + ' is divisible by 5');
 }
```
rohni: ✅

Week3/homework/3-step3.js
```diff
@@ -2,23 +2,52 @@
...
+  for (let i = 0; i < num; i++) {
+    if (num >= 0) {
+      value += str;
+    } else {
+      return str = "";
+    }
+  }
+  str = value;
   return str;
 }
```

Week3/homework/3-step3.js
```diff
@@ -2,23 +2,52 @@
...
+    if (num >= 0) {
+      value += str;
+      i++;
+    } else {
+      return str = "";
+    }
+  }
+  str = value;
   return str;
 }
```

Week3/homework/3-step3.js
```diff
@@ -2,23 +2,52 @@
...
+      value += str;
+      i++;
+    } else {
+      return str = "";
+    }
+  }
+  while (i < num);
+  str = value;
   return str;
 }
```
rohni: ✅ Nice that you check to make sure that `num` has a valid starting value.

Week3/homework/4-step3.js
```diff
@@ -1,2 +1,19 @@
...
+function dog() {
+  this.name = "Rupert";
+  this.color = "brown";
+  this.numLegs = 4;
+}
+// Add your code below this line
+
+const hound = new dog();
+
+console.log(hound);
```

Week3/homework/5-step3.js
```diff
@@ -1,2 +1,18 @@
...
+  }
+  console.log(product);
+}
+
+// Modify values below to test the code
+multiplyAll([[1, 2], [3, 4], [5, 6, 7]]);
+
+multiplyAll([[1], [2], [3]]);
+
+multiplyAll([[5, 1], [0.2, 4, 0.5], [3, 9]]);
```
rohni: ✅ Nice work!

Week3/homework/6-step3.js
```diff
@@ -4,3 +4,19 @@ const arr2d = [[1, 2], [3, 4], [5, 6]];
...
+function printTheItemsOfArray(arr) {
+  for (let i = 0; i < arr.length; i++) {
+    for (let j = 0; j < arr[i].length; j++) {
+
+      printAllTheItems += (arr[i][j] + ' ');
+    }
+  }
+  console.log(printAllTheItems);
+}
+
```
rohni: While this works for the given examples, will it work for this nested array?
```
```
- I would suggest looking into `recursion`.  It takes some getting used to, but once you get the idea, for the same amount of code it would handle any depth of nesting.
An example that does not mutate any variables:
```
  let values = [];
    if (Array.isArray(arr[i])) {
    } else {
    }
  }
  return values;
};
```

- I would also have kept the array where you accumulated the values in the function.  Otherwise it is not safe from being changed by other code.
- The name of the array that accumulates the values is also confusing.  I would be inclined to call it `allTheValues` as it does not `print` anything, just holds values.  The function does the printing.
In general, you want to name data using nouns, and functions as verbs.

Week3/homework/7-step3.js
```diff
@@ -23,3 +23,8 @@ console.log(y);
 
 // Add your explanation as a comment here
 
+/*In Javascript all the primitives are passed by value 
+so it means we only manipulate the value in the function and
+once the function is finished, our variable has still the 
+same value. Unlike Objects that are passed by reference and once 
+you change it using even functions the object will be also changed*/
```

Week3/homework/step4-bonus.js
```diff
@@ -2,9 +2,12 @@
...
 
-// Add your function here. Try and come up with a good name for this function
+function makeItUnique(arr) {
+  return Array.from(new Set(arr));
+}
+
 
 // Replace `yourFunction` with the name of the function you just created
-const uniqueValues = yourFunction(values);
+const uniqueValues = makeItUnique(values);
```
rohni: ✅ Very nice use of making different datatypes work for you.  Of course, the `Set` did not exist in Javascript a few years ago, and you had to do it another way, usually using a filter.

Week3/homework/step4.js
```diff
@@ -1,6 +1,14 @@
...
 
 // Add your code here
+function createBase(num1) {
+  const func = function (num2) {
+    let result = 0;
+    result = num1 + num2;
+    console.log(result);
+  };
+  return func;
+}
```
rohni: ✅ As a point of style and usefulness when creating higher-order functions.  I would not just print the value out, but also return it from the function.
Otherwise very nicely laid out.

rohni:
Amir,
Nice job!  I've added a few comments, but the work was clean and well constructed.  Well done!

Good luck with the rest of the course, and the next step of using your new skills to make the world a better place. 😄

PR #44, opened 2018-06-01
Homework week3

zmagzoub:


Week3/homework/1-step3.js
```diff
@@ -1,11 +1,12 @@
...
 function foo(func) {
-  // What to do here? 
+    // What to do here? 
+    func();
 }
 
 function bar() {
-  console.log('Hello, I am bar!');
+    console.log('Hello, I am bar!');
 }
```
rohni: I would return the value of calling func(). foo should know nothing about the function it is calling, and so should not make assumptions about it.  Usually one runs a function to use its return value.

Week3/homework/2-step3.js
```diff
@@ -1,8 +1,21 @@
...
 
 threeFive(10, 15, sayThree, sayFive);
+
+function sayThree(number) {
+    console.log(number + ' is divisible by three');
+}
+
+function sayFive(number) {
+    console.log(number + ' is divisible by five');
+}
```
rohni: ✅ Nicely done.  Very concise, and you only run through once, adding to the array as you go.  Very nice.

Week3/homework/3-step3.js
```diff
@@ -2,24 +2,38 @@
...
 function repeatStringNumTimesWithFor(str, num) {
-  // add your code here
-  return str;
+    // add your code here
+    let repeatedString = '';
+    for (let i = num; i > 0; i--) {
+        repeatedString += str;
+    }
+    return repeatedString;
 }
```

Week3/homework/3-step3.js
```diff
@@ -2,24 +2,38 @@
...
-  // add your code here
-  return str;
+    // add your code here
+    let repeatedString = '';
+    while (num > 0) {
+        repeatedString += str;
+        num--;
+    }
+    return repeatedString;
 }
```

Week3/homework/3-step3.js
```diff
@@ -2,24 +2,38 @@
...
-  // add your code here
-  return str;
+    // add your code here
+    let repeatedString = '';
+    do {
+        repeatedString += str;
+        num--;
+    } while (num > 0)
+    return repeatedString;
 }
```

Week3/homework/4-step3.js
```diff
@@ -1,2 +1,11 @@
...
 // paste your freeCodeCamp solutions in here
+//https://learn.freecodecamp.org/javascript-algorithms-and-data-structures/object-oriented-programming/use-a-constructor-to-create-objects/
+
+function Dog() {
+    this.name = "Rupert";
+    this.color = "brown";
+    this.numLegs = 4;
+}
+// Add your code below this line
+var hound = new Dog()
```
rohni: 👍

Week3/homework/5-step3.js
```diff
@@ -1,2 +1,14 @@
...
+        for (let j = 0; j < arr[i].length; j++) {
+            product *= arr[i][j];
+        }
+    }
+    // Only change code above this line
+    return product;
+}
+
+// Modify values below to test your code
+multiplyAll([[1, 2], [3, 4], [5, 6, 7]]);
```

Week3/homework/6-step3.js
```diff
@@ -4,3 +4,14 @@ const arr2d = [[1, 2], [3, 4], [5, 6]];
...
+function printArray(a) {
+    for (var i = 0; i < a.length; i++) {
+        for (var z = 0; z < a[i].length; z++) {
+            for (var y = 0; y < a[i][z].length; y++)
+                console.log(a[i][z][y]);
+        }
+    }
+}
+
+printArray(arr3d)
```
rohni: This will not work for arr2d.
It would also not work for something like:
```
```
I would suggest reading more up on `recursion`.
What you really want is something that will handle any level of depth of nesting.  Here is an example using for loops, and then calling itself.
```
  let values = [];
    if (Array.isArray(arr[i])) {
    } else {
    }
  }
  return values;
};

console.log(printAllTheItems(arr2d));
console.log(printAllTheItems(arr3d));
console.log(printAllTheItems(arrNd));
```
Fewer and shorter lines of code, and handles just about anything.

When I first tried to solve this problem I used map, and in the `class_15` slack group Jim has posted a for-loop version of a much more generic utility called flatten, I would suggest having a look.
This is a version using map:
```
const runArray = (arr) => {
  arr.map((elem) => {
    if (Array.isArray(elem)) {
      return runArray(elem);
    } else {
      console.log(elem);
    }
  });
};
```
I'm not that happy with it, because, unlike Jim's `flattenArray`, it is not reusable.

Week3/homework/7-step3.js
```diff
@@ -23,3 +24,6 @@ console.log(y);
 
 // Add your explanation as a comment here
 
+// In Pass by Value(the first function), Function is called by directly passing the value of the variable as the argument.Changing the argument inside the function doesn’t affect the variable passed from outside the function. In Javascript numbers and strings follows pass by reference.
+
+// In Pass by Reference(the second function), Function is called by directly passing the reference/address of the variable as the argument. Changing the argument inside the function affect the variable passed from outside the function. In Javascript objects and arrays follows pass by reference.
```

Week3/homework/step4-bonus.js
```diff
@@ -4,7 +4,8 @@ const values = ['a', 'b', 'c', 'd', 'a', 'e', 'f', 'c'];
 
 // Add your function here. Try and come up with a good name for this function
 
-// Replace `yourFunction` with the name of the function you just created
-const uniqueValues = yourFunction(values);
+const uniqueValues = values.filter(function (value, index) { return values.indexOf(value) === index });
+
+// I got help from here https://stackoverflow.com/questions/9229645/remove-duplicate-values-from-js-array
 
 console.log(uniqueValues);
```
rohni: 👍 Nice job with this.  Nice use of higher-order functions.

Week3/homework/step4.js
```diff
@@ -1,8 +1,14 @@
...
+        return firstNum + secondNum;
+    };
+}
 
 const addSix = createBase(6);
 
-addSix(10); // returns 16
-addSix(21); // returns 27
+console.log(addSix(10)); // returns 16
+console.log(addSix(21)); // returns 27
```
rohni: ✅ Excellent! Nice naming of the variables here.  Makes it very clear what is happening.

rohni:
Hi Zaed,
Nice work!  I've added a few comments.
There were a couple of points to look at, but other than that very clean, and correct code.  Well done!

Good luck with the rest of your course, and after that, using it out in the world. 
All the best.  😄

PR #46, opened 2018-06-02
Homework-week3

AleOrtizA:
Homework-week3

### DIFF
--- a/Week3/homework/1-step3.css
+++ b/Week3/homework/1-step3.css
@@ -1,0 +1,10 @@
+body,
+html {
+    margin: 25px;
+    padding: 25px;
+    width: 100%;
+    max-width: 100%;
+    height: auto;
+    overflow-x: hidden;
+    background-color: rgb(181, 181, 197);
+}
--- a/Week3/homework/1-step3.js
+++ b/Week3/homework/1-step3.js
@@ -1,7 +1,7 @@
 'use strict';
 
 function foo(func) {
-  // What to do here? 
+  console.log(func);
 }
 
 function bar() {
--- a/Week3/homework/3-step3.js
+++ b/Week3/homework/3-step3.js
@@ -2,24 +2,51 @@
 
 // use a 'for' loop
 function repeatStringNumTimesWithFor(str, num) {
-  // add your code here
-  return str;
-}
+  let times = 0;
+  let text = "";
 
-console.log('for', repeatStringNumTimesWithFor('abc', 3));
+  for (times = 0; times < num; times++) {
+    text = text + str;
+  }
+  return text;
+}
+console.log(repeatStringNumTimesWithFor("abc", 3));
+console.log(repeatStringNumTimesWithFor("abc", -4));
+
+repeatStringNumTimesWithFor("abc", 3);
+
 
 // use a 'while' loop
 function repeatStringNumTimesWithWhile(str, num) {
-  // add your code here
-  return str;
-}
+  let times = 0;
+  let text = "";
 
-console.log('while', repeatStringNumTimesWithWhile('abc', 3));
+  while (times < num) {
+    times += 1;
+    text = text + str;
+  }
+  return text;
+}
+console.log(repeatStringNumTimesWithWhile("abc", 3));
+console.log(repeatStringNumTimesWithWhile("abc", -1));
+
+console.log("while", repeatStringNumTimesWithWhile('abc', 3));
 
 // use a 'do...while' loop
 function repeatStringNumTimesWithDoWhile(str, num) {
-  // add your code here
-  return str;
+  let times = 0;
+  let text = "";
+  if (num <= 0) {
+    return "";
+  }
+  do {
+    text = text + str;
+    times += 1;
+  }
+  while (times < num);
+  return text;
 }
+console.log(repeatStringNumTimesWithDoWhile("abc", 3));
+console.log(repeatStringNumTimesWithDoWhile("abc", -1));
 
 console.log('while', repeatStringNumTimesWithDoWhile('abc', 3));
--- a/Week3/homework/4-step3.js
+++ b/Week3/homework/4-step3.js
@@ -1,2 +1,12 @@
 'use strict';
-// paste your freeCodeCamp solutions in here
+function Dog() {
+  this.name = "Rupert";
+  this.color = "brown";
+  this.numLegs = 4;
+}
+console.log(hound);
+
+
+const hound = new Dog();
+
+//in this pne Freecode camp doesnt give me an error, but the browser does.
--- a/Week3/homework/5-step3.js
+++ b/Week3/homework/5-step3.js
@@ -1,2 +1,20 @@
 'use strict';
-// paste your freeCodeCamp solutions in here
+function multiplyAll(arr) {
+  let product = 1;
+  // Only change code below this line
+  for (let i = 0; i < arr.length; i++) {
+    for (let j = 0; j < arr[i].length; j++) {
+      product *= arr[i][j];
+    }
+  }
+
+
+  // Only change code above this line
+  return product;
+}
+
+// Modify values below to test your code
+console.log(multiplyAll([[1], [2], [3]]));//return 6
+console.log(multiplyAll([[1, 2], [3, 4], [5, 6, 7]])); //return 5040
+console.log(multiplyAll([[5, 1], [0.2, 4, 0.5], [3, 9]])); //return 54
+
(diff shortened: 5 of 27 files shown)


Week3/homework/1-step3.css
```diff
@@ -0,0 +1,10 @@
+body,
```
rohni: Alejandra, 
Something has definitely not gone right here. 
You have committed (in one commit) all the code for the last 3 weeks.  This makes things much more difficult to sift through and find the work for week 3.
I would suggest that you do a few more git tutorials, as this skill is actually critical when programming in the real world.  
This time it is just you working on the repository, but imagine there are 10 different people working on it, and you suddenly create double commits through big chunks of the repository.  I can promise you (because I've done it) that your co-workers will not be happy with you, and you will be spending hours fixing it. 😄

Week3/homework/1-step3.js
```diff
@@ -1,7 +1,7 @@
 'use strict';
 
 function foo(func) {
-  // What to do here? 
+  console.log(func);
 }
 
 function bar() {
```
rohni: This won't work, unfortunately.  
You are supposed to call the function and return it.
`bar` is the one that calls `console.log`.
The line should look like this:
```
return func();
```

Week3/homework/1-step3.css
AleOrtizA: Hi Rohan, thank you for the feedback I definitely need to pay attention to the details. I will fix it and will be more careful in the future.

Week3/homework/3-step3.js
```diff
@@ -2,24 +2,51 @@
...
+
+  for (times = 0; times < num; times++) {
+    text = text + str;
+  }
+  return text;
 }
+console.log(repeatStringNumTimesWithFor("abc", 3));
+console.log(repeatStringNumTimesWithFor("abc", -4));
+
+repeatStringNumTimesWithFor("abc", 3);
```

Week3/homework/3-step3.js
```diff
@@ -2,24 +2,51 @@
...
+  let times = 0;
+  let text = "";
+
+  while (times < num) {
+    times += 1;
+    text = text + str;
+  }
+  return text;
 }
+console.log(repeatStringNumTimesWithWhile("abc", 3));
```

Week3/homework/3-step3.js
```diff
@@ -2,24 +2,51 @@
...
+    text = text + str;
+    times += 1;
+  }
+  while (times < num);
+  return text;
 }
+console.log(repeatStringNumTimesWithDoWhile("abc", 3));
+console.log(repeatStringNumTimesWithDoWhile("abc", -1));
 
 console.log('while', repeatStringNumTimesWithDoWhile('abc', 3));
```

Week3/homework/4-step3.js
```diff
@@ -1,2 +1,12 @@
...
+  this.name = "Rupert";
+  this.color = "brown";
+  this.numLegs = 4;
+}
+console.log(hound);
+
+
+const hound = new Dog();
+
+//in this pne Freecode camp doesnt give me an error, but the browser does.
```

Week3/homework/5-step3.js
```diff
@@ -1,2 +1,20 @@
...
+
+
+  // Only change code above this line
+  return product;
+}
+
+// Modify values below to test your code
+console.log(multiplyAll([[1], [2], [3]]));//return 6
+console.log(multiplyAll([[1, 2], [3, 4], [5, 6, 7]])); //return 5040
+console.log(multiplyAll([[5, 1], [0.2, 4, 0.5], [3, 9]])); //return 54
```

Week3/homework/step4-bonus.js
```diff
@@ -1,10 +1,6 @@
 'use strict';
 
-const values = ['a', 'b', 'c', 'd', 'a', 'e', 'f', 'c'];
-
-// Add your function here. Try and come up with a good name for this function
-
-// Replace `yourFunction` with the name of the function you just created
-const uniqueValues = yourFunction(values);
+const values = ['a', 'b', 'c', 'd', 'a', 'e', 'f', 'c', 'j', 'h', 'kk', 'pzc', 'aa', 'e', 'b', 'bb'];
+const uniqueValues = Array.from(new Set(values));
```
rohni: ✅ Excellent, this as how I solved the problem as well.

Week3/homework/step4.js
```diff
@@ -1,8 +1,12 @@
 'use strict';
 
-// Add your code here
+function createBase(x) {
+  return function (num) {
+    return x + num;
+  };
+}
```
rohni: ✅ Excellent.

Week3/homework/7-step3.js
```diff
@@ -21,5 +21,10 @@ f2(y);
 
 console.log(y);
 
-// Add your explanation as a comment here
+// In this article is explained easily. 
+//https://codeburst.io/javascript-passing-by-value-vs-reference-explained-in-plain-english-8d00fd06a47c
+
+//Basically it says that when we have two primitives, copies are made by the system. If we change the value of-
+//one of the original primitives, those will affect the outcome of only those, and not that of the copies, since
+//the copies are not aware of the changes. Hence two different results will come out.
```
rohni: Okay that is pass by value.  
What sort of values are passed by value? i.e. which ones are copied?
What sort of values are passed by reference?
```
let myobj = {
  x: 'abc',
  y: 1
};

// define a function that uses myobj
function double(obj) {
  obj.x = obj.x + obj.x;
  obj.y = obj.y + obj.y;
  return obj;
}
// call the double function with myobj
double(myobj);
// What are the current values of myobj.x, and myobj.y?

// and now with non compound data, i.e. strings and numbers
let str = 'something';
let num = 2;

// define a second double to work with non-objects
function doubleVal(val){
  return val + val;
}
// call the doubleVal function with our str and num variables.
double(str);
double(num);
// What will the current value of str and num be?
```
If you can answer these questions without using `console.log` you should be good to go.  If you can't experiment with console.log on the results of these function calls and the original values and see if it makes sense.

Week3/homework/6-step3.js
```diff
@@ -4,3 +4,4 @@ const arr2d = [[1, 2], [3, 4], [5, 6]];
 const arr3d = [[[1, 2], [3, 4]], [[5, 6], [7, 8]]];
 
 // add your solution here, or add a comment on how you would tackle this problem
+//Need help creating multidimensional arrays.
```
rohni: Okay, a multi-dimensional array is a fancy word for arrays in arrays.
So, let's say I have a normal array.
```
let arr1 = [1, 2, 3];
```
and I create another array
```
let arr2 = [3 4 5];
```
Both of these arrays are just normal arrays, you run from one end to the other and you have "simple" values (numbers in this case).
Now, what if you put them both in an array:
```
let arr3 = [arr1, arr2]
```
What does this look like?
Well it is just a normal array like any other, but instead of the values being "simple" values (numbers, strings, booleans etc.) the values are another array.  So it looks like this:
```
[arr1, arr2]
```
But, arr1 looks like this: `[1, 2, 3]` 
and arr2 looks like this: `[4, 5, 6]`
so if you get it all expanded arr3 looks like this:
```
[[1, 2, 3], [4, 5, 6]]
```
If you do a for loop over it and print out the results you will get:
```
[1, 2, 3]
[4, 5, 6]
```
So each element of the array instead of being a 1 or 2 or 3 etc. is an array.  So how do you get to the values inside these "inner" arrays?
You have to do exactly the same thing you were doing before (going through the array) but on the inner arrays.

The easiest way to do this sort of thing is called `recursion`.  It is a simple but very useful trick.  I'll create something that will go through and print out the values in the array that are not themselves arrays.
```
function printAllTheItems( arr ) {  
  let values = [];  // This is where we will collect all the values 
  for (let i = 0; i < arr.length; i++){  // typical for loop to traverse all the elements of the array

   // When we come to a new element, we have to ask ourselves what kind of value is it?  Is it another array?
  // If it is, then we need to repeat what we are doing in this function with this "new" array.  So we call the 
  // function again and pass in the current item i.e. arr[i] which is itself an array.
  // in the end our function will return a list of values and we need to catch that and add it to our current values
    if (Array.isArray(arr[i])) {  
      values = values.concat(printAllTheItems(arr[i]));  // setting our values to its current value plus 
                                                                                        //the return from the new call to printAllTheItems.
    } else { // if the current value is not an Array, we just add it to the values array. 
    }
  }
// Once we have made it through the whole array, including getting the values from the "sub" arrays, we return
// the values array.
  return values; 
};
```
I hope this makes it clearer what is happening.  All we do is go through the array, and every time we run into another array, we call the same function but with the sub array as input, and collect the results that come back.  The cool thing about this is that it will just keep on going from one Array to another and another until it runs out of arrays and collects all the non-array values and returns them.

Calling the above functions works with `arr2d` and `arr3d` and will work on an array that contains lots and lots of sub-arrays.  So it will work on something like this:
```
```
I'm not sure if this explanation has made things better or worse. 😉

PR #47, opened 2018-06-04
Homework week3

ThreePalmTrees:


Week3/homework/1-step3.js
```diff
@@ -1,11 +1,12 @@
-'use strict';
+"use strict";
 
 function foo(func) {
-  // What to do here? 
+  func();
+  console.log("This line will appear second");
 }
```
rohni: The results of calling `func()` should be returned from the function. `foo` knows nothing about the function it is calling, and cannot make assumptions that it can throw away the return value from `func`.

Week3/homework/2-step3.js
```diff
@@ -1,8 +1,37 @@
...
+    }
+    if (i % 5 === 0) {
+      divisibleBy5.push(i);
+      fiveCallback(i);
+    }
+  }
+
+  console.log(`Original array is: ${values}`);
+  console.log(`Numbers that can be divided by 3 are: ${divisibleBy3}`);
+  console.log(`Numbers that can be divided by 5 are: ${divisibleBy5}`);
```
rohni: 👍 Nice use of string templates here.

Week3/homework/2-step3.js
```diff
@@ -1,8 +1,37 @@
...
 function threeFive(startIndex, stopIndex, threeCallback, fiveCallback) {
   const values = [];
-  // Add your code here
+  const divisibleBy3 = [];
+  const divisibleBy5 = [];
+
+  if (startIndex > stopIndex) {
+    stopIndex = [startIndex, (startIndex = stopIndex)][0];
+  }
+
```
rohni: The idea here is nice, if I understand it correctly.
You are switching the start and stop indexes in case the start is lower than the stop, correct?

Your way of doing it here is pretty confusing, and I had to test it to see if it would actually work.  Kudos that it actually does.  But it is very hard to figure out what is happening here.  You are assigning a new value by putting the other value in an array, and then assigning a new value to the startIndex in the same array, and then taking that first value of the array. 
I can see why you are doing this, to hold your values temporarily, but I would *emphatically* suggest not doing this in the real world. 😉 
I can't think of a way to get around using a temporary variable, that is clean.
```
if (startIndex > stopIndex) {
  let tempStop = stopIndex;
  stopIndex = startIndex;
  startIndex = stopIndex
}
```
It ain't sexy, but it gets the job done and is still readable.
**But** extra credit for the very creative solution.  I kind of love it, even though I would never dare use it in the wild. 😆

Week3/homework/3-step3.js
```diff
@@ -1,25 +1,43 @@
...
+  let repeatedString = "";
+  for (let i = 0; i < num; i++) {
+    if (num < 0) {
+      repeatedString = "";
+    } else {
+      repeatedString += str;
+    }
+  }
+  return repeatedString;
 }
```

Week3/homework/3-step3.js
```diff
@@ -1,25 +1,43 @@
...
+  let repeatedString = "";
+  while (num > 0) {
+    repeatedString += str;
+    num--;
+  }
+  if (num < 0) {
+    repeatedString = "";
+  }
+  return repeatedString;
 }
```

Week3/homework/3-step3.js
```diff
@@ -1,25 +1,43 @@
...
+  let repeatedString = "";
+  do {
+    repeatedString += str;
+    num--;
+  } while (num > 0);
+  return repeatedString;
 }
-
-console.log('while', repeatStringNumTimesWithDoWhile('abc', 3));
+console.log("while", repeatStringNumTimesWithDoWhile("abc", -7));
```

Week3/homework/4-step3.js
```diff
@@ -1,2 +1,5 @@
-'use strict';
-// paste your freeCodeCamp solutions in here
+// 3.4 Some practice with objects:
+
+//     https://www.freecodecamp.com/challenges/construct-javascript-objects-with-functions
+
+// Link doesn't work
```
rohni: Fair enough.  I actually went to the site and searched for the `javascript`, `objects`, `function` and found a new page, but without a challenge.

Basically, you just needed to show that you can create an object using a function.

Week3/homework/5-step3.js
```diff
@@ -1,2 +1,17 @@
...
+    for (var j = 0; j < arr[i].length; j++) {
+      product *= arr[i][j];
+    }
+  }
+  // Only change code above this line
+  return product;
+}
+
+// Modify values below to test your code
+console.log(multiplyAll([[1, 2], [3, 4], [5, 6, 7]]));
```

Week3/homework/6-step3.js
```diff
@@ -1,6 +1,18 @@
...
+
+function printArr(arr) {
+  for (let i = 0; i < arr.length; i++) {
+    if (Array.isArray(arr[i])) {
+      printArr(arr[i]);
+    } else {
+      console.log(arr[i]);
+    }
+  }
+}
```
rohni: 👍 Excellent!  Nice use of recursion to get the job done.  
And I love that you created your own more complicated arr to really give your function a good test.  
Really excellent work.  You are the only one to have done this. 
In my work, when you have to figure this stuff out yourself, coming up with good test data is a creative challenge, and being hard on your code very important.
👏

Week3/homework/7-step3.js
```diff
@@ -23,3 +22,6 @@ console.log(y);
 
 // Add your explanation as a comment here
 
+// in the first function we're making the value of the parameter (val) increment by one every time we call the function, x is 9 (pass by value)
+// in the second function we're "hard coding" the value of the object key "x", x is 10 (pass by reference)
+// Primitive Data Types (number, string, boolean, etc..) are passed By Value, and Objects are passed By Reference.
```

Week3/homework/step4-bonus.js
```diff
@@ -1,10 +1,13 @@
...
+function getUnique(arr) {
+  const uniqueValues = arr.filter((elm, index) => arr.indexOf(elm) == index);
+  return uniqueValues;
+}
 
-// Replace `yourFunction` with the name of the function you just created
-const uniqueValues = yourFunction(values);
+const uniqueValues = getUnique(values);
 
 console.log(uniqueValues);
```
rohni: ✅ Excellent.  Good use of filter and higher-order functions to accomplish this.

Week3/homework/step4.js
```diff
@@ -1,6 +1,11 @@
...
+"use strict";
 
 // Add your code here
+function createBase(val) {
+  return function(base) {
+    return console.log((base += val));
+  };
+}
 
 const addSix = createBase(6);
```
rohni: ✅ Nice job here.  The only thing is I would return the value (`base += val`) from the function itself, and `console.log` the result.  Normally, you will want to do something with the value from a function, and use it in further calculations.  Assuming it is only to be displayed is not a good idea.  
Basically, this is the same problem as in the first exercise with `foo` and `func`. 😄

rohni:
Well Husam, 
Some interesting work here:
- the creative way to get around making a temp variable
- Really testing the recursiveness of your function

This sort of creative thinking is one of the most important parts of being a programmer, even though it is seldom mentioned.  Well Done!

Good luck with the rest of the course, and using your new skills out in the world.
All the best.
Rohan

ThreePalmTrees:
Thank you Rohan, your kind words are motivating and the review is very extensive and helpful.
I really appreciate the effort you're putting to help :)
Have a nice summer vacation :tropical_drink:

PR #48, opened 2018-07-17
week1 homework

end0620:


Week1/homework/app.js
```diff
@@ -1,11 +1,79 @@
...
+      let author = document.createTextNode(booksInfo[element].author)
+      titleHeader.appendChild(title);
+      authorHeader.appendChild(author);
+      languageHeader.appendChild(language);
+      li.appendChild(titleHeader);
+      li.appendChild(authorHeader);
+      li.appendChild(languageHeader);
+      ul.appendChild(li);
+    });
+    setImages();
```
JawharBairakdar: Hi @end0620, sorry to be late with another look at your homework but I noticed a thing you may use better.
here in line 75 you can pass the `element` to the `setImages` function instead of making another for loop there on a new Object you create, so I am expecting some changes :upside_down_face: !

JawharBairakdar: Hi @end0620, sorry to be late with another look at your homework but I noticed a thing you may use better.
here in line 75 you can pass the `element` to the `setImages` function instead of making another for loop there on a new Object you create, so I am expecting some changes :upside_down_face: !

PR #49, opened 2018-07-18
adding Shadi's homework

shadoo77:


Week1/homework/app.js
```diff
@@ -1,11 +1,137 @@
...
+        let bookCover = document.createElement("img");
+
+        h2Tag.appendChild(bookName);
+        h3Tag.appendChild(bookAuthor);
+        h4Tag.appendChild(bookLanguage);
+        h5Tag.appendChild(publishedDate);
+        let imageSrc = 'images/' + bookTitles[i] + '.jpg';
+        bookCover.setAttribute('src', imageSrc);
+        bookCover.setAttribute('alt', bookName);
+        newLi.innerHTML += h2Tag.outerHTML + h3Tag.outerHTML + h4Tag.outerHTML + h5Tag.outerHTML;
```
JawharBairakdar: it would be better to append these tags to the `li` tag you created `innerHTML` have some issues so you can use it but it's better to stay away of it, same mistake as [genegr88](https://github.com/HackYourFuture/JavaScript2/pull/52/files/1bf12e23c9fdd436cb486e4fe951bdbdb22e5465#diff-ddfdb57c4f6bf9f858f8da54c6c6a285R120)

Week1/homework/app.js
```diff
@@ -1,11 +1,137 @@
...
+        let h5Tag = document.createElement("h5");
+        let bookCover = document.createElement("img");
+
+        h2Tag.appendChild(bookName);
+        h3Tag.appendChild(bookAuthor);
+        h4Tag.appendChild(bookLanguage);
+        h5Tag.appendChild(publishedDate);
+        let imageSrc = 'images/' + bookTitles[i] + '.jpg';
+        bookCover.setAttribute('src', imageSrc);
+        bookCover.setAttribute('alt', bookName);
```
JawharBairakdar: `typeof bookName` is Object so that means the alternative text will be an object, which means you have this alt repeatedly `[object Text]`.
you can solve this by `bookName.textContent` or you may choose to change the content of the variable to be the existed `bookName`

PR #51, opened 2018-07-20
Week1 homework

sima-milli:


Week1/homework/app.js
```diff
@@ -1,11 +1,122 @@
...
+  }
+
+  function showBookInfo() {
+    for (let key in bookInfo) {
+      const h4 = createElement('h4');
+      h4.innerText = bookInfo[key]["title"];
+      getAndAppend(key, h4);
+      const p = createElement('p');
+      getAndAppend(key, p);
+      p.innerHTML = `<p>Author: ${bookInfo[key]["author"]} </p><p>Language: ${bookInfo[key]["language"]} </p>`;
```
JawharBairakdar: when ever you can use something else than `innerHTML` use it because this have a security issues you don't have to worry about that now but you should be aware of it

sima-milli: Thanks @JawharBairakdar for your feedback , I'll keep that in mind next time 👍

PR #52, opened 2018-07-20
Homework-week1

genegr88:


genegr88:
I hope I did everything correctly

Week1/homework/app.js
```diff
@@ -1,11 +1,139 @@
...
+function createBookList() {
+  let myMainDiv = document.getElementById('main');
+  let myH1 = document.createElement('h1');
+  myMainDiv.appendChild(myH1);
+  let pageTitile = document.createTextNode("My favorite books");
+  myH1.appendChild(pageTitile);
+  let myUl = document.createElement('ul');
+  myMainDiv.appendChild(myUl);
+  for (let i = 0; i < bookTitles.length; i++) {
+    let myLi = document.createElement('li');
```
JawharBairakdar: let's assume we created a variable here as the following:
```javascript
const key = bookTitles[i]
```
can you guess what is the value of it and how it can be useful?

Week1/homework/app.js
```diff
@@ -1,11 +1,139 @@
...
+    myUl.appendChild(myLi);
+    let bookName = document.createTextNode(Object.values(booksInformation)[i].title);
```
JawharBairakdar: if you guess it I think you know how you can use it here and instead of every `Object.keys` or `Object.values` in this `for` Loop :smiley:

Week1/homework/app.js
```diff
@@ -1,11 +1,139 @@
...
+    let bookAuthor = document.createTextNode("The book is written by " + Object.values(booksInformation)[i].author);
+    let booklanguage = document.createTextNode("The book is written in " + Object.values(booksInformation)[i].language);
+    let myH2 = document.createElement("h2");
+    let myPtag1 = document.createElement("p");
+    let myPtag2 = document.createElement("p");
+    myH2.appendChild(bookName);
+    myPtag1.appendChild(bookAuthor);
+    myPtag2.appendChild(booklanguage);
+    myLi.innerHTML += myH2.outerHTML + myPtag1.outerHTML + myPtag2.outerHTML;
```
JawharBairakdar: it would be better to append these tags to the `li` tag you created `innerHTML` have some issues so you can use it but it's better to stay away of it.

PR #55, opened 2018-07-21
Homework js2 week1

WardMad:


Week1/homework/app.js
```diff
@@ -1,11 +1,103 @@
...
+ 
+function showBook() {
+
+for (let [k, z] of Object.entries(bookInfo)){
+
+  
+  
+
+    let header_2 = document.createElement("H2");
+    let textnode = document.createTextNode(k + " - " +  z.title + "  " + z.language  + "  " + z.author);
```
JawharBairakdar: I agree with you `Object.entries` is a very useful feature in Javascript, But what if you get this situation that `bookInfo` returned from a server an empty object, it doesn't have anything?
in this case you have `[k, z]` and both of them are undefined and in this line exactly where you will got an error back that says __`TypeError: z.title is undefined`__ and if you are expecting a function inside of this your whole app will just stop.

Also no problem using `Object.entries` But as a person of people for the first time I will just look to the code and be confused of the new pattern I see, and `Object.keys` I am not talking about taste :upside_down_face: but I am talking about legibility *(sometimes it looks better)*, so following the mostly used pattern is very useful on the low level as a homework and on the high level as a worldwide of developers.

so instead you can use `Object.keys` as the following: 
```javascript
function myBeautifulFunction() {
  const bookInfoKeys = Object.keys(bookInfo);
  // for avoiding errors
  if (!bookInfoKeys.length) return;
  for(let key of bookInfoKeys){
    const title = key.title;
    const author = key.author;
    // etc...
  }
}
```

`Object.entries` way: 
```javascript
function myBeautifulFunction() {
  const bookInfoKeys = Object.entries(bookInfo);
  // for avoiding errors
  if (!bookInfoKeys.length) return;
  for(let [k, z] of bookInfoKeys){
    // keys are there but the values are empty we have to check again
    if (!z.title) return;
    // what about human readable names?
    // Your code...
  }
}
```

Week1/homework/app.js
```diff
@@ -1,11 +1,103 @@
...
+    let book_title = document.createTextNode(z.title );
+        let book_author = document.createTextNode( "\n" + z.author );
+    
+    list.appendChild(textnode);
+    header_2.appendChild(book_title);
+     header_2.appendChild(book_author);
+    
+ list.innerHTML += header_2.outerHTML
```
JawharBairakdar: it would be better to append these tags to the `li` tag you created `innerHTML` have some issues so you can use it but it's better to stay away of it.

Week1/homework/style.css
```diff
@@ -1 +1,60 @@
...
+    font-size: 16px;
+    margin: 4px 2px;
+    width:20%;
+    transition-duration: 0.4s;
+    cursor: pointer;
+}
+
+button {
+    background-color: white; 
+    color: black;
```
JawharBairakdar: well it's not a Css class but why are you overwriting it :stuck_out_tongue_winking_eye: ?

PR #57, opened 2018-07-22
js2 Homework Week 1

KaramGhazzi:


JawharBairakdar:
Hi @KaramGhazzi, thank you for submiting your homework, you did a very greatful job, but there is a one single Main point in your homework:
As you probebly saw in the [homework](https://github.com/HackYourFuture/JavaScript2/blob/master/Week1/MAKEME.md) step 2.4 that it says don't use arrays; the reason of that is to give you as a student the enough space to practice on the Objects because it's one of the most commonly using in your next job as a web developer even if you changed the language you are working with you will stay use it, going back to our point what you have done is to make an object that works almost exactly like an array and act on it the homework, yes it's works fine if you did it with an array with changing on word same result you will got, so I want to see this Homework with a something defferent than the object you created.

something else I will post it soon in clas16 channel, for more detailes [`.innerHTML`](https://github.com/HackYourFuture/JavaScript2/pull/56#discussion-diff-207695136R108)

JawharBairakdar:
Hi @KaramGhazzi thanls for reply,
Yes you are using an Object, But I have a qusetion I will answer it my self *(it's a bit tricky)*

 - what is the keys of an array?
    - it's the indexes
 - what is your keys in the Object you have?
    - it's the indexes or maybe something similler to the indexes numbers *(same order and count)*

in this way what you are providing is an object acting like an array, so your for loop is as the following:
```javascript
const myObj = {
  0: 'hello',
  1: 'world'
}
for (let i = 0; i < Object.keys(myObj).length; i++) { 
  console.log(myObj[i]) // it's correct as long as the object has a keys same as the indexes
}
// same as:
const myArr = {
  'hello',
  'world'
}
for (let i = 0; i < Object.keys(myObj).length; i++) { 
  console.log(myArr[i]) // it takes the indexes and loop throgh them one next to the other
}
```
But what if you got a strings as a keys? like the following:
```javascript
const myObj = {
  "greeting": "hello",
  "to": "world"
}
for (let i = 0; i < Object.keys(myObj).length; i++) { 
  console.log(myObj[i]) // this won't work !!
  // your homework is to solve this when ever you solve it you can then solve the homework it self (almost same idea)
}
```

JawharBairakdar:
Looking Forward to see the changes on it's branch with a clean Repository branches on thursday 😃

PR #60, opened 2018-07-23
homework2 finished

end0620:


Week2/homework/map_filter.js
```diff
@@ -3,3 +3,7 @@
 const numbers = [1, 2, 3, 4];
 
 // Add your code here
+const newNumbers = numbers.filter(x => x %2 ).map(x => x * 2);
+
+console.log(`the double numbers are`, newNumbers);
```
harditsingh: You should avoid using backticks in this console log since you are not using any placeholders. Use normal quotes/double-quotes instead.

Week2/homework/maartjes_work.js
```diff
@@ -45,3 +45,17 @@ const tuesday = [
 const tasks = monday.concat(tuesday);
 
 // Add your code here
+const tasksInHours = tasks.map(function(x){
+  let y = x;
```
harditsingh: Try to give meaningful names to your variables, it helps a lot with readability.

PR #64, opened 2018-07-25
Homework week2p1

acimanx:


Week2/homework/maartjes_work.js
```diff
@@ -1,47 +1,17 @@
...
+  { name: "Some more web development", duration: 180 },
+  { name: "Staring out the window", duration: 10 },
+  { name: "Talk to a lot of people", duration: 200 },
+  { name: "Look at application assignments new students", duration: 40 }
 ];
-
 const tasks = monday.concat(tuesday);
-
-// Add your code here
+const salaryCounters = `€ ${tasks.filter(less120 => 120 <= less120.duration).map(hourlyRate => 20 * (hourlyRate.duration / 60)).reduce((a, b) => a + b).toFixed(2)}`;
```
NouranMahmoud: I just have a comment about naming your parameters e.g. `filter(less120 => ...)` and `map(hourlyRate => ...)`. Names of parameters should be meaningful for what the function will receive so for filter it will receive `task` not `less120` also for map it will receive `task` also not `hourlyRate`.

NouranMahmoud:
By the way, I like how you thought out of the box and used filter then map then reduce. Good Job! 👍

PR #70, opened 2018-07-27
My Homework W2

Donal86:


Week2/homework/maartjes_work.js
```diff
@@ -45,3 +45,8 @@ const tuesday = [
 const tasks = monday.concat(tuesday);
 
 // Add your code here
+
+const maartjeEarns = tasks.map(element => element.duration / 60).filter(element => element >= 2).map(element => element * 17.50).reduce((total, element) => total + element, 0);
```
NouranMahmoud: Same as KaramGhazzi's homework, 

You can combine last .map to the .reduce like following

```
const maartjeEarns = tasks.map(element => element.duration / 60).filter(element => element >= 2).reduce((total, element) => total + (element * 18.70), 0);


```

Donal86: OK thank you, it is just because we studied together and we did the homework, thanks for you suggestion I will edit it and I will keep in my mind that I have to find out how to do the homework in a different way next time.

PR #71, opened 2018-07-28
gaffari js2week2

Gaffari02:
i# Please enter the commit message for your changes. Lines starting

### DIFF
--- a/Week2/homework/maartjes_work.js
+++ b/Week2/homework/maartjes_work.js
@@ -45,3 +45,7 @@ const tuesday = [
 const tasks = monday.concat(tuesday);
 
 // Add your code here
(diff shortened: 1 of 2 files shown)


Week2/homework/maartjes_work.js
```diff
@@ -45,3 +45,7 @@ const tuesday = [
 const tasks = monday.concat(tuesday);
 
 // Add your code here
+const salary = tasks.map(x => x.duration / 60).filter(x => x >= 2)
```
NouranMahmoud: You can combine .map and .reduce together like following

```
const salary = tasks.map(x => x.duration / 60).filter(x => x >= 2)
.reduce((acc, cur) => acc + cur*10);
```

NouranMahmoud:
Use meaningful naming instead of x you can name it "task"

PR #72, opened 2018-07-29
add reduce map filter

WardMad:


Week2/homework/maartjes_work.js
```diff
@@ -1,47 +1,57 @@
...
+    },
+    {
+        name: 'Look at application assignments new students',
+        duration: 40
+    }
 ];
 
-const tasks = monday.concat(tuesday);
+const tasks = monday.concat(tuesday)
+let minutesToHours = tasks.filter(a => a.duration >= 120).map(h => h.duration / 60);
```
harditsingh: You mixed up the order of first 2 steps here. While that does not matter much in this case, you should read the instructions carefully.

Week2/homework/maartjes_work.js
```diff
@@ -1,47 +1,57 @@
...
 
-// Add your code here
+console.log(minutesToHours);
+
+let perHour = minutesToHours.map(v => v * 23) // per hour 23 euro
```
harditsingh: You should always have meaningful names for the variables, 'v' is not an intuitive name.

Week2/homework/maartjes_work.js
```diff
@@ -1,47 +1,57 @@
...
+
+console.log(perHour);
+
+let total = perHour.reduce(function(a, b) {
```
harditsingh: You should have used '=>' function here instead of 'function()', that's what the assignment mentioned.

Week2/homework/map_filter.js
```diff
@@ -2,4 +2,7 @@
 
 const numbers = [1, 2, 3, 4];
 
-// Add your code here
+
+const newNumbers = numbers.filter(w => w % 2 !== 0).map(x => x * 2);
```
harditsingh: Another example of non-intuitive naming :)

PR #75, opened 2018-07-31
my js2 week3 homework

marwanaboudaoud:


Week3/homework/6-step3.js
```diff
@@ -4,3 +4,15 @@ const arr2d = [[1, 2], [3, 4], [5, 6]];
 const arr3d = [[[1, 2], [3, 4]], [[5, 6], [7, 8]]];
 
 // add your solution here, or add a comment on how you would tackle this problem
+let items = '';
+for (let i = 0; i < arr3d.length; i++) {
+    for (let y = 0; y < arr3d[i].length; y++) {
+        for (let k = 0; k < arr3d[i][j].length; k++) {
+            itemsOfArray += arr3d[i][j][k] + ' ';
```
JawharBairakdar: it's a nice thing that we can put a for loop statement inside of others.
Well this question is a kind of challenge of your understanding the tools you have so far!
And as you probably noticed it says:
>What if you didn't know how deep the array was nested? (You don't have to write code for this but think about it.)

so because of that I guess it's a kind of Okay to tell the tools you have and how to deploy them all together.

And because we are expecting a function that will go to the last item it doesn't matter how the array in depth it's.
so and here how you can solve it.
```javascript
function sumInDepth(arr) {

  let sumArrayValues = null;

  const loopInAndSum = childArray => {
    for(let i = 0; i < childArray.length; i++){
      let valueOf_childArray = childArray[i];
      if (Array.isArray(valueOf_childArray)) {
        loopInAndSum(valueOf_childArray);
      } else { // here we can do our logic with the last level of the array
        sumArrayValues += valueOf_childArray;
      }
    }
  }

  loopInAndSum(arr);
  return sumArrayValues;
}
sumInDepth(arr3d) // 36
```
BTW `j` is Not defined & `itemsOfArray` is not defined

Week3/homework/7-step3.js
```diff
@@ -21,5 +21,10 @@ f2(y);
...
 
-// Add your explanation as a comment here
-
+//In Pass by Reference, Function is called by directly passing 
+//the reference/address of the variable as the argument.
+//Changing the argument inside the function affect the variable
+//passed from outside the function. In Javascript objects and 
+//arrays follows pass by reference.
+//so if we are passing object or array as an argument to the method,
+//then there is a possibility that value of the object can change.
```
JawharBairakdar: did you meant that, by changing the parameter inside the function I am effecting the parameter directly but by changing a reference inside of it I am connecting the reference directly?

PR #76, opened 2018-07-31
Shadi's homework week3 Javascript2

shadoo77:


Week3/homework/step4-bonus.js
```diff
@@ -3,8 +3,24 @@
 const values = ['a', 'b', 'c', 'd', 'a', 'e', 'f', 'c'];
 
 // Add your function here. Try and come up with a good name for this function
-
+function uniqueElement(myArr) {
+    const els = myArr.map(el => el);
```
JawharBairakdar: I was trying to maybe improve this function a little more well it needs think, but this line right here is just creating a new array by `.map` which is not it's work, you can create a one simply by `myArr.slice()` or `[...myArr]`

Week3/homework/2-step3.js
```diff
@@ -1,8 +1,30 @@
...
 function threeFive(startIndex, stopIndex, threeCallback, fiveCallback) {
   const values = [];
-  // Add your code here
+  const temp = stopIndex - startIndex;
+  const outputArr = [];
+  for (let i = 0; i <= temp; i++) {
+    values[i] = startIndex;
+    startIndex++;
+  }
+  const results = values.filter(value => value % 3 === 0 || value % 5 ===0);
```
JawharBairakdar: actually you can use `values.forEach` immediately because you are checking there the equality again right?

Week3/homework/3-step3.js
```diff
@@ -3,6 +3,15 @@
...
 function repeatStringNumTimesWithFor(str, num) {
   // add your code here
+  let x = "";
+  if (num <= 0) {
+    str = "";
+  } else {
+    for (let i = 0; i < num; i++) {
+      x += str;
+    }
+    str = x;
```
JawharBairakdar: In my re-reviewing your homework you don't need this `x` you can do all things with the `str` it self

shadoo77: Oh ... yes you have right 😅😅

PR #77, opened 2018-08-01
Homework js2 week3

WardMad:


Week3/homework/3-step3.js
```diff
@@ -1,25 +1,61 @@
...
-  // add your code here
-  return str;
+'use strict'
+
+function repeatStringNumTimes(str, num) {
+    var repeat = "";
+
+    for (var i = 0; i < num; i++) {
+
+        var x = repeat += str;
```
JawharBairakdar: In my re-reviewing your homework, you don't need this `x` variable here, you can do't with the `repeat` variable it self

PR #79, opened 2018-08-02
Week3 homework

sima-milli:


Week3/homework/7-step3.js
```diff
@@ -23,3 +27,5 @@ console.log(y);
 
 // Add your explanation as a comment here
 
+//in f1 the x is assigned using const which means it is immutable.
```
JawharBairakdar: Well let's first open the console and try change a `const` variable what do you see?
Error right? :upside_down_face: 

well actually what is happening exactly is the constant variable outside is taking a place in the memory.

when ever you pass it to the function as an argument the function will receive it inside as a parameter taking a new place there, the variable will have whether a direct value or a reference value the direct one you can play around it inside the function how ever you want *(in it's new memory space)*, the reference when ever you changed it, it changes every where.

pass-by-reference-vs-pass-by-value-animation

PR #80, opened 2018-08-02
js2-week1-homework

mehmetsaidcoskun:


mehmetsaidcoskun:
The link which is written in `README.md` file leads to only the `master` branch but all my work is in the `js2-week1` branch. That's why the web page I created does not seem.
Because of this, I also added the homework of week1-JS2 to another repository just for displaying it.
Here is the link: https://mehmetsaidcoskun.github.io/hyf-html-css/JS2-week1/

Week1/homework/app.js
```diff
@@ -1,11 +1,114 @@
...
+      myDiv.appendChild(img);
+
+      img.setAttribute("src", Object.values(cover)[a]);
+
+      img.setAttribute("alt", myObj[Object.keys(myObj)[a]].title);
+
+      for (let k = 1; k < 5; k++) {
+        let li = document.createElement('li');
+        ul.appendChild(li);
+        li.innerHTML = (Object.keys(myObj[Object.keys(myObj)[a]])[k] + ": " + Object.values(myObj[Object.keys(myObj)[a]])[k]);
```
JawharBairakdar: I think here is the best example I can introduce for you about the important of a meaningful names.
In your code in best cases with the names you have we can improve it by order like so:
```javascript
li.innerHTML = (
  Object.keys(
      myObj[
        Object.keys(myObj)[a] // Magic 👻
      ]
  )[k] + ": " + Object.values(
      myObj[
        Object.keys(myObj)[a]
      ]
  )[k]
);
```
And it still Not a pattern and obviously not something by the first look as a developer legible.
so what you can do is:
 - to define the `Object.keys` with a meaningful name like "booksNames" *(you may have a better name)* before the first for loop even *(you used it multiple times)*
 - then use it everywhere you are using it now

even if this one step didn't fix that much but you will still have cleaner code and more readable.

Week1/homework/app.js
```diff
@@ -1,11 +1,114 @@
...
+    "what_men_live_by": "https://images-na.ssl-images-amazon.com/images/I/412JkTATtBL._SX331_BO1,204,203,200_.jpg",
+    "the_alchemist": "https://images.gr-assets.com/books/1483412266l/865.jpg"
+  };
+
+
+
+  function printBookDetails(placeId, myObj, cover) {
+    let div = document.getElementById(placeId);
+
+    function size(obj) {
```
JawharBairakdar: I liked the `size` function and it's implementing, and I didn't think for a once of counting the object size like so.
But as the next comment says if you get the `Object.keys` till here you can use `booksNames.length`.
if this function was only for practice! Keep the great work :muscle:, but in a separated file or branch, because your homework is our reference let's say on your understanding right?

mehmetsaidcoskun:
thank you @JawharBairakdar , as soon as possible I will make changes according to your comments.

PR #81, opened 2018-08-02
js2-week2 homework

mehmetsaidcoskun:


Week2/homework/maartjes_work.js
```diff
@@ -45,3 +45,34 @@ const tuesday = [
 const tasks = monday.concat(tuesday);
 
 // Add your code here
+
+
+//Step-1: Map the tasks to durations in hours.
+const workHour = tasks.map(myElement =>
+  (Math.round(100 * (myElement.duration / 60))) / 100);
```
JawharBairakdar: same result.
```javascript
// Yours
(Math.round(100 * (myElement.duration / 60))) / 100
// Mine 😃
Math.floor(myElement.duration / .6) / 100
```
.6 is used to show duration in hour.
100 is used to round with two decimals such as 0.33 or 1.23

Week2/homework/maartjes_work.js
```diff
@@ -45,3 +45,34 @@ const tuesday = [
...
+const morethan2Hours = workHour.filter(hour => hour >= 2);
+// durations which are less than 2, are gone.
+
+
+//Step-3: Multiply the each duration by a per-hour rate for billing
+//(you can decide yourself what Maartje should earn per hour) 
+//and sum it all up.
+const totalHour = morethan2Hours.reduce((per, total) => total += per);
+
+const earningAmount = `€ ${Math.round((totalHour * 24.5) * 100) / 100}`;
```
JawharBairakdar: So far you are doing very well, But here became the confusion.
when ever you have a hard coded numbers means something you should at least put a comment says what these numbers are!
or you may put each in a variable making them easier to understand like: 
```javascript
const SALARY = 24.5
const earningAmount = `€ ${Math.round((totalHour * SALARY) * 100) / 100}`; 
```
Actully I didn't know what the other are but i think they are useful :nerd_face:  :100:

mehmetsaidcoskun:
thank you @JawharBairakdar  for your precious comments :) the most important thing I have to improve is naming I guess. I will put effort into it :) see you
By the way, thanks to you and js3 homework, I can use addEventListener better. Thank you very much :)

Week2/homework/maartjes_work.js
mehmetsaidcoskun: the both of them can be used, right? 
Math.round(myElement.duration / .6) / 100
Math.floor(myElement.duration / .6) / 100
They look like each other :)

Week2/homework/maartjes_work.js
mehmetsaidcoskun: you are absolutely right, the numbers are meaningful in my mind but it might confuse. After that, I will define new variables for each information. Thank you :)

Week2/homework/maartjes_work.js
JawharBairakdar: Yes exactly you can use any of them both, the only deffirent is floor always giving you the nearset lower value, nor round giving you the nearest value of course depeniding on the case you have. 👍

PR #84, opened 2018-08-04
Barbaros JS2Week3

whitepiolin:


Week3/homework/3-step3.js
```diff
@@ -2,23 +2,43 @@
 
 // use a 'for' loop
 function repeatStringNumTimesWithFor(str, num) {
-  // add your code here
+  if (num < 1) {
+    return "";
+  }
+
+  let myStr = str;
```
harditsingh: Try to use variable names which actually describe what the variables are actually being used for. Names like 'mySomething' can be misleading and hence lead to confusing code

Week3/homework/3-step3.js
```diff
@@ -2,23 +2,43 @@
...
+  }
   return str;
 }
 
 console.log('for', repeatStringNumTimesWithFor('abc', 3));
 
 // use a 'while' loop
 function repeatStringNumTimesWithWhile(str, num) {
-  // add your code here
+  if (num < 1) { return "";}
```
harditsingh: You should space the code out properly, this helps in making the code look cleaner and improves readability.

Week3/homework/6-step3.js
```diff
@@ -3,4 +3,15 @@
...
 // add your solution here, or add a comment on how you would tackle this problem
+
+for (var i = 0; i < arr3d.length; i++) {
+    for (var j = 0; j < arr3d[i].length; j++) {
+        for (var k = 0; k < arr3d[j].length; k++) {
+            
+            console.log(arr3d[i][j][k]);
+        }
+    }
+}
```
harditsingh: You should also mention in this solution what you would do about arrays bigger than 3 dimensions.

Week3/homework/7-step3.js
```diff
@@ -22,4 +22,8 @@ f2(y);
 console.log(y);
 
 // Add your explanation as a comment here
+/*
+In the f1 function we use x which is assigned before as immutable. 
+But in f2, we see that object assigned to a variable can be mutabled.
```
harditsingh: Good! Do try to explain more about the binding maybe?

PR #97, opened 2018-08-09
Js2 week3

Donal86:


Week3/homework/1-step3.js
```diff
@@ -1,11 +1,12 @@
 'use strict';
 
 function foo(func) {
-  // What to do here? 
+  setTimeout(function () { bar() }, 5000);
```
JawharBairakdar: I am not pretty sure about what are you trying to do here, in all cases this will work fine after 5 seconds, but what if you have `bar` function in another file you can't pass it like so!

the only solution for this is pass it by arguments.

Week3/homework/2-step3.js
```diff
@@ -2,7 +2,23 @@
...
 function threeFive(startIndex, stopIndex, threeCallback, fiveCallback) {
   const values = [];
-  // Add your code here
+  for (startIndex; startIndex <= stopIndex; startIndex++) {
+    values.push(startIndex);
+  }
+  for (const i in values) {
+    if (values[i] % 3 === 0) {
+      threeCallback(values[i]);
+    } else values[i] % 5 === 0;
```
JawharBairakdar: it's working but where is the `if` statement?
And what if you have a condition that you will see divisible on 5 and 3 like 15 or 30.

Week3/homework/3-step3.js
```diff
@@ -1,8 +1,13 @@
 'use strict';
 
 // use a 'for' loop
+
 function repeatStringNumTimesWithFor(str, num) {
   // add your code here
+  if (num <= 0) return "";
+  const repeat = str;
+  for (let i = 0; i < num - 1; ++i) // why repeat 4 times ??
```
JawharBairakdar: it's repeating 4 times because of the 0 index start :slightly_smiling_face: .

BTW where is the brackets `{ }`?

Week3/homework/3-step3.js
```diff
@@ -11,6 +16,10 @@ console.log('for', repeatStringNumTimesWithFor('abc', 3));
 // use a 'while' loop
 function repeatStringNumTimesWithWhile(str, num) {
   // add your code here
+  if (num <= 0) return "";
+  const repeat = str;
+  while (num > 1)
+    str += repeat, num--;
```
JawharBairakdar: there is no need for the `if` statement, because in any case while number is a truthy *(NOT `0`)* iterate through it.
And the next one...

Week3/homework/6-step3.js
```diff
@@ -4,3 +4,10 @@ const arr2d = [[1, 2], [3, 4], [5, 6]];
 const arr3d = [[[1, 2], [3, 4]], [[5, 6], [7, 8]]];
 
 // add your solution here, or add a comment on how you would tackle this problem
+for (let i = 0; i < arr3d.length; i++) {
+  for (let j = 0; j < arr3d[i].length; j++) {
+    for (let k = 0; k < arr3d[i][j].length; k++) {
+      console.log(arr3d[i][j][k]);
```
JawharBairakdar: it's a good job, this is solving this specific problem But what if got a deeper array would each time think how to solve it?
Also it's not working on `arr2d`.

Week3/homework/7-step3.js
```diff
@@ -22,4 +22,5 @@ f2(y);
 console.log(y);
 
 // Add your explanation as a comment here
-
+// in f1 x is a variable (there is no change of the function) 9.
+// in f2 x is an object inside the variable y (there is a change of the function) Object { x: 10 }.
```
JawharBairakdar: I couldn't really understand, could you please explain more?

PR #98, opened 2018-08-12
week1 done

Samtabanja:
again !


Week1/homework/app.js
```diff
@@ -2,10 +2,157 @@
...
+
+  function listGenerator() {
+    let main = document.getElementById('content');
+    let list = document.createElement('ul');
+    main.appendChild(list);
+
+    for (let i = 0; i < array; i++) {
+      let listChild = document.createElement('li');
+      list.appendChild(listChild);
+
```
JawharBairakdar: let's assume we created a variable here as the following:
```javascript
const key = bookTitles[i]
const book = Object.keys(books)[key]
```
can you guess what is the value of it and how it can be useful?

> if you guess it I think you know how you can use it here and instead of every `Object.keys` or `Object.values` in this `for` Loop :smiley:.

Week1/homework/app.js
```diff
@@ -2,10 +2,157 @@
...
+      let text = document.createTextNode('Book name  ' + Object.values(books)[i].title);
```
JawharBairakdar: because you have a use of `Object.values(books)[i]` in a multi places you can just assign it to a variable before all of these code.

Or you can just work with the previous comment :slightly_smiling_face:

Week1/homework/app.js
```diff
@@ -2,10 +2,157 @@
...
+      listChild.appendChild(h1);
+      h1.appendChild(text);
+
+      let text2 = document.createTextNode('Writer  ' + Object.values(books)[i].author);
+      let h2 = document.createElement('h2');
+      listChild.appendChild(h2);
+      h2.appendChild(text2);
+
+      let text3 = document.createTextNode('Language ' + Object.values(books)[i].language);
+      let anotherH2 = document.createElement('h2');
```
JawharBairakdar: `anotherH2` doesn't make any sense for a person who is reading your code!
it's a heading right?
so what you may call the heading `id` is css? something make sense right so it would be very nice to read a code like a story :slightly_smiling_face:

PR #99, opened 2018-08-12
week2 done

Samtabanja:
again !


Week2/homework/maartjes_work.js
```diff
@@ -44,4 +44,24 @@ const tuesday = [
...
 const tasks = monday.concat(tuesday);
 
-// Add your code here
+// Map the tasks to durations in hours.
+const durationsToHours = tasks.map((hours) => hours.duration / 60);
+console.log(durationsToHours);
+
+// Filter out everything that took less than two hours(i.e., remove from the collection)
+
+const moreHours = durationsToHours.filter((x) => x > 1);
```
JawharBairakdar: Oh you are Filtering out the less than 1 :upside_down_face:.
BTW `moreHours` I think it's a meaningful name for you as the developer who written it, for others it would be so confusing if this `moreHours` transported to other file without the main `tasks` array or so.

PR #102, opened 2018-08-15
Js2 week3 homework

salar01:


Week3/homework/2-step3.js
```diff
@@ -1,8 +1,32 @@
...
+    values.push(i);
+  }
+  console.log(values);
+
+  values.forEach(num => {
+    if (num % 3 === 0) {
+      threeCallback(num);
+    }
+    if (num % 5 === 0);
+    {
```
JawharBairakdar: the Brackets `{...}` should be in the same line of the `if` statement plus the `if` statement doesn't have a semicolon

PR #103, opened 2018-08-15
updated week1 Homework

Gaffari02:


Week1/homework/app.js
```diff
@@ -1,11 +1,163 @@
...
 
+ function makeList() {
+  let bookcontainer = document.getElementById("mybooks");
+  let div = document.createElement("div");
+  bookcontainer.appendChild(div);
+  
+
+  for (let i = 0; i<bookTitles.length; i++) {
+    
+    let contentTitle = document.createTextNode(bookDetails[bookTitles[i]].title);
```
JawharBairakdar: I saw you are using `bookDetails[bookTitles[i]]` in a multiple places with this long format.
it could more readable if you defined it in a variable at the beginning of your for loop and using it everywhere else.

Week1/homework/app.js
```diff
@@ -1,11 +1,163 @@
...
+    li.appendChild(liAuthor);
+    li.appendChild(img);
+    img.src = Object.values(bookImg)[i].adress;
+
+    li.setAttribute("class", "bookList");
+    img.setAttribute("class", "bookImage")
+    liTitle.setAttribute("class", "detailsLi");
+    liLanguage.setAttribute("class", "detailsLi");
+    liAuthor.setAttribute("class",  "detailsLi");
+    div.appendChild(li);
```
JawharBairakdar: `li` tag inside of a `div` tag is not a valid syntax the `li` only valid suing inside of `ul` or `ol` tags.

Week1/homework/app.js
```diff
@@ -1,11 +1,163 @@
...
+    let detailsAuthor = object.values(bookDetails) [i].author;*/
+    
+    liTitle.appendChild(contentTitle);
+    liLanguage.appendChild(contentLanguage);
+    liAuthor.appendChild(contentAuthor);
+    li.appendChild(liTitle);
+    li.appendChild(liLanguage);
+    li.appendChild(liAuthor);
+    li.appendChild(img);
+    img.src = Object.values(bookImg)[i].adress;
```
JawharBairakdar: you don't have to use `Object.values` taking the values of `bookImg` you can do so by the `bookTitles[i]` you have, because it's same keys as `bookImg` Right?

Week1/homework/app.js
```diff
@@ -1,11 +1,163 @@
...
+   "Het diner" : {
+    title :     "Het diner",
+    language : "Dutch",
+    Author : "Herman Koch",
+    
+  }
+ }
+ //console.log(bookDetails);
+
+ const bookImg = {
```
JawharBairakdar: you have a very nice `base64` images, but you should be aware of legibility, it takes a very long line which means with `wrap` option in a Code editor it's long scrolling, and and object ruiner :boom:, don't do that again :slightly_smiling_face: unless you have the url address for these images.

PR #105, opened 2018-08-30
added link events

gijscor:
added helpful link for understanding what events are


gijscor:
@remarcmij no that wasn't on purpose.  I'll change this, thanks for the heads up

PR #114, opened 2018-09-24
Week1

kawaraa:
Here is the homework week1, Looking forward to here from you!

### DIFF
--- a/Week1/homework/app.js
+++ b/Week1/homework/app.js
@@ -1,11 +1,88 @@
 'use strict';
 {
+  // Step 1
   const bookTitles = [
-    // Replace with your own book titles
-    'harry_potter_chamber_secrets'
+    "suhuf",
+    "tawrat",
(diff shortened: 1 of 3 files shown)


kawaraa:
Thanks for your comments! I have Implemented it.

Week1/homework/app.js
```diff
@@ -1,11 +1,53 @@
...
+    if (nodes.id) {
+      elem.setAttribute(nodes.id, nodes.val);
+    }
+    return elem;
+  }
 
-  // Replace with your own code
-  console.log(bookTitles);
+  const ul = createEl("ul", root);
+  for (let i of bookTitles) {
```
remarcmij: If you are looping through `bookTitles`, a single element of that array would be a `bookTitle` (or simply `title`). So that should be the name for the loop variable, instead of `i`.

PR #116, opened 2018-09-27
Week1

Mohamad-AezAldin:
please see my homework for week1


Week1/homework/app.js
```diff
@@ -2,10 +2,130 @@
...
+    'geek_love',
+    'gilead',
+    'disgrace',
+    'infinite_jest',
+    'lolita',
+    'maus',
+    'poems',
+    'things_fall_apart'
   ];
+  console.log(bookTitles);
```
remarcmij: You can use this array to loop through the properties of your `bookDetails` object. See below.

Week1/homework/app.js
```diff
@@ -2,10 +2,130 @@
...
+  };
+  const arr = [];
+  for (const j in booksCovers) {
+    arr.push(j);
+  }
+  console.log(arr);
+
+  function createImg(val) {
+    for (const j in val) {
+      const getId = document.getElementById(j);
```
remarcmij: `val` and `j` are meaningless names. What do these variables hold? Once you have identified that you also know the best name for them. Without a detailed inspection of your code I can only guess what they represent. I always image myself that I'm writing code for an audience, not for myself. Can others instantly understand what I'm doing? If not, I need to come up with better names to more clearly express my intentions. 

I'm also not fond of this name `getId`. The use of a verb in a variable name suggest an action that I would associate with a function rather than a value. In this case the variable does not hold an 'id'. It actually holds a reference to an `li` element. I had to look hard to find that. A more accurate name could have saved me that trouble.

Week1/homework/app.js
```diff
@@ -2,10 +2,130 @@
...
+  console.log(booksDetails);
+  for (const key in booksDetails) {
+    console.log(key);
+    const x = booksDetails[key];
+    console.log(x);
+
+    for (const prop in x) {
+      console.log(prop + " : " + x[prop]);
+    }
+  }
```
remarcmij: A PR is an explicit request to others to review your code. This code is for debugging. It should not be in a PR. You don't want to burden reviewers with having to find out if code is relevant for the functioning of an application. If you want to keep this code for your own purposes, that's fine. Then for the homework you could create a new branch in which you remove all this debug code and use that for the PR.

Week1/homework/app.js
```diff
@@ -2,10 +2,130 @@
...
+
+  // 2.5
+
+  function createAndAppend(details) {
+    const h1 = document.createElement("h1"),
+      ul = document.createElement("ul"),
+      root = document.getElementById("root");
+    h1.innerHTML = "10 Books to read before you die";
```
remarcmij: Defining multiple variables in a single declaration was done a lot in the past with the older versions of JavaScript. Nowadays it is considered a best practice to have one variable declaration per line, e.g.:

```js
 const h1 = document.createElement("h1")
const  ul = document.createElement("ul"),
const root = document.getElementById("root");
```

Week1/homework/app.js
```diff
@@ -2,10 +2,130 @@
...
+    console.log(x);
+
+    for (const prop in x) {
+      console.log(prop + " : " + x[prop]);
+    }
+  }
+
+  // 2.5
+
+  function createAndAppend(details) {
```
remarcmij: I would have chosen the name `books` rather than `details` for this parameter.

Week1/homework/app.js
```diff
@@ -2,10 +2,130 @@
...
+    const h1 = document.createElement("h1"),
+      ul = document.createElement("ul"),
+      root = document.getElementById("root");
+    h1.innerHTML = "10 Books to read before you die";
+    root.appendChild(h1);
+    root.appendChild(ul);
+
+    for (const key in details) {
```
remarcmij: If you use the `bookTitles` array to determine which books to display and in which order, this `for` loop could look like:

```js
for (const title of bookTitles) {
  const book = books[title];
  // ... etc
}
```

Week1/homework/app.js
```diff
@@ -2,10 +2,130 @@
...
+      const li = document.createElement("li");
+      ul.appendChild(li);
+      li.setAttribute('id', key);
+      const obj = details[key];
```
remarcmij: `obj`, `details`, `key`:  they are all generic names that do not say much about what they contain.

Week1/homework/app.js
```diff
@@ -2,10 +2,130 @@
...
+    'lolita': 'images/lolita.jpg',
+    'maus': 'images/maus.jpg',
+    'poems': 'images/poems.jpg',
+    'things_fall_apart': 'images/Things Fall Apart.jpg'
+  };
+  const arr = [];
+  for (const j in booksCovers) {
+    arr.push(j);
+  }
+  console.log(arr);
```
remarcmij: Debug code to be removed from a PR.

Week1/homework/app.js
```diff
@@ -2,10 +2,130 @@
...
+    'lolita',
+    'maus',
+    'poems',
+    'things_fall_apart'
   ];
+  console.log(bookTitles);
 
+  //2.3 
+
+  /*  // hold this function  based on request 2.5
```
remarcmij: No commented out code in a PR.

Week1/homework/app.js
```diff
@@ -2,10 +2,130 @@
...
+    root.appendChild(ul);
+
+    for (let i = 0; i < titles.length; i++) {
+      const li = document.createElement("li");
+      ul.appendChild(li);
+      li.setAttribute('id', titles[i]);
+    }
+  }
+
+  createAndAppend(bookTitles);
```
remarcmij: Put this call in a function `main` that you execute through `window.onload`. This ensures that your code runs when the web page has completed loading.

Week1/homework/app.js
```diff
@@ -2,10 +2,130 @@
...
+      for (const key2 in obj) {
+        const childLi = document.createElement("li");
+        childLi.innerText = key2 + " : " + obj[key2];
+        childUl.appendChild(childLi);
+        key2 + ' : ' + obj[key2];
+      }
+
+    }
+  }
+  createAndAppend(booksDetails);
```
remarcmij: This should done inside the function `main()`.

Week1/homework/app.js
```diff
@@ -2,10 +2,130 @@
...
+      const getId = document.getElementById(j);
+      const img = document.createElement("img");
+      img.setAttribute("src", booksCovers[j]);
+      img.setAttribute("width", "300");
+      img.setAttribute("height", "300");
+      img.setAttribute("alt", "The Pulpit Rock");
+      getId.appendChild(img);
+    }
+  }
+  createImg(booksCovers);
```
remarcmij: Move inside the function `main()`.

PR #118, opened 2018-09-28
Hiba HomeWork Week1

heboshka:


Week1/homework/app.js
```diff
@@ -1,11 +1,140 @@
...
+  //2.5
+  function bookDetailsFunction() {
+    const main_list = document.getElementById('main_list');
+    const h1 = document.createElement('h1');
+    h1.innerText = 'My Books:';
+    main_list.appendChild(h1);
+    const ul = document.createElement('ul');
+    main_list.appendChild(ul);
+    const bookDetailsKeys = Object.keys(bookDetails);
+    for (let bookDetailsIndex = 0; bookDetailsIndex < bookDetailsKeys.length; bookDetailsIndex++) {
```
remarcmij: Instead of getting the keys from the object, it is better to use the titles from the `bookTitles` array, as that array was used to create the initial `li` elements.

So, the `bookDetails` object defines which books are available and the `bookTitles` array defines which books to display and in which order to display them.

Week1/homework/app.js
```diff
@@ -1,11 +1,140 @@
...
+      const languageP = document.createElement("p");
+      languageP.innerText = 'Language: ' + bookDetails[currentKeys].language;
+      li.appendChild(languageP);
+      const authorP = document.createElement('p');
+      authorP.innerText = 'Author: ' + bookDetails[currentKeys].author;
+      li.appendChild(authorP);
+
+    }
+  }
+  bookDetailsFunction();
```
remarcmij: This function should be called from within the `main()` function (or in your case, your `bodyFunction`). Try and move the `<script src='app.js'> </script>` tag in your index.html file to inside the `<head>` tag and see what happens.

Week1/homework/app.js
```diff
@@ -1,11 +1,140 @@
...
+
+  function bookWithImg() {
+    const main_list = document.getElementById('main_list');
+    const h1 = document.createElement('h1');
+    h1.innerText = 'My Books:';
+    main_list.appendChild(h1);
+    const ul = document.createElement('ul');
+    main_list.appendChild(ul);
+    const bookDetailsKeys = Object.keys(bookDetails);
+    for (let bookDetailsIndex = 0; bookDetailsIndex < bookDetailsKeys.length; bookDetailsIndex++) {
```
remarcmij: The same comments I made for the previous `for` loop applies here to: add the images to the existing `ul` and its corresponding `li` element.

Week1/homework/app.js
```diff
@@ -1,11 +1,140 @@
...
+        ul.appendChild(images);
+      }
+
+
+    }
+
+  }
+
+
+  bookWithImg();
```
remarcmij: This function should be called from within the `main()` function.

Week1/homework/app.js
```diff
@@ -1,11 +1,140 @@
...
+      const booksId = Object.keys(booksImg);
+      if (bookDetailsKeys[bookDetailsIndex] === booksId[bookDetailsIndex]) {
+        let mySrc = Object.values(booksImg);
+        let images = document.createElement('img');
+        images.setAttribute("src", mySrc[bookDetailsIndex]);
+        ul.appendChild(images);
+      }
+
+
+    }
```
remarcmij: The preceding blank lines serve no purpose. It is best to get rid of them.

Week1/homework/app.js
```diff
@@ -1,11 +1,140 @@
 'use strict';
 {
+  //2.1
   const bookTitles = [
-    // Replace with your own book titles
-    'harry_potter_chamber_secrets'
+    'brain_camp ',
```
remarcmij: There is a trailing space that should not be there.

Week1/homework/app.js
```diff
@@ -1,11 +1,140 @@
...
+    'sophie_washington_secret_santa',
+    'el_deafo',
+    'american_born_chinese',
+    'queen_of_the_world!',
```
remarcmij: Do not use special characters because then you can't use these names as element IDs. Make sure these names conform to the rules for JavaScript variable names: letters, digits and the underscore.

remarcmij:
There are still some issues unaddressed in your latest update.

PR #122, opened 2018-09-30
js2 wk1 Karam not finished

KaramGhazzi:


Week1/homework/app.js
```diff
@@ -1,11 +1,109 @@
...
+  parent.appendChild(elm);
+  return elm;
+}
+function booksInformation() {
+  const list = document.createElement("section");
+  list.setAttribute("id", "bookList");
+  const ul = createAndAppend("ul", list);
+  document.body.appendChild(list);
+  for (let i = 0; i < Object.keys(bookTitles).length; i++) {
+    const myObject = Object.keys(bookTitles);
```
Jawhar-B: `myObject` is a meaningful name for the writer while the reader is expecting a name that make sense for what `myObject` contains exactly.

Week1/homework/app.js
```diff
@@ -1,11 +1,109 @@
...
+    const bookName = myObject[i];
+    const bookDetails = bookTitles[bookName];
+
+    const li = createAndAppend("li", ul);
+    const h1 = createAndAppend("h1", li);
+    h1.setAttribute("class", "heading");
+
+    const img = createAndAppend("img", li);
+    img.setAttribute("class", "allImg");
+    img.setAttribute("src", "booksimg/" + [i] + ".jpg");
```
Jawhar-B: it's a very nice way of using your tools as they are 🙂.
But it would be nice to have another look to this line while adding the strings to each others.

to be more precise **_(look to the types)_**

Jawhar-B:
Oh I putted a `needs work` label and suggested changes instead of approving.

I am sorry my bad It was working fine so I don't think it needs work
There are only some minor things, nothing to worry about.

PR #123, opened 2018-10-01
Week1

karamad10:


Week1/homework/app.js
```diff
@@ -1,11 +1,84 @@
...
+      'lincoln_in_the_bardo': "https://images.penguinrandomhouse.com/cover/9780553397574",
+
+    };
+    const keys = Object.keys(images);
+    const values = Object.values(images);
+
+    for (let i = 0; i < myBooks.length; i++) {
+      const li = document.getElementById((myBooks[i]));
+      const imgDiv = document.createElement('div');
+      const books = 'book' + [i];
```
remarcmij: This expression `[i]` is an array with one element: the value of `i`, e.g. `[0]` for the first loop iteration. This is probably not what you intended.

Week1/homework/app.js
```diff
@@ -1,11 +1,84 @@
...
+      imgDiv.className = books;
```
remarcmij: This class name does not seem to be used. Why would you give each `div` here a unique class name anyway?

Week1/homework/app.js
```diff
@@ -1,11 +1,84 @@
-'use strict';
-{
-  const bookTitles = [
-    // Replace with your own book titles
-    'harry_potter_chamber_secrets'
+'use strict'; {
```
remarcmij: I would place the opening curly brace on a new line. It has no relation `'use strict';`.

Week1/homework/app.js
```diff
@@ -1,11 +1,84 @@
...
+    'the_god_delusion': { title: 'The God Delusion', author: 'Richard Dawkins', language: 'English' },
+    'the_old_man_and_the_sea': { title: 'The Old Man And The Sea', author: 'Ernest Hemingway', language: 'English' },
+    'righteous': { title: 'Righteous', author: 'Joe Ide', language: 'English' },
+    'the_fourth_monkey': { title: 'The Fourth Monkey', author: 'Jeffery Deaver', language: 'English' },
+    'norse_mythology': { title: 'Norse Mythology', author: 'Neil Gaiman', language: 'English' },
+    'lincoln_in_the_bardo': { title: 'Lincoln In The Bardo', author: 'George Saunders', language: 'English' }
+  };
+
+
+  function myBooksDetails(ul) {
```
remarcmij: For function names, alway try to give a name that indicates some sort of action, because functions 'do' something. So the name should have a verb as its head word. For instance, this function could be called `renderBookDetails()`.

Week1/homework/app.js
karamad10: I did intend that so the heading would have a class of 'title0' and the corresponding image would have a class of 'book0'.

Week1/homework/app.js
karamad10: when I assigned it I had more styling ideas in mind but due to lack of time, I didn't implement much styling.

Week1/homework/app.js
karamad10: thank you, I actually thought they are related I will fix it.

Week1/homework/app.js
karamad10: I will keep that in mind next time.

Week1/homework/app.js
remarcmij: In that case you should use `book + i` and leave out the brackets.

PR #125, opened 2018-10-02
HW JS2 Week2 from Solomon

ZerSolo:


Week2/homework/maartjes_work.js
```diff
@@ -44,4 +42,22 @@ const tuesday = [
 
 const tasks = monday.concat(tuesday);
 
-// Add your code here
+const durationsInHours = tasks.map(task => Number(task.duration / 60).toFixed(2));
+
+const twoHrsLessDurations = durationsInHours.filter(twohrs => twohrs >= 2);
+
+const sallaryPerTask = durationsInHours.map(sallaryPerduration => sallaryPerduration * 20);
```
remarcmij: Instead of `durationInHours` you should use the filtered durations from the previous statement.

The number 20 comes out of the blue without an explanation. Developers often call this a 'magic' number. I assume it is the hourly rate that Maartje earns. It is best to give it name so we can recognise it for what it is:

```js
const hourlyRate = 20;
```

Now use this variable instead of the number 20 in the mapping function.

Since what you are mapping are durations, is it again best to name the function argument `duration` rather than `sallaryPerduration`.

Note also the spelling error: `sallary` -> `salary`. The VSCode spell checker should have flagged this.

Week2/homework/maartjes_work.js
```diff
@@ -44,4 +42,22 @@ const tuesday = [
...
+console.log(sallaryPerTask);
```
remarcmij: We don't need this debug console.log in a PR.

Week2/homework/maartjes_work.js
```diff
@@ -44,4 +42,22 @@ const tuesday = [
...
+
+const totalSallery = sallaryPerTask.reduce((accum, value) => accum + value, 0);
+
+console.log(totalSallery);
+
+
+function formatEuro(money) {
+  return '€ ' + Number(money).toFixed(2)
+}
```
remarcmij: This function is a bit overkill. It does next to nothing and it is used once only. You do not need to convert `money` to a number. It should be a number already.

Week2/homework/map_filter.js
```diff
@@ -1,5 +1,9 @@
 'use strict';
 
+//2.1
+
 const numbers = [1, 2, 3, 4];
 
-// Add your code here
+const doubledOnlyOddnumbers = numbers.filter(number => number % 2 !== 0).map(number => number * 2)
```
remarcmij: Excellent! As a minor point, many developers prefer to format a statement like this as:

```js
const doubledOnlyOddnumbers = numbers
  .filter(number => number % 2 !== 0)
  .map(number => number * 2);
```

ZerSolo:
Dear Gim,
well received with thanks. According to your feedback, I will correct my
code as soon as possible

best regards,
Zerisenay Solomon

PR #126, opened 2018-10-03
Week1

joudi6:
javaScript2 Week1


Week1/homework/app.js
```diff
@@ -2,10 +2,104 @@
 {
   const bookTitles = [
     // Replace with your own book titles
-    'harry_potter_chamber_secrets'
+    'dont_tell_me_your_afraid', 'dont_tell_mummy', 'positive_thinking', 'time_management', 'The alchemist', 'the_power_of_now', 'ekadoli', 'Before_we_were_yours', 'hebrew_female_in_my_heart', 'amareeta'
```
remarcmij: This is a rather long line. It is better to each string on a new line, e.g.:

```js
const bookTitles = [
  'dont_tell_me_your_afraid',
  'dont_tell_mummy',
  ...
];
```

Week1/homework/app.js
```diff
@@ -2,10 +2,104 @@
...
   // Replace with your own code
-  console.log(bookTitles);
+  //  2.3
+  function makeBooksTitlesList() {
+    const ul = document.createElement('ul');
+    ul.innerHTML = 'Books Titles: ';
+    document.getElementsByTagName('body')[0].appendChild(ul);
+    for (let i = 0; i < bookTitles.length; ++i) {
+      const li = document.createElement('li');
+      li.innerHTML = bookTitles[i];
```
remarcmij: The content of the `bookTitles` array that you created is not intended for dispIay. The purose of these titles so that  _you can use these later as HTML id attributes_. This is also mentioned as a hint at the end of step 2.7.

So in this loop where your create and and append `li` elements you should not set the `innerHTML` but do this:

```js
li.setAttribute('id', bookTitles[i]);
```

Week1/homework/app.js
```diff
@@ -2,10 +2,104 @@
...
+      ul.appendChild(li);
+    }
+
+  }
+
+  makeBooksTitlesList();
```
remarcmij: It is is best to place this function call inside of a function `main` along with the function call in 103. Then you could assign function `main` to `window.onload` like this:

```js
window.onload = main;
```

This ensure that you code is executed when the web page is fully loaded.

Week1/homework/app.js
```diff
@@ -2,10 +2,104 @@
...
+    }
+
+  };
+  //2.5
+
+  function makeBooksDetailsPage() {
+    const container = document.createElement('div');
+    document.getElementsByTagName('body')[0].appendChild(container);
+    container.innerText = 'Books Details:';
+    for (const obj in booksDetails) {
```
remarcmij: These book details should be added to the `li`-s that you created in the first loop. If used `setAttribute()` to set the `id` of each `li` to a book title your `for` loop could iterate through these book titles, get a reference to the corresponding `li` by using `getElementById()` and then append the book details to that `li`.

Week1/homework/app.js
```diff
@@ -2,10 +2,104 @@
...
+    'dont_tell_me_your_afraid', 'dont_tell_mummy', 'positive_thinking', 'time_management', 'The alchemist', 'the_power_of_now', 'ekadoli', 'Before_we_were_yours', 'hebrew_female_in_my_heart', 'amareeta'
   ];
 
-
   // Replace with your own code
-  console.log(bookTitles);
+  //  2.3
+  function makeBooksTitlesList() {
+    const ul = document.createElement('ul');
+    ul.innerHTML = 'Books Titles: ';
```
remarcmij: A `ul` can only have `li`-s as children. It should not have text content. That would make for invalid HTML.

Week1/homework/app.js
```diff
@@ -2,10 +2,104 @@
...
+      img.setAttribute('src', './' + booksDetails[obj].title + '.jpg');
+      img.setAttribute('alt', 'img ' + booksDetails[obj].title);
+      paragraph1.innerHTML = 'Wrote in: ' + booksDetails[obj].language;
+      paragraph2.innerHTML = 'By: ' + booksDetails[obj].author;
+
+      document.getElementsByTagName('div')[0].appendChild(h1);
+      document.getElementsByTagName('div')[0].appendChild(img);
+      document.getElementsByTagName('div')[0].appendChild(paragraph1);
+      document.getElementsByTagName('div')[0].appendChild(paragraph2);
+    }
```
remarcmij: As I mentioned, you should be appending all this stuff the `li` that corresponds to the current book rather than to some separate `div`.

PR #127, opened 2018-10-03
js hw week2

karamad10:


Week2/homework/maartjes_work.js
```diff
@@ -43,5 +43,70 @@ const tuesday = [
...
+let hours = 0;
+for (let i = 0; i < tasks.length; i++) {
+  const num = tasks[i].duration;
+  hours += num;
+}
+hours /= 60;
+console.log(hours);
+*/
+
+let hours = 0;
```
remarcmij: We don't need this `hours` variable and the computations that are related to it.

Week2/homework/maartjes_work.js
```diff
@@ -43,5 +43,70 @@ const tuesday = [
...
+const duration = (tasks.map(tasks => tasks.duration));
```
remarcmij: You need to be more precise in your naming. When we map a `tasks` array the mapping function receives one `task` at a time. Therefore the argument name should be `task` and not `tasks`. The result of `.map()` is an array. Therefore the result should be assigned to a variable `durations` and not `duration`.

In the assignment is was asked to convert the durations from minutes to hours: _Map the tasks to durations in hours._ So you need to divide by 60.

Finally, there is not need for the extra set of parentheses around enclosing the whole expression:

```js
const durations = tasks.map(task => task.duration / 60);
```

Week2/homework/maartjes_work.js
```diff
@@ -43,5 +43,70 @@ const tuesday = [
...
+for (let i = 0; i < duration.length; i++) {
+  hours += duration[i];
+}
+hours /= 60;
+console.log(hours);
```
remarcmij: Your PR should not include console.log statements that you used for debugging.

Week2/homework/maartjes_work.js
```diff
@@ -43,5 +43,70 @@ const tuesday = [
...
+for (let i = 0; i < duration.length; i++) {
+  if (duration[i] >= 120) twoOrMoreHours.push(duration[i]);
+}
+console.log(twoOrMoreHours);
+*/
+
+
+const twoOrMoreHours = (duration.filter(values => {
+  return values >= 120;
+}
```
remarcmij: Again, make sure that you get your use of singular and plural in order:

```js
const twoOrMoreHours = durations.filter(duration => duration >= 120);
```

If you had converted the durations to hours as was requested, the condition would of course have been `duration >= 2`.

Week2/homework/maartjes_work.js
```diff
@@ -43,5 +43,70 @@ const tuesday = [
...
+let salary = 0;
+
+for (let i = 0; i < duration.length; i++) {
+  salary = hours * salaryPerHour;
+}
+console.log('€ ' + salary.toFixed(2));
+
+*/
+
+const HoursArray = [hours];
```
remarcmij: Instead of this array with a single element, you should use the `twoOrMoreHours` array of durations that you created earlier.

Week2/homework/maartjes_work.js
```diff
@@ -43,5 +43,70 @@ const tuesday = [
...
+function add(sum, num) {
+  return sum + num;
+}
+const salary = HoursArray.map(HoursArray => HoursArray * 24.35).reduce(add, 0);
```
remarcmij: Variable names should be in camelCase. You should apply the `.map()` method to the `twoOrMoreHours` array.

The number 24.35 comes out of the blue without explanation. Developers call this a `magic` number.  I assume it represents the hourly rate that Maartje earns. Therefore best to name it as such:

```js
const hourlyRate = 24.35;
```

```js
const salary = twoOrMoreHours
  .map(duration => duration * hourlyRate)
  .reduce(add, 0);
```

Week2/homework/map_filter.js
```diff
@@ -1,5 +1,24 @@
...
+*/
+function isOdd(num) {
+  return num % 2 !== 0;
+}
+const filteredNumber = (numbers.filter(isOdd));
+function double(num) {
+  return num * 2;
+}
+
+console.log('The doubled numbers are', filteredNumber.map(double));
```
remarcmij: This is a lot of code for something really simple. I would just use anonymous arrow functions and method chaining, e.g.:

```js
const newNumbers = numbers
  .filter(number => number % 2 !== 0)
  .map(num => num * 2);
```

karamad10:
I've made plenty of changes based on your comments.

PR #130, opened 2018-10-04
week2

salar01:


Week2/homework/maartjes_work.js
```diff
@@ -1,47 +1,84 @@
...
-    name: 'Look at application assignments new students',
+    name: "Look at application assignments new students",
     duration: 40
   }
 ];
 
 const tasks = monday.concat(tuesday);
 
-// Add your code here
+let durationOfTasks = tasks.map(Element => Element.duration);
```
remarcmij: Since you are mapping the elements of a `tasks` array each element of that array is a `task`. So it would be more explicit to call the function argument `task` instead of the more generic `Element`. Moreover, variable, function and parameters names should be in camelCase (i.e. start with a lowercase letter).

Since you are extracting the `duration` values from the `tasks` it is a bit redundant to call the result of this mapping `durationOfTasks`. A simple `durations` will do fine. This name also implies that it is an array (because of the plural 's').

VSCode (ESLint) give a warning that the `durationOfTasks` variable is not re-assigned and therefore a `const` would be preferred over `let`. Do you get such warnings too? If not, check out the [VSCode Tips](https://github.com/HackYourFuture/fundamentals/tree/master/VSCodeTips#installation-instructions) to install ESLint. If you do get such warnings, please act on them. They help you to improve you code.

Week2/homework/maartjes_work.js
```diff
@@ -1,47 +1,84 @@
...
+
+ console.log(durationOfTasks);
```
remarcmij: You should not leave debug console.log statements in a PR.

Week2/homework/maartjes_work.js
```diff
@@ -1,47 +1,84 @@
...
+
+// Map the tasks to durations in hours.
+let durationInHours = durationOfTasks
+  .map(element => element / 60)
```
remarcmij: Since you are not chaining array methods here, it is best to place .map()` on the same line as you did in line 48. Since you are mapping durations here, a single elements of that array can be called a `duration`. Therefore that is a more explicit (and therefore) better name for the arrow function argument than the more generic `element`.

ESLint: The `let` can be a `const`.

As the result is the result of the mapping is an array, it is best to reflect that in the name by using a plural for durations: `durationsInHours`.

Week2/homework/maartjes_work.js
```diff
@@ -1,47 +1,84 @@
...
+
+ console.log(durationInHours);
```
remarcmij: No debug type console.logs in a PR.

Week2/homework/maartjes_work.js
```diff
@@ -1,47 +1,84 @@
...
+
+// Filter out everything that took less than two hours
+  
+let lessThanTwoHours = durationInHours
+    
+.filter(element => element >= 2)
+console.log(lessThanTwoHours);
```
remarcmij: If you filter out all tasks that take less than two hours then you filtered array contains all elements that take two hours or more. Therefore the variable name `lessThanTwoHours` is incorrect and misleading. A name such as `durationsOTwoHoursOrMore` or something similar would be correct.

Lines 60 and 62 are part of the same statement. It is very confusing to have a blank line in between. I would append `.filter()` directly at the end of line 60.

The console.log should be present in a PR.

ESLint: the `let` should be a `const`.

Week2/homework/maartjes_work.js
```diff
@@ -1,47 +1,84 @@
...
+
+// Multiply the each duration by a per-hour rate for billing and sum it all up.
+
+let paymentPerHour = durationOfTasks
+  .map(element => (element / 60) * 100)
```
remarcmij: The assignment states: _Each step should build on the result of the previous step._ This implies that you should use the **filtered** durations for the payment calculation.

I assume the number 100 is the hourly rate for Maartje (if that is the case, I want her job... oh wait, it is already taken).  Developers call such numbers that appear out of the blue without an obvious explanation 'magic' numbers. You should not use magic numbers because they make it hard to understand what is happening. Instead, define a named constant and use that throughout your code:

```js
const hourlyRate = 100;
```

Week2/homework/maartjes_work.js
```diff
@@ -1,47 +1,84 @@
...
+
+console.log(paymentPerHour);
```
remarcmij: No debug type console.log statements in a PR.

Week2/homework/maartjes_work.js
```diff
@@ -1,47 +1,84 @@
...
+
+let totalPayment = paymentPerHour
+  .reduce(function (a, b) {
+    return a + b;
+    
+  });
+console.log(`Total hours payment is: ${totalPayment}`);
```
remarcmij: You are already printing out the result in line 82. This console.log should not be present in a PR.

Week2/homework/maartjes_work.js
```diff
@@ -1,47 +1,84 @@
...
+
+// Output a formatted Euro amount, rounded to Euro cents, e.g: € 12.34.
+
+let amountEUR = totalPayment.toFixed(2);
+console.log(`The total amount to Euro is: ${amountEUR} EUR.`);
+
```
remarcmij: Because you are not using the **filtered** durations the calculated total amount is not the expected value according to the assignment.

Week2/homework/maartjes_work.js
```diff
@@ -1,47 +1,84 @@
...
     duration: 10
   },
   {
-    name: 'Talk to a lot of people',
+    name: "Talk to a lot of people",
     duration: 200
   },
   {
-    name: 'Look at application assignments new students',
+    name: "Look at application assignments new students",
```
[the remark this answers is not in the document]
salar01: Hello Jim, Thanks for your comments, i have disabled it :)

Week2/homework/maartjes_work.js
salar01: Thanks for your preciouscommnets, i changed all things that you highlighted. I have already installed ESLint, but i did not get such a warning. I'll aperciate you if you check my VSCode tomorrow. I will come to you. 👍

PR #132, opened 2018-10-05
week2 homework

joudi6:


Week2/homework/maartjes_work.js
```diff
@@ -45,3 +45,11 @@ const tuesday = [
 const tasks = monday.concat(tuesday);
 
 // Add your code here
+
+const maartjeHours = tasks.map(duration => duration.duration / 60).filter(durations => durations >= 2);
+
+console.log('maartje got ' + maartjeHours.reduce((accumulator, currentValue) => accumulator + (currentValue * 17),
```
remarcmij: The number 17 is I assume Maartje's hourly rate. Developers call such a number that comes out of the blue without some obvious explanation a 'magic' number. Best to avoid it as it makes your code harder to understand. Instead, define a constant that gives a number to this number to take away the magic:

```js
const hourlyRate = 17;
```

PR #134, opened 2018-10-06
my homework week1 JS

RimSebai:
sorry for late


Week1/homework/app.js
```diff
@@ -1,11 +1,140 @@
...
+  //   }
+    
+  // }
+
+  function createBookList(books){
+
+    let list = document.createElement("UL");
+    document.body.appendChild(list) ; 
+    list.setAttribute("id","book_list");
+    for(let i in books){
```
remarcmij: Rather than looping through the property names of your `bookInfo` object it is better to loop through the `bookTitles` array and use those titles to get the book details from your `bookInfo` object. In that way the `bookTitles` array controls which books from your `bookInfo` object should be displayed and also in which order. And the `bookInfo` object is just a place where you keep all these details.

Week1/homework/app.js
```diff
@@ -1,11 +1,140 @@
...
+      let li = document.createElement("LI");
```
remarcmij: Most developers would use lowercase letters for the element names you want to create, although obviously uppercase works too.

Week1/homework/app.js
```diff
@@ -1,11 +1,140 @@
...
+    }
+
+   
+    } 
+   
+  
+   
+
+covering(bookCovers);
+console.log(bookCovers);
```
remarcmij: This looks like a console.log for debugging. It should not be in a PR.

Week1/homework/app.js
RimSebai: I’ve read your feedback, and I made changes you asked me to do ...thank you!

RimSebai:
I’ve read your feedback, and I made changes you asked me to do ...thank you!

remarcmij:
Thanks, @RimSebai, but I think you left some feedback uncorrected.

PR #135, opened 2018-10-06
Week2

RimSebai:
Sorry for late

### DIFF
--- a/Week2/homework/maartjes_work.js
+++ b/Week2/homework/maartjes_work.js
@@ -43,63 +43,6 @@ const tuesday = [
 ];
 
 const tasks = monday.concat(tuesday);
-
-//console.clear()
-const numbers = [1, 2, 3, 4];
-let newNumbers = [];
-
-// for (let i = 0; i < numbers.length; i++) {
-//   if (numbers[i] % 2 !== 0) {
-//     newNumbers.push(numbers[i] * 2);
-//   }
-// }
-newNumbers = numbers.filter(el => el % 2 !== 0).map(el => el * 2)
-
-console.log('The doubled numbers are', newNumbers);
-
-const monday = [
-  {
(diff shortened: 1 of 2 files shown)


Week2/homework/maartjes_work.js
```diff
@@ -43,5 +43,13 @@ const tuesday = [
 ];
 
 const tasks = monday.concat(tuesday);
+const Maartje = tasks.map(obj => obj.duration / 60).filter(el => el >= 2)
+  .reduce(
+    (sum, hours) => hours + sum
+  ) * 30.34
+
```
remarcmij: This method chain would look a bit better if you would put each method on a new line:

```
const Maartje = tasks
  .map(obj => obj.duration / 60)
  .filter(el => el >= 2)
  .reduce((sum, hours) => hours + sum) * 30.34;
```

The result of this calculation is not `Maartje` but perhaps `totalEarnings`. 

Since you are mapping `tasks`, the callback function that you pass to `.map()` receives one `task` at a time. So why not call that argument `task` rather than the rather nondescript name `obj`? Similarly, the `.filter()` method filters durations, so why not use `duration` rather `el` for the name?

The number 30.34 is Maartje's hourly rate I assume. Developers call a number like this that comes out of the blue with some inexplicable value a 'magic' number. Why not take the magic away and give this number a name, e.g.:

```
const hourlyRate = 30.34;
```

Now, use this name instead of the number in your calculation.

Week2/homework/maartjes_work.js
```diff
@@ -43,5 +43,13 @@ const tuesday = [
...
+console.log(Maartje.toFixed(2) + ' €')
```
remarcmij: It is customary to put the Euro sign in front of the number.

PR #136, opened 2018-10-06
js2 week2

KaramGhazzi:
homework Javascript 2 week2

### DIFF
--- a/Week2/homework/maartjes_work.js
+++ b/Week2/homework/maartjes_work.js
@@ -44,4 +44,23 @@ const tuesday = [
 
 const tasks = monday.concat(tuesday);
 
+
 // Add your code here
+//shortened code
+let paidPerHour = tasks.map(time => time.duration / 60).filter(filteredTime => filteredTime > 2).reduce((accumulator, elm) => accumulator + (elm * 16.50));
+let totalSalary = "€" + paidPerHour.toFixed(2);
+console.log(totalSalary);
+
+//readable code
+//mapped the time per hour 60min
+let salaryPerHour = tasks.map(time => time.duration / 60);
+
+//filtered all durations from 2 hours and above
+let filtered = salaryPerHour.filter(filteredTime => filteredTime > 2);
+
+//reduced the working hours to 16.50 euro per hour
+let reduced = filtered.reduce((accumulator, elm) => accumulator + (elm * 16.50));
+//result
+let totalResult = "€" + reduced.toFixed(2);
+console.log(totalResult);
+
--- a/Week2/homework/map_filter.js
+++ b/Week2/homework/map_filter.js
@@ -3,3 +3,6 @@
 const numbers = [1, 2, 3, 4];
 
 // Add your code here
+
+let resultNumber = numbers.filter(number => number % 2 !== 0).map(number => number * 2);
+console.log(resultNumber);


Week2/homework/maartjes_work.js
```diff
@@ -44,4 +44,23 @@ const tuesday = [
 
 const tasks = monday.concat(tuesday);
 
+
 // Add your code here
+//shortened code
+let paidPerHour = tasks.map(time => time.duration / 60).filter(filteredTime => filteredTime > 2).reduce((accumulator, elm) => accumulator + (elm * 16.50));
```
remarcmij: This long statement is best broken down in multiple lines, for instance:

```js
let paidPerHour = tasks
  .map(time => time.duration / 60)
  .filter(filteredTime => filteredTime >= 2)
  .reduce((accumulator, elm) => accumulator + (elm * 16.50));
```

You use of the `reduce` method has a subtle error in it that cause the calculation to be wrong. See below in your second version for an explanation. Here I want to focus on the names you choose for the variables. Choosing appropriate names can help to increase the readability of the code.

1. When you use `.map()` on an array the function that you pass as an argument to the `.map()` method receives one element of that array at a time. So when you map the `tasks` array a single element of that array is a `task` and not a `time`. You want to map it to a time (or duration), but the argument is a `tasks`.

The same applies to the `.filter()` method you used. What you filter are durations and a single element is called a `duration`. That should be the name of the parameter for the `.filter()` callback and not `filteredTime`. 

The `reduce()` method here works on the filtered durations, so rather than using `elm` (which I assume means element - avoid abbreviations) I would call this argument `duration`.

The number `16.50` is the hourly rate for Maartje, I assume. Developers call such a number that comes out of the blue without some obvious explanation a 'magic' number. It is best avoid magic numbers and define a names constant that takes the magic away and use that instead of the number, e.g.:

```js
const hourlyRate = 16.50;
```

The final result of this calculation is not `paidPerHour`, it is the total amount of money that Maartje earned. So `totalEarnings` or something similar would be more appropriate for this result.

ESLint warns that the variable `paidPerHour` and `totalSalary` are not reassigned and therefore should preferably be a `const. Did you get such warnings too? If you do, you should follow up on them. If not, check you VSCode setup, see the [VSCode Tips](https://github.com/HackYourFuture/fundamentals/tree/master/VSCodeTips#installation-instructions).

Week2/homework/maartjes_work.js
```diff
@@ -44,4 +44,23 @@ const tuesday = [
...
+
+//readable code
+//mapped the time per hour 60min
+let salaryPerHour = tasks.map(time => time.duration / 60);
+
+//filtered all durations from 2 hours and above
+let filtered = salaryPerHour.filter(filteredTime => filteredTime > 2);
+
+//reduced the working hours to 16.50 euro per hour
+let reduced = filtered.reduce((accumulator, elm) => accumulator + (elm * 16.50));
```
remarcmij: See my comments on naming above.

Now, about the error with the `.reduce` method. The `.reduce()` method takes an optional second parameter: the initial value for the accumulator. If you leave it out it will take the first element of the array to initialise the accumulator and will iterate through the rest of the array. In you case you are multiplying each array element by the hourly rate. However, the first element that was used by `.reduce()` to initialise the accumulator was **not** multiplied by the hourly rate. The end result is that you are underpaying Maartje.

A simple way to correct this is to explicitly call reduce with a second parameter, the constant `0`, to initialise the accumulator. Now, `.reduce()` doesn't take the first array element to initialise the accumulator and iterates through all array elements, including the first one.

Week2/homework/map_filter.js
```diff
@@ -3,3 +3,6 @@
 const numbers = [1, 2, 3, 4];
 
 // Add your code here
+
+let resultNumber = numbers.filter(number => number % 2 !== 0).map(number => number * 2);
```
remarcmij: Correct. But I would break up the method chain into multiple lines.

Week2/homework/maartjes_work.js
KaramGhazzi: thank you so much Mr. Jim i am working on it right now its really major issues!!

KaramGhazzi: and yes i am got that warning from VScode but when i am happy that i've done the homework i just dont see in my eyes anymore (: and directly push the homework.

PR #137, opened 2018-10-07
java script2 Week2 hw

Anashost:


Week2/homework/maartjes_work.js
```diff
@@ -45,3 +45,14 @@ const tuesday = [
 const tasks = monday.concat(tuesday);
 
 // Add your code here
+const inHours = tasks.map(hour => hour['duration']/60);
```
remarcmij: What you map here are tasks. The callback function that you pass to .`map()` receives one task at a time. Therefore it is logical to name the parameter of that function `task` and not `hour`.

You should use dot-notation to access properties from object if not know the name of the property beforehand, which is true in the case. The result of the `.map()` method in this case is an array of durations. Will all that, for clarity, you could rewrite this statement as:

```js
const durations = tasks.map(task => task.duration / 60);
```

Week2/homework/maartjes_work.js
```diff
@@ -45,3 +45,14 @@ const tuesday = [
...
+console.log(inHours);
```
remarcmij: Do not leave console.log that you used for debug purposes when you submit your final work in a PR.

Week2/homework/maartjes_work.js
```diff
@@ -45,3 +45,14 @@ const tuesday = [
...
+
+const moreThanTwo = inHours.filter(hour => hour > 2);
```
remarcmij: The assignment states: _Filter out everything that took less than two hours (i.e., remove from the collection)_, which implies that every task that took two hours or more should stay in. But in your case you leave out tasks that took exactly two hours.

The name `moreThanTwo` does not really reflect the fact that this will be an array. For instance, a name such as `filteredDurations` (note the plural 's' at the end) would reflect this better.

Week2/homework/maartjes_work.js
```diff
@@ -45,3 +45,14 @@ const tuesday = [
...
+console.log(moreThanTwo);
```
remarcmij: Take out the console.log

Week2/homework/maartjes_work.js
```diff
@@ -45,3 +45,14 @@ const tuesday = [
...
+
+const amountPerHour = moreThanTwo.map(hour => hour * 15);
+console.log(amountPerHour);
+
+const  roundedToEuro = amountPerHour.toLocaleString('it-IT', {style: 'currency', currency: 'EUR'});
```
remarcmij: For this assignment we are only interested in a single number that represents the total earnings for Maartje in Euros. Here you are apparently trying to report a payment for each individual task and leave it to us to add it all up. Besides the formatting in Euros doesn't seem to work. And why did you choose Italy ('it-IT') as the locale?

Anashost: Thanks for feedback , i have tried tofixed(2) but its did not works for me then i found this on google .toLocaleString('it-IT', {style: 'currency', currency: 'EUR'}); and it works just fine in the browser , but now i have fixed all what u have mentioned.

remarcmij: Have you also tried `toFixed(2)` (mind the exact spelling)? If you google and find something that seems to work but that you can't explain, then you shouldn't use it. It is not good enough if it works without you knowing why.

PR #140, opened 2018-10-10
Week3 

karamad10:


Week3/homework/2-step3.js
```diff
@@ -1,8 +1,27 @@
...
+{
+  function threeFive(startIndex, stopIndex, threeCallback, fiveCallback) {
+    const values = [];
+    for (let i = startIndex; i <= stopIndex; i++) {
+      values.push(i);
+    }
+
+    values.forEach(num => {
+      if (num % 3 === 0)
+        threeCallback(num);
```
remarcmij: Make it a habit to always enclose `if-else` and `for` loop blocks in curly braces (`{' and  `}`), even if it only contains one statement. This makes it easy to add more statements to that block later. If you leave out the curly braces it would be easy to make a mistake like this:

```js
if (num % 3 === 0)
  console.log('bla bla');
  threeCallback(num);
if (num % 5 === 0)
  fiveCallback(num);
```

The indentation might lead you to conclude that the first `if` statement block contains two statements. But only the  `console.log` is controlled by the `if` statement.

Week3/homework/2-step3.js
```diff
@@ -1,8 +1,27 @@
...
-  // Add your code here
-}
+{
+  function threeFive(startIndex, stopIndex, threeCallback, fiveCallback) {
+    const values = [];
+    for (let i = startIndex; i <= stopIndex; i++) {
+      values.push(i);
+    }
+
+    values.forEach(num => {
```
remarcmij: A single element of a `values` array is `value`. I would prefer that name over the name `num`.

Week3/homework/3-step3.js
```diff
@@ -1,25 +1,42 @@
...
+  // use a 'do...while' loop
+  function repeatStringNumTimesWithDoWhile(str, num) {
+    let newStr = '';
+    do {
+      newStr += str;
+      num--;
+    }
+    while (num > 0);
+    return newStr;
+  }
```
remarcmij: This `do-while` version fails one of the tests of the freeCodeCamp exercise: `repeatStringNumTimes("abc", -2)` should return `""`.

Week3/homework/4-step3.js
```diff
@@ -1,2 +1,19 @@
...
-// paste your freeCodeCamp solutions in here
+//----------1
+{
+  function Dog() {
+    this.name = 'Jäger';
+    this.color = 'black and white';
+    this.numberOfLegs = 4;
+  }
+
+  const Hound = new Dog();
```
remarcmij: Variable, function and parameter names should be in camelCase. Only constructor functions (function that you call in combination with the keyword `new`) and ES6 class names (not yet covered in the lectures) should be in PascalCase.

So this would be correct:

```js
const hound = new Dog();
```

`Dog()` is a constructor function. `hound` is an ordinary variable name.

Week3/homework/4-step3.js
```diff
@@ -1,2 +1,19 @@
...
+
+  console.log(Hound);
+
+  const brownDog = new Dog();
+  brownDog.name = 'Jackal';
+  brownDog.color = 'brown';
+
+  console.log(brownDog);
+}
```
remarcmij: I think you grasped the concepts of object creation with the `new` keyword well.

Week3/homework/5-step3.js
```diff
@@ -1,2 +1,17 @@
...
+      }
+    }
+    return product;
+  }
+
+  multiplyAll([[1, 2], [3, 4], [5, 6, 7]]);
+
+  console.log(multiplyAll([[1, 2], [3, 4], [5, 6, 7]]));
+  console.log(multiplyAll([[1, 2], [3, 4]]));
+}
```
remarcmij: Excellent that you replaced the `var` keywords with `let`.

Week3/homework/6-step3.js
```diff
@@ -4,3 +4,25 @@ const arr2d = [[1, 2], [3, 4], [5, 6]];
...
+
+    for (let k = 0; k < arr3d[i].length; k++) {
+
+      for (let j = 0; j < arr3d[i][k].length; j++) {
+
+        console.log('[' + arr3d[i][k][j] + ']');
+      }
+    }
+  }
+}
```
remarcmij: Okay, but what made you choose the order of `i, k, j` instead of `i, j, k` (alphabetical order) for the loop index variables?

Week3/homework/7-step3.js
```diff
@@ -22,4 +22,15 @@ f2(y);
...
+the value is passed in by value.This means that any changes to that variable
+while in the function are completely separate from anything that happens outside the function.
+*/
+
+/*
+Passing in an object, however,
+passes it in by reference. In this case, any property of that object is accessible within the function
+*/
+
+//via https://snook.ca/archives/javascript/javascript_pass
```
remarcmij: I applaud that you included a link to the source where found this information. I assume you also understand what it stated here.

Week3/homework/step4-bonus.js
```diff
@@ -1,10 +1,61 @@
...
+{
+  const values = ['a', 'b', 'c', 'd', 'a', 'e', 'f', 'c'];
+
+  const unique = [...new Set(values)];
+
+  console.log(unique);
+}
+
+
+// Note: This step was made with EXTREME googling :)
```
remarcmij: We developers google a lot to help us write our code. However, you must always be able to explain the code that you are using. Otherwise we're doing a sort of bismillah-type of programming. Can you explain all these versions yourself?

Which version is the most easy to explain for yourself?

PR #141, opened 2018-10-10
Hiba HW week3 JS2

heboshka:


Week3/homework/2-step3.js
```diff
@@ -1,8 +1,22 @@
...
+}
+function sayFive(number) {
+  console.log(number + "  Is divisible by five");
+}
 
 function threeFive(startIndex, stopIndex, threeCallback, fiveCallback) {
   const values = [];
-  // Add your code here
+  for (let z = startIndex; z <= stopIndex; z++) {
+    values.push(z);
```
remarcmij: You are not completely following the assignment: it stated that creation of the array and testing for divisibility by 3 and 5 are to be done in two separate steps:

>The function should first generate an array containing values from start value to end value (inclusive).
>
>Then the function should take the newly created array and iterate over it, and calling the first callback if the array value is divisible by 3.

Apart from that, everything is OK here.

Week3/homework/3-step3.js
```diff
@@ -2,23 +2,39 @@
...
-  // add your code here
-  return str;
+  let newStr = str;
+  if (num > 0) {
+    for (let i = 2; i <= num; i++) {
+      str += newStr;
+    }
+    return str;
+  }
+  else { return str = "" }
```
remarcmij: If the `if` block is executed it will be never consider the `else` block (because the `if` block ends in `return`).  Therefore line 12 might simply be: 

```js
return "";
```

Week3/homework/3-step3.js
```diff
@@ -2,23 +2,39 @@
 
 // use a 'for' loop
 function repeatStringNumTimesWithFor(str, num) {
-  // add your code here
-  return str;
+  let newStr = str;
```
remarcmij: The `let` could have been a `const` (preferred).

Week3/homework/3-step3.js
```diff
@@ -2,23 +2,39 @@
...
   return str;
 }
 
 console.log('while', repeatStringNumTimesWithWhile('abc', 3));
 
 // use a 'do...while' loop
 function repeatStringNumTimesWithDoWhile(str, num) {
-  // add your code here
+  let newStr = str;
+  if (num > 1) {
```
remarcmij: This need to add an extra test at the top of the `do-while` loop illustrates nicely why the `do-while` loop is not so useful. I never use it myself. But still, you need to know about it.

Week3/homework/4-step3.js
```diff
@@ -1,2 +1,8 @@
 'use strict';
-// paste your freeCodeCamp solutions in here
+function Dog() {
+  this.name = "Doogy";
+  this.color = "Brown";
+  this.numLegs = 4;
+}
+let hound = new Dog();
```
remarcmij: This `let` could be a `const`.

Week3/homework/6-step3.js
```diff
@@ -4,3 +4,16 @@ const arr2d = [[1, 2], [3, 4], [5, 6]];
...
+  }
+}
+
+for (let x = 0; x < arr3d.length; x++) {
+  for (let z = 0; z < arr3d[x].length; z++) {
+    for (let v = 0; v < arr3d[x][z].length; v++) {
+      console.log(arr3d[x][z][v]);
+    }
+  }
+}
```
remarcmij: See the zip file I posted in slack how to do this for an array of _any_ dimension.

Week3/homework/step4-bonus.js
```diff
@@ -2,9 +2,14 @@
 
 const values = ['a', 'b', 'c', 'd', 'a', 'e', 'f', 'c'];
 
-// Add your function here. Try and come up with a good name for this function
-
-// Replace `yourFunction` with the name of the function you just created
-const uniqueValues = yourFunction(values);
-
+function uniqueValuesFinder() {
+  let uniqueValueArr = [];
```
remarcmij: This `let` could be a `const`.

Week3/homework/step4.js
```diff
@@ -1,6 +1,12 @@
 'use strict';
 
 // Add your code here
+function createBase(num) {
+  return function (num1) {
+    let i = num + num1;
+    return i;
```
remarcmij: The variable `i` doesn't really add anything. I would just simply return the result directly:

```js
return num + num1;
```

PR #146, opened 2018-10-12
week3

salar01:


Week3/homework/2-step3.js
```diff
@@ -1,8 +1,28 @@
...
 
 function threeFive(startIndex, stopIndex, threeCallback, fiveCallback) {
   const values = [];
-  // Add your code here
+  for (let i = startIndex; i <= stopIndex; i++) {
+    values.push(i);
+  }
+  console.log(values);
+  for (let i = 0; i < values.length; i++) {
+    if (values[i] % 3 === 0 && values[i] % 5=== 0 ) {
```
remarcmij: It is unnecessary to test for the combination of divisibility by 3 and 5 at the same time. This is handled automatically when you you test for divisibility by 3 and by 5 separately:

```js
if (values[i] % 3 === 0) {
  threeCallback(values[i]);
}
if (values[i] % 5 === 0) {
  fiveCallback(values[i]);
}
```

Week3/homework/3-step3.js
```diff
@@ -1,25 +1,49 @@
-'use strict';
+"use strict";
 
 // use a 'for' loop
 function repeatStringNumTimesWithFor(str, num) {
-  // add your code here
-  return str;
+  if (num <= 0) return "";
```
remarcmij: If your `if` statement return, there is no need for an `else`.

Also, it is best to always use curly braces for an `if` block:

```js
if (num <= 0) {
  return "";
}
```

However, you do not need the `if` statement at all, because the loop body will not be executed if `num` is less or equal to zero:

```js
  let result = "";
  for (let i = 0; i < num; i++) {
    result += str;
  }
  return result;
}
```

Week3/homework/3-step3.js
```diff
@@ -1,25 +1,49 @@
...
-console.log('for', repeatStringNumTimesWithFor('abc', 3));
+repeatStringNumTimesWithFor("abc", 3);
+
+console.log("for", repeatStringNumTimesWithFor("abc", 3));
 
 // use a 'while' loop
 function repeatStringNumTimesWithWhile(str, num) {
-  // add your code here
-  return str;
+  if (num <= 0) {
```
remarcmij: As per my previous comment, no need for this `if` statement.

Week3/homework/3-step3.js
```diff
@@ -1,25 +1,49 @@
...
 }
-
-console.log('while', repeatStringNumTimesWithWhile('abc', 3));
+console.log("while", repeatStringNumTimesWithWhile("abc", 3));
 
 // use a 'do...while' loop
 function repeatStringNumTimesWithDoWhile(str, num) {
-  // add your code here
-  return str;
+  if (num <= 0) {
```
remarcmij: Contrary to the other two cases, here you need to test for `num <= 0`, because a `do-while` loop tests the condition at the end of the loop. Therefore it may execute the loop even if the loop condition does not hold true. Which makes the `do-while` loop rather useless in practice. I never use it myself.

Week3/homework/4-step3.js
```diff
@@ -1,2 +1,14 @@
...
-// paste your freeCodeCamp solutions in here
+function Dog() {
+  this.name = "Rupert";
+  this.color = "brown";
+  this.numLegs = 4;
+}
+
+let hound = new Dog();
+////////////////
+function Dog() {
```
remarcmij: This is a redefinition of the `Dog()` constructor function. You should never redefine a function in the same scope. At minimum it is a bad practice; usually it is considered an error.

So what value do you think  the` hound` variable has? Is it a`Rupert` or  a `Jacky`?

Week3/homework/5-step3.js
```diff
@@ -1,2 +1,17 @@
...
+function multiplyAll(arr) {
+  var product = 1;
+  for (var i = 0; i < arr.length; i++) {
+    for (var j = 0; j < arr[i].length; j++) {
+      console.log(arr[i][j]);
+      product = product * arr[i][j];
+       console.log(product * arr[i][j]);
+    }
+  }
+  return product;
```
remarcmij: The console.log statements inside the function body are not telling me much. I'm far more interested in the final product, but you don't console.log, e.g.:

```js
```

Week3/homework/5-step3.js
```diff
@@ -1,2 +1,17 @@
...
+}
+
+// Modify values below to test your code
+multiplyAll([[1], [2], [3]]);
+multiplyAll([[1, 2], [3, 4], [5, 6, 7]]);
+multiplyAll([[5, 1], [0, 2, 4, 0.5], [3, 9]]);
```
remarcmij: This third array has issues because it has three sub-arrays of unequal length, respectively 2, 3 and 2. Furthermore, because one of the array elements is zero, the final product will also be zero.

Week3/homework/6-step3.js
```diff
@@ -1,6 +1,20 @@
...
+"use strict";
 
 const arr2d = [[1, 2], [3, 4], [5, 6]];
 const arr3d = [[[1, 2], [3, 4]], [[5, 6], [7, 8]]];
 
-// add your solution here, or add a comment on how you would tackle this problem
+for (let i = 0; i < arr3d.length; i++) {
+  for (let j = 0; j < arr3d[i].length; j++) {
+    for (let k = 0; k < arr3d[i][j].length; k++) {
+
```
remarcmij: These `for` loops don't produce anything.

Week3/homework/6-step3.js
```diff
@@ -1,6 +1,20 @@
...
+    }
+  }
+}
+
+// With for of
+
+for (let item of arr3d) {
```
remarcmij: This `for` loop prints just the elements of the outer array. These elements happen to be elements themselves. I expected each individual value of the array to be printed out by your `for` loops.

Week3/homework/7-step3.js
```diff
@@ -9,6 +9,12 @@ function f1(val) {
 f1(x);
 
 console.log(x);
+/* Here we have created a variable and assigned it to a value x 9. Then we created a function
+   and we pass a value to it, so the result should be 10, but here we just returned the 
+   value of constant x that has value 10. */
```
remarcmij: The constant `x` has the value 9. We couldn't even change it to something else because we used `const` to declare it.

Week3/homework/7-step3.js
```diff
@@ -20,6 +26,7 @@ function f2(val) {
 f2(y);
 
 console.log(y);
-
-// Add your explanation as a comment here
+/* Here we have an object and we have changed the value of the property of X 9 
+   to value X 10, so we can change the property even if the object is constant.
+ */
```
remarcmij: This is correct.

Week3/homework/step4-bonus.js
```diff
@@ -2,9 +2,10 @@
...
+  let newArr = Array.from(new Set(values));
+  return newArr;
+}
 
-// Replace `yourFunction` with the name of the function you just created
-const uniqueValues = yourFunction(values);
-
-console.log(uniqueValues);
+/* Second way */
+let uniqueValues = [... new Set(values)];
```
remarcmij: This looks like google work. Can you explain how it works?

Because you have used the same name for the function and the variable, I can not test it without modifying your code. That puts the burden on the reviewer (me). Make sure that I can run your code without modification.

Week3/homework/step4.js
```diff
@@ -1,7 +1,17 @@
...
+    return num + item;
+  }
+}
 
+const addSix = createBase(6);
+addSix(10); // returns 16
+addSix(21); // returns 27
+
+/* Second way to structuring the Function*/
+createBase = base => console.log(base + num);
```
remarcmij: See my comment above about reusing names.

This second version does something different (it use console.log) than the first version.

PR #148, opened 2018-10-13
 add homework of week3

Mohamad-AezAldin:
please check the homework of week3

### DIFF
--- a/Week3/homework/step4.js
+++ b/Week3/homework/step4.js
@@ -2,7 +2,10 @@
 
 // Add your code here
 
-const addSix = createBase(6);
+const addSix = function createBase(num) {
+  const baseNum = 6;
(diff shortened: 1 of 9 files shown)


Week3/homework/step4.js
```diff
@@ -2,7 +2,10 @@
 
 // Add your code here
 
-const addSix = createBase(6);
+const addSix = function createBase(num) {
```
remarcmij: The `createBase(num)` function should return a new function that uses the `num` (or whatever name you use for that argument) as the base. So:

```js
const addSix = createBase(6);
const addSeven  = createBase(7);
console.log(addSix(10)); // --> 16
console.log(addSeven(10)); // --> 17
```

PR #150, opened 2018-10-13
homeweork js2 week3 Karam

KaramGhazzi:


Damla78:
Hi Karam. I'm writing for JavaScript3 - week1 feedback homework. All code is nice but only in
3-step3.js
in while loop you start with space.

Week3/homework/6-step3.js
```diff
@@ -3,4 +3,19 @@
...
+  for (let j = 0; j < arr2d[i].length; j++) {
+    console.log(arr2d[i][j]);
+  }
+}
+
+// print out three dimension array
+for (let i = 0; i < arr3d.length; i++) {
+  for (let j = 0; j < arr2d[i].length; j++) {
+    for (let x = 0; x < arr3d[i][j].length; x++) {
+      console.log(j, arr3d[i][j][x]);
```
remarcmij: Printing out `j` here, in addition to the array value makes the output confusing. We are only interested in the array value.

Week3/homework/6-step3.js
```diff
@@ -3,4 +3,19 @@
...
+for (let i = 0; i < arr2d.length; i++) {
+  for (let j = 0; j < arr2d[i].length; j++) {
+    console.log(arr2d[i][j]);
+  }
+}
+
+// print out three dimension array
+for (let i = 0; i < arr3d.length; i++) {
+  for (let j = 0; j < arr2d[i].length; j++) {
+    for (let x = 0; x < arr3d[i][j].length; x++) {
```
remarcmij: It is customary when using nested loops likes this to use consecutive letters from the alphabet, starting with the letter `i`, e.g. `i`, `j`, 'k` etc. So in this case, I would use `k` instead of `x`.

Week3/homework/6-step3.js
```diff
@@ -3,4 +3,19 @@
...
+      console.log(j, arr3d[i][j][x]);
+    }
+  }
+}
```
remarcmij: See the zip file I posted in slack for a version that can print an array of any dimension by making use of recursion.

PR #151, opened 2018-10-20
Giselle homework week1 - Book Store

gisnetz:


gijscor:
Hi @gisnetz I think you handed in your homework at the wrong place.

PR #166, opened 2018-11-14
homework js2 wk2

mohamad-monir:
hope some one can review it

### DIFF
--- a/Week2/homework/maartjes_work.js
+++ b/Week2/homework/maartjes_work.js
@@ -45,3 +45,8 @@ const tuesday = [
 const tasks = monday.concat(tuesday);
 
 // Add your code here
+const minArr = tasks.map((x) => x.duration);
+const hoursArr = minArr.map((y) => y / 60).filter((y) => y >= 2);
+const perHour = hoursArr.map((x) => x * 10);
+const sum = perHour.reduce((x) => x + x);
+console.log(sum);
--- a/Week2/homework/map_filter.js
+++ b/Week2/homework/map_filter.js
@@ -3,3 +3,5 @@
 const numbers = [1, 2, 3, 4];
 
 // Add your code here
+const newNum = numbers.filter(x => x % 2 !== 0).map(x => x * 2);
+console.log(newNum);


Week2/homework/maartjes_work.js
```diff
@@ -45,3 +45,8 @@ const tuesday = [
 const tasks = monday.concat(tuesday);
 
 // Add your code here
+const minArr = tasks.map((x) => x.duration);
+const hoursArr = minArr.map((y) => y / 60).filter((y) => y >= 2);
+const perHour = hoursArr.map((x) => x * 10);
```
rohni: Poor Maartje, she is not being paid very much. 😛

Week2/homework/maartjes_work.js
```diff
@@ -45,3 +45,8 @@ const tuesday = [
...
+const sum = perHour.reduce((x) => x + x);
```
rohni: This use of reduce is a bit incorrect.
I would suggest reading
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/Reduce

`const sum = perHour.reduce((acc, x) => acc + x, 0);`
The zero is optional, but generally good form, because there are lots of times it is important to have an initial value.

Week2/homework/maartjes_work.js
```diff
@@ -45,3 +45,8 @@ const tuesday = [
...
+console.log(sum);
```
rohni: According to the Makeme.md, this was supposed to be formatted an a currency format.
I would suggest having a look at the string method `toLocaleString`  it is very powerful.

Week2/homework/map_filter.js
```diff
@@ -3,3 +3,5 @@
 const numbers = [1, 2, 3, 4];
 
 // Add your code here
+const newNum = numbers.filter(x => x % 2 !== 0).map(x => x * 2);
```
rohni: Very nice.  Works perfectly. 
As you seem to enjoy reduce (as seen above) here is what I put together (using the `spread` operator, look it up 😛 )
```
const redDouble = numbers.reduce((acc, n) => (n % 2 !== 0 ? [...acc, n * 2] : acc), []);
```

PR #169, opened 2018-11-16
Burhan class18 js2 week1

burhankhuzam:
Delivering the Homework.


Week1/homework/app.js
```diff
@@ -1,11 +1,51 @@
...
+    { 'title': 'The Messy Middle', 'language': 'English', 'author': 'Scott Belsky' },
+    { 'title': 'Lake Success', 'language': 'English', 'author': 'Gary Shteyngart' },
+    { 'title': 'Fashion Climbing', 'language': 'English', 'author': 'Bill Cunningham' },
+    { 'title': 'Crudo', 'language': 'English', 'author': 'Olivia Laing' },
+    { 'title': 'Winners Take All', 'language': 'English', 'author': 'Anand Giridharadas' },
+    { 'title': 'French Exit', 'language': 'English', 'author': 'Patrick deWitt' },
+    { 'title': 'How To Be Famous', 'language': 'English', 'author': 'Caitlin Moran' }
   ];
 
+  function makeList() {
```
rohni: This is purely a personal call, but I would have broken all the functionality in this function into a few different functions. (Actually, I did do that when I did the homework. 😄 ).
It allows you to test each part of the layout code separately.

Week1/homework/app.js
```diff
@@ -1,11 +1,51 @@
...
-  // Replace with your own code
-  console.log(bookTitles);
-}
+    let ul = document.createElement('ul');
+    b1.appendChild(ul);
+
+    for (let i = 0; i < bookTitles.length; i++) {
+
+      let li = document.createElement('li');
+      li.setAttribute("id", "book" + [i]);
```
daniela-olaru: I think you meant `"book"+i` (without the [ ]). The construct `[i]` creates an Array, which, luckily, when concatenated with a string does output the intended "book0", "book1" and so on, however, it is not needed

Week1/homework/app.js
```diff
@@ -1,11 +1,51 @@
...
+      let div = document.createElement('div');
+      li.appendChild(div);
+      div.setAttribute('id', 'cover');
+      div.style.width = "250px";
+      div.style.height = "350px";
+
+
+      let h1 = document.createElement('h1');
+      li.appendChild(h1);
+      h1.innerHTML = (Object.values(bookTitles)[i]).title;
```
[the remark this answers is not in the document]
rohni: Actually, using `Object.values(bookTitles)[i]` is incorrect.  The reason is that Objects make no guarantees about the order that items are returned.
You are relying on the same order being returned from each Object.values() call.  Which, luckily it did, but might not in another browser, or at another time in the same browser.
My advice about setting the `currentBook` and using that all for all the code in the loop would ensure that the same book object was being used for each field access.

PR #170, opened 2018-11-16
my home work week1 js2

soumaiamamo:


Week1/homework/app.js
```diff
@@ -1,11 +1,128 @@
...
+
+  let theHeaderContent = document.createTextNode ('My library ');
+  myHeader.appendChild(theHeaderContent);
+
+for (let i=0; i<myBooksName.length; i++){
+let myLi=document.createElement('li');
+//myLi.textContent = myBooksName[i];
+myUl.appendChild(myLi);
+
+let bookName= document.createTextNode(Object.values(details)[i].title);
```
daniela-olaru: One thing to keep in mind, is that the elements in the array `myBooksName` are the same as the keys in the `details` object. So in order to quickly access information about a specific key, all you have to do is `details[myBooksName[i]]`. For example, let's take the case where where **i = 0** : 
1. _myBooksName[0]_ is "milk_and_vine" 
2. _details["milk_and_vine"]_ is { title:"Milk and Vine ", author: ... }
----------
so by combining 1. and 2. you get `details[myBooksName[0]]` => this means that if the first line your `for` block is 
```
for (let i=0; i<myBooksName.length; i++){
    const myDetail = details[myBooksName[i]];
    // .. here you can use myDetail.title, myDetail.author and myDetail.language
}
```

... thus avoiding the whole construct with _Object.values(...)_

Week1/homework/app.js
```diff
@@ -1,11 +1,128 @@
...
+let myH1 = document.createElement("h2");
+let myH3 = document.createElement("h3");
+let myH4 = document.createElement("h3");
+
+myH1.appendChild(bookName);
+myH3.appendChild(bookAuthor);
+myH4.appendChild(bookLanguage);
+myLi.innerHTML =  myH1.outerHTML + myH3.outerHTML + myH4.outerHTML ;
+
+let bookCovers = document.createElement("img");
```
daniela-olaru: This is very good and it works! 

But just for the sake of following the homework assignment, this part specifically : 
> Loop over these entries (hint: Object.keys(objectName) gives you an array containing the keys). Then write a function which places an image at the corresponding li element. Remember that objects are not ordered, so you cannot guarantee that the first key is the first li element. (Hint: you could give each li item an id tag by modifying the function you made before.)

After the first _for_ loop, the assignment said to do another loop through the keys of `myBookImages` and add the image to the corresponding _li_. In order to do this, you would need to add an _id_ attribute to each _li_, probably the key in `details` object itself (which correspond with keys in `myBookImages`). Then, the second loop would go something like this:

```javascript
let imageKeys = Object.keys(myBookImages);
for(i = 0; i < imageKeys.length; i++){
   let theLi = document.getElementById(imageKeys[i]);
   if (theLi){
      // the code you already wrote here
      let bookCovers = document.createElement("img");
      bookCovers.setAttribute("src", myBookImages[imageKeys[i]]);
      bookCovers.setAttribute("alt", details[imageKeys[i]].title);
      theLi.appendChild(bookCovers);
   }
}
```
and avoiding again the _Object.values(...)_ construct

PR #173, opened 2018-11-17
Mohamad class18 js2 wk1

mohamad-monir:
homewok js2 wk1


daniela-olaru:
I see that in the last commit you are deleting _app.js_. Why is that?

mohamad-monir:
@daniela-olaru  no reason just play with github to see what is the possibility

PR #174, opened 2018-11-17
Gaffari class18 js2 wk1

Gaffari02:


Week1/homework/app.js
```diff
@@ -1,11 +1,157 @@
...
+    het_diner : "images/het diner.jpg"
+    };
+ 
+  function bookList () { 
+
+    let bookcontainer = document.getElementById("mybooks");
+    let div = document.createElement("div");
+    bookcontainer.appendChild(div);
+    let li = document.createElement("li");
+    div.appendChild(li);
```
rohni: I'm pretty sure you wanted this to be a `ul` tag.  It is the container for the rest of the list.

Week1/homework/app.js
```diff
@@ -1,11 +1,157 @@
...
+    
+
+
+    for (let i = 0; i< bookTitles.length; i++) {
+
+      // let content = document.createTextNode(bookTitles[i]);
+      // li.appendChild(content);
+      let titleList = document.createElement("li")
+      li.appendChild(titleList);
+      let titleContent = document.createTextNode(bookDetails[bookTitles[i]].title);
```
rohni: Well done.  You are accessing the values of the data structures correctly.  `bookDetails[bookTitles[i]]`.
Another idea to help readability would be to define a const with the current book, and then use that (and it saves typing)
e.g.
```
for(....){
  const currentBook = bookDetails[bookTitles[i]];
....
...   currentBook.title
...
```
It's an idea... 😄

Week1/homework/app.js
```diff
@@ -1,11 +1,157 @@
...
+      li.appendChild(languageList);
+      let languageContent = document.createTextNode("Language: " + bookDetails[bookTitles[i]].language);
+      languageList.appendChild(languageContent);
+      let authorList = document.createElement("li")
+      li.appendChild(authorList);
+      let authorContent = document.createTextNode("Author: "+ bookDetails[bookTitles[i]].author);
+      authorList.appendChild(authorContent);
+      let img = document.createElement("img");
+      li.appendChild(img);
+      img.src = Object.values(bookImg)[i];
```
rohni: Huh, I'm surprised to see this, after you were approaching the data access above correctly.

Using `Object.values(bookImg)[i]` hoping to get the same entry is incorrect.  If `bookImg` had been an array of images, in the same order as the `bookTitles` you would have been correct, and would have referred to the correct image like this:
```
img.src = bookImg[i];
```
Objects are a key/value store, and you access the items in the object by referring to the key.  So, if I want to access 
```
"images/Hannas' verjaardag.jpg"
```
I have to get it by asking the object for the value of the `hannas_verjaardag` key.
When I call `Object.values` it gives me back all the items in an array, but the order of the items in the array is not guaranteed.
So, you call it hoping that it will give you back the values in the order you have defined them, but that is not something you can rely on.  As a result, it is very possible that the image displayed for your book is the wrong one. 
You should have accessed the image in the same way you accessed the details. i.e.
```
img.src = bookImg[bookTitles[i]];
```
That would have guaranteed the correct image for each book.

Week1/homework/app.js
```diff
@@ -1,11 +1,157 @@
...
+      
+      
+  
+
+      li.setAttribute("class", "bookList");
```
rohni: This is nicely laid out, makes reading which style classes go with which part of the book section very clear.  Well done!

Gaffari02:
Dear @rohni, thanks for time and effort that you have spent while reviewing my homework. I will work on the advices that you have given.

PR #176, opened 2018-11-18
Keyya class18 js2 week1

keyyaali:


Week1/homework/index.html
```diff
@@ -1 +1,4 @@
-<!-- replace this with your HTML content -->
\ No newline at end of file
+<!-- replace this with your HTML content -->
```
rohni: Okay, I'm not sure if you have seen this or not, but I checked out the branch, and went to the index.html and saw....
....
........  nothing. 
You have done your work in the master branch, not the feature branch.
I would suggest you get familiar with `git` and how to use it smoothly, to branch, and merge, push etc.   Ask a classmate who seems to understand how it works to help you and walk you through how this works.
It is a bit difficult to give you feedback here as there is nothing to see.

To make it worse, the only file with any content in this pull request is for week2.
Consider this a lesson on checking that what you think you did, is what actually happened.  With computers, this will happen all the time to you, and unfortunately, computers and their software are not yet powerful enough to catch this sort of error for you. 😄

PR #177, opened 2018-11-19
Yaacob class18 js2 wk2

SignosRk:
you can check only the week2 part in this repository(commits on November 19), still adds the changes of the week1 homework because the master repository from HyF doesn't have them applied so you will see the whole history of homework.


Week2/homework/map_filter.js
```diff
@@ -3,3 +3,14 @@
...
+// const newNumbers = [];
+// for (let i = 0; i < numbers.length; i++) {
+//     if (numbers[i] % 2 !== 0) {
+//       newNumbers.push(numbers[i] * 2);
+//     }
+//   }
+//   console.log(newNumbers);
+
+const newNum = numbers.filter(numbers => numbers % 2 !== 0).map(numbers => numbers * 2);
+console.log(newNum);
```
yash-kapila: Hi @SignosRk, I would recommend that when you think your code is ready to be pushed, remove all the `console.log` statements you might have added. It helps you as a developer primarily as a debugging tool but need not be pushed. 🙂

Week2/homework/map_filter.js
```diff
@@ -3,3 +3,14 @@
...
+
+// const newNumbers = [];
+// for (let i = 0; i < numbers.length; i++) {
+//     if (numbers[i] % 2 !== 0) {
+//       newNumbers.push(numbers[i] * 2);
+//     }
+//   }
+//   console.log(newNumbers);
+
+const newNum = numbers.filter(numbers => numbers % 2 !== 0).map(numbers => numbers * 2);
```
yash-kapila: This is good but I would suggest that next time maybe think of taking another step forward and try to wrap this logic inside a function. For example, having a function(`doubleOddNumbers`) like below would allow you to pass an array of any numbers and another developer who might need a similar behaviour can use your function without having to repeat the logic.
```
const doubleOddNumbers = numbers => numbers
    .filter(numbers => numbers % 2 !== 0)
    .map(numbers => numbers * 2);
```

Week2/homework/maartjes_work.js
```diff
@@ -45,3 +45,23 @@ const tuesday = [
...
+console.log(hours);
+
+// part 2 putting everything out lower than 2 hours
+
+const moreThan2 = hours.filter(hours => hours >= 2);
+console.log(moreThan2); 
+
+// part 3 Multiply the each duration by a per-hour rate for billing (* 20)
+
+const earning = moreThan2.map(hours => hours *20).reduce((total, amount) => total + amount);
```
yash-kapila: I really liked that you used reduce to sum all the elements of the array. 👍

Week2/homework/maartjes_work.js
```diff
@@ -45,3 +45,23 @@ const tuesday = [
 const tasks = monday.concat(tuesday);
 
 // Add your code here
+
+// part 1 mapping to hours
+
+const hours = tasks.map(tasks => tasks.duration / 60);
```
yash-kapila: I would like to ask you to remove all the `console.log` statements as we don't really need them once we have got working code ready.

SignosRk: it is done my friend, removed the console.log... and I left the last one commented because is part of the homework, because it resume the whole amount to two decimals.

PR #178, opened 2018-11-20
Moataz class18 js2 week2

Moataz7:


Week2/homework/map_filter.js
```diff
@@ -3,3 +3,7 @@
 const numbers = [1, 2, 3, 4];
 
 // Add your code here
+let newNumbers = numbers.filter((number) => number % 2 !== 0);
```
yash-kapila: Hi @Moataz7 - we don't really need to use `let` here to create the `newNumbers` variable. Try to make use of `const` please. Instead of doing the filter & map operations in two different statements, you can try to use method chaining using the `dot(.)` operator and then you can assign the output of it to the `const` newNumbers.

I would also recommend that you try to use `const` as much as possible. Since it is a non-changing(almost a constant) variable, we can be assured that the value we assign to it is going to remain the same throughout. If I use `let` here, in the future, you or another developer can accidentally change its value which can then potentially cause bugs in our app.

Also, if we are passing a single argument using arrow function, we don't need `paranthesis()`. You can have `filter(number => number % 2 !== 0);` instead. Same goes for the map operation below. 🙂

Week2/homework/map_filter.js
```diff
@@ -3,3 +3,7 @@
...
+newNumbers = newNumbers.map((number)  => number*2);
```
yash-kapila: A tiny spacing issue here please - `number*2`

Week2/homework/maartjes_work.js
```diff
@@ -45,3 +45,14 @@ const tuesday = [
 const tasks = monday.concat(tuesday);
 
 // Add your code here
+//console.log(tasks);
+let tasksHours = tasks.map((task)  => task.duration/60);
```
yash-kapila: I like that you created individual variables which hold different states of information instead of having a single variable trying to hold all values at different stages of time. However, we don't need to use `let` keyword here to create those variables. Please try to use `const` here for all variable initialisations.

Also, please remove the extra parentheses around the arguments being passed in the `map` & `filter` functions.

Week2/homework/maartjes_work.js
```diff
@@ -45,3 +45,14 @@ const tuesday = [
...
+//console.log(tasksHours);
+let tasksHoursfiltered = tasksHours.filter((taskhour) => taskhour >= 2 );
+//console.log(tasksHoursfiltered);
+let tasksPrice =tasksHoursfiltered.map((taskfiltered)  => taskfiltered*10);
+//console.log(tasksPrice);
+let salary = tasksPrice.reduce((x, taskPrice) => x + taskPrice, 0);
```
yash-kapila: Could you please rename `x` to something more meaningful like `total` or `sum`? It is quite useful to understand code with such names. 🙂 

Also, using reduce this way works perfectly fine but I would like to share another thing with you. If you don't specify an initial value of the `accumulator` which is `x` in your case, it will be automatically assigned the first element of the array and `taskPrice` would then point to the second element of the array. 

For example, if the array is like `[10,20,30,40]` and when we write reduce like you have done above, `x` will start with value `0` and `taskPrice` would be `10`. However, if you don't specify the `0 at the end, then `x` will start with value `10` and `taskPrice` would be `20` and your result would be the same in both cases. You can try it later.

PR #179, opened 2018-11-20
Wael Alhomsi class18 js2 week2 homework

Wael-Alhomsi:


Week2/homework/maartjes_work.js
```diff
@@ -45,3 +45,16 @@ const tuesday = [
 const tasks = monday.concat(tuesday);
 
 // Add your code here
+
+function computeAmountEarned(arrayOfTasksAndDurations, wagePerHour) {
```
yash-kapila: I like the way you have written this function and used method chaining. Couple of tiny improvements that you can still make:
- The name of function arguments need not describe the type of the argument i.e `arrayOfTasksAndDurations` need not have `array` in the name. Quite like how you have named the actual function like `computeAmountEarned`.
- You can make use of the => syntax to define your function.

PR #183, opened 2018-11-22
Gaffari class18 js2 wk2

Gaffari02:


Week2/homework/const.js
```diff
@@ -0,0 +1,16 @@
...
+    return number % 2 !== 0;
+});
+console.log(oddNumbers);
+
+const newNumbers = oddNumbers.map(number => {
+    return number * 2;
+});
+
+console.log('The doubled numbers are', newNumbers);
+
```
rohni: Nice job making this testable by breaking everything up into its own constant.
You can also chain the two operations together.
```
numbers.filter(number => number % 2 !== 0)
    .map(number => number * 2);
```

Week2/homework/maartjes_work.js
```diff
@@ -44,4 +44,22 @@ const tuesday = [
...
+const moreThan2Hours = hours.filter(x => {
+  return x >= 2;
+});
+
+//step 3 total salary
+const salary = moreThan2Hours.map(earn => earn * 17)
+.reduce((accumulator, currentValue) => accumulator + currentValue);
+
+//step 4 rounded salary
+console.log(`€${salary.toFixed(0)}`);
```
rohni: I would be inclined to make the parameter to `.toFixed` 2, so you have more of a currency feel with two decimal places.

Gaffari02:
Thank you @rohni, I will work on the advices that you have given.

PR #185, opened 2018-11-23
Hw js2 week2

soumaiamamo:


Week2/homework/map_filter.js
```diff
@@ -2,4 +2,15 @@
 
 const numbers = [1, 2, 3, 4];
 
-// Add your code here
+
+const newNumber = numbers.filter( number => number % 2 !== 0);
+console.log(newNumber);
```
yash-kapila: Please remove the extra `console.log` statements. Since they are only used for debugging by a developer, they should be removed if you think the code is ready to be pushed. 🙂

Week2/homework/map_filter.js
```diff
@@ -2,4 +2,15 @@
...
+ 
+const mapping = newNumber.map(number =>number * 2);
```
yash-kapila: A very tiny spacing correction needed here 🙂

Week2/homework/maartjes_work.js
```diff
@@ -45,3 +45,30 @@ const tuesday = [
 const tasks = monday.concat(tuesday);
 
 // Add your code here
+ let durationInHours = tasks.map(hours => hours.duration / 60 );
```
yash-kapila: I like that you created individual variables which hold different states of information instead of having a single variable trying to hold all values at different stages of time. However, we don't need to use `let` keyword here to create those variables. Please try to use `const` here for all variable initialisations.

The reason being that `map`, `filter` and `reduce` methods create a new array everytime and hence the variables `durationInHours`, `moreThanTwo` and others don't change their values. 🙂

Week2/homework/maartjes_work.js
```diff
@@ -45,3 +45,30 @@ const tuesday = [
...
+//(5) [3, 3.3333333333333335, 4, 3, 3.3333333333333335] 
+
+let salary = moreThanTwo.map(hour => hour * 10 );
+
+let amount = salary.reduce((getsum , total) => getsum + total );
+console.log(amount);
+
+//166.66666666666669
+
+let theSalary = '€ ' +  amount.toFixed(2);
```
yash-kapila: I am not sure if this has been brought up in a class by a teacher before but if you wish to do a concatenation like this, you can explore template literals - https://developer.mozilla.org/nl/docs/Web/JavaScript/Reference/Template_literals

Could you try to do the same please? 🙂

PR #187, opened 2018-11-24
Hallo! This my homework for week2 JS2. Alek

7more10:


Week2/homework/map_filter.js
```diff
@@ -1,5 +1,69 @@
-'use strict';
```
rohni: You seem to have squashed both exercises into the same file.  Not sure why. 😄

Week2/homework/map_filter.js
```diff
@@ -1,5 +1,69 @@
...
-
 const numbers = [1, 2, 3, 4];
 
-// Add your code here
+for (let i = 0; i < numbers.length; i++) {
```
rohni: This is not needed.  The `filter` and `map` methods look through the array automatically.
What you have done is loop through the array 4 * (1 (filter) + 1 (map)).  And that is why the results are shown 4 times.
This would do the trick:
```
const newNumbers = numbers.filter(number => number % 2);
const finalNumbers = newNumbers.map(number => number * 2);
console.log('The final numbers are ' + finalNumbers);
```
`map`, `filter` and `reduce` will be able to replace almost all your needs for `for`, `for...in`, and `while`, `do...while` etc. looping structures.

Week2/homework/map_filter.js
```diff
@@ -1,5 +1,69 @@
...
+  console.log(tasks);
+
+  const durationInHours = tasks.map(tasks => tasks.duration / 60);
+  console.log(durationInHours);
+
+  const lessThanTwoTasks = durationInHours.filter(hour => hour >= 2);
+  console.log(lessThanTwoTasks);
+
+  const ratePerHourInEuro = 10;
+
```
rohni: Some of the indenting is a bit off.  You should probably have the `formatOnSave` setting set to true in VSCode. (I assume you are using VSCode, if not, I suggest asking another classmate to help you get set up with it).
If you are using another editor, and want to stay there, no problem, but code formatting needs to be handled and configured for the editor.  Most editors handle the `.editorconfig` file properly.

Week2/homework/map_filter.js
```diff
@@ -1,5 +1,69 @@
...
+  const lessThanTwoTasks = durationInHours.filter(hour => hour >= 2);
+  console.log(lessThanTwoTasks);
+
+  const ratePerHourInEuro = 10;
+ 
+  function arraySum(array){
+    let sum = 0;
+    for(let i = 0; i < array.length; i++){
+    sum += array[i] * ratePerHourInEuro;
+    }
```
rohni: Nice, this works perfectly.
You can also do this using reduce (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/Reduce),   and it looks like this:
```
const summed = lessThanTwoTasks.reduce((total, item) => total + item);
```

PR #192, opened 2018-11-28
Zerisenay Solomon HW JS Week3

ZerSolo:


Week3/homework/1-step3.js
```diff
@@ -1,11 +1,8 @@
-'use strict';
-
 function foo(func) {
-  // What to do here? 
+  return func()
```
rohni: Excellent!

Week3/homework/2-step3.js
```diff
@@ -2,7 +2,35 @@
...
+  console.log('the number ' + number + ' is divisible by 5')
+
 }
 
-threeFive(10, 15, sayThree, sayFive);
+function sayThree(number) {
+  console.log('the number ' + number + ' is divisible by 3')
+
+}
+console.log(threeFive(10, 15, sayThree, sayFive))
```
rohni: Excellent!

Week3/homework/3-step3.js
```diff
@@ -1,25 +1,41 @@
...
+  do {
+    stringAccumlator += str;
+    num--;
+  }
+  while (num > 0);
+  return stringAccumlator;
 }
 
-console.log('while', repeatStringNumTimesWithDoWhile('abc', 3));
+console.log('Do while', repeatStringNumTimesWithDoWhile('abc', 3));
```
rohni: Yup! 👍

Week3/homework/4-step3.js
```diff
@@ -1,2 +1,11 @@
...
+
+//constructor, Dog, with properties name, color, and numLegs that are set to a string, a string, and a number, respectively.
+function Dog() {
+    this.name = 'Buki';
+    this.color = 'Blue';
+    this.numLegs = 4;
+
+}
+//create a new instance of Dog, assigning it to a variable hound
+let hound = new Dog();
```
rohni: Good, this follows the brief.
However, what if I want another dog, this one called Stella, who's white and with 4 legs?
As a generic definition of a Dog, this falls down. Here is another way to approach it:

function Dog(name, colour, numLegs) {
  this.name = name;
  this.colour = colour;
  this.numLegs = numLegs;
}

const stella = new Dog('Stella', 'white', 4);
const holly = new Dog('Holly', 'caramel', 4);
const tucker = new Dog('Tucker', 'Cinnamon and white', 4);

console.log(stella, holly, tucker);
Then we can reuse Dog for each example of a dog.

Week3/homework/5-step3.js
```diff
@@ -1,2 +1,24 @@
...
+console.log(multiplyAll([
+  [1, 2],
+  [3, 4],
+  [5, 6, 7]
+]));
+console.log(multiplyAll([
+  [1, 2],
+  [3, 4],
+  [1, 1, 1]
+]));
```
rohni: Nice job!

Week3/homework/6-step3.js
```diff
@@ -1,6 +1,67 @@
...
+            for (let z in array3D[i][j]) {
+                new3DArray[i][j][z] = array3D[i][j][z];
+
+            }
+        }
+    }
+    return new3DArray;
+}
+
+console.log(matrix3D(arr3d))
```
rohni: Nice job, handling these two array dimensions.
And how would you handle a 4-dimensional array? This won't be able to handle it.
And a 5, 6, 7, or 8 dimensional array?
The best answer to handle arrays with an unknown number of dimensions is to use something called recursion.

Week3/homework/7-step3.js
```diff
@@ -22,4 +26,6 @@ f2(y);
 console.log(y);
 
 // Add your explanation as a comment here
-
+//  In the case  console.log(x) is just simply asking to console log the value of  x which is 9  because the variable (global) x is assigned to number 9. 
+// In the case console.log(y); the variable y(global) is an abject with property x and its value number 9. There fore the result is   { x: 9 }. It is not related with the function calling. 
+// If the  console log (f2(y)):  the result will be different.
```
rohni: I don't think you got the point of this exercise.
`x` is an atomic value i.e. `9`.
When you pass x into the `f1` function, it is passed as a value i.e. not a reference.  You then make changes to `val` within `f1` but this has no effect on `x` because `val` was a copy of `x`.  Printing `x` gives the original value of `9`.  If you had printed `val` it would be `10`.
`y` on the other hand is an object.and it is passed to `f2` as a reference.  So any changes made to items within the object, are reflected in `y`.  So, `f2` make changes to `val.x`, which is actually changing `y.x` as well.
When you print `y` after `f2` has run, you see the changes made in `f2` reflected in `y`.  And that is why `y.x` now shows `10`.

Week3/homework/step4-bonus.js
```diff
@@ -1,10 +1,17 @@
...
+
 
-console.log(uniqueValues);
+let uniqueValues = function (array) {
+    let uniquearray = []
+    for (let i = 0; i < array.length; i++) {
+        if (uniquearray.indexOf(array[i]) === -1 && array[i] !== '') {
+            uniquearray.push(array[i]);
+        }
+    }
```
rohni: Excellent.  Well done.

rohni: Just so you know. There is a new datatype in Javascript called a Set (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set) which only stores unique values. If the order of the resulting array does not matter, then you can do this, and have your unique values:
```
const uniques = Array.from(new Set(values));
```

Week3/homework/step4.js
```diff
@@ -1,8 +1,11 @@
...
 
-addSix(10); // returns 16
-addSix(21); // returns 27
+const addSix = function (number) {
+    return number + 6;
+}
+
+
+console.log(addSix(10)); // returns 16
+console.log(addSix(21)); // returns 27
```
rohni: Yes, this fills the brief.  But I think the idea was that you created a generic solution that would allow someone to create an `addSeven` or an `addTwenty` with the same code.  Something like this:
```
const createBase = (adder) => (n) => adder + n;
```

PR #193, opened 2018-11-28
added homework week 3 step 3 and step 4

Wael-Alhomsi:


Week3/homework/1-step3.js
```diff
@@ -2,6 +2,7 @@
 
 function foo(func) {
   // What to do here? 
+  func();
```
yash-kapila: 👍

Week3/homework/2-step3.js
```diff
@@ -3,6 +3,27 @@
 function threeFive(startIndex, stopIndex, threeCallback, fiveCallback) {
   const values = [];
   // Add your code here
+  for (let i = startIndex; i <= stopIndex; i++) {
+    values.push(i);
+  }
+
+  for (let i = 0; i <= values.length; i++) {
+    if (values[i] % 3 === 0 && values[i] % 5 === 0) {
```
yash-kapila: Hi @Wael-Alhomsi - I think you can also write the `if` block as `if (values[i] % 15 === 0)`. If a number is both divisible by 3 and 5, it would mean that it would always be divisible by 15. This way you can avoid the extra `&&` condition.

yash-kapila: Otherwise, you were able to ace the task. 😉

Week3/homework/3-step3.js
```diff
@@ -3,6 +3,15 @@
...
   // add your code here
+  if (num < 0) {
+    str = "";
+  } else {
+    let repeatedString = "";
+    for (let i = 0; i < num; i++) {
+      repeatedString += str;
+    }
+    str = repeatedString;
+  }
```
yash-kapila: Hi @Wael-Alhomsi - a couple of things here. Firstly, I liked the way you wrote the `if-else` block to handle the negative scenarios but you missed a `num === 0` operation. You can maybe add it in the if block itself like `num <= 0`. 

Secondly, as a good developer practice, you should tend to avoid writing to arguments which your function receives. In this case, we should avoid setting `str` anywhere in our function code. As an alternative, you can simply return the `repeatedString`. I saw the freeCodeCamp example and they are returning the same `str`. It works just fine in this case, however, it might cause some issues had `str` been an array or an object. I would maybe suggest playing with the below code snippet. 🙂 
```
const func = arr => {
  arr.push(10);
  return arr;
}

const arr = [7,8,9];
console.log(arr);  // arr before function call
const arr1 = func(arr);
console.log(arr);  // arr after function call
```

Week3/homework/step4-bonus.js
```diff
@@ -3,8 +3,16 @@
 const values = ['a', 'b', 'c', 'd', 'a', 'e', 'f', 'c'];
 
 // Add your function here. Try and come up with a good name for this function
-
+function constructArrayWithUniqueValues(arr) {
+    const newValues = [];
+    for (const element of arr) {
```
yash-kapila: This is nice but would you like to try it with `Array.reduce`? 😉

Week3/homework/step4.js
```diff
@@ -1,8 +1,16 @@
 'use strict';
 
 // Add your code here
+function createBase(base) {
+
+    function addToBase(value) {
+        return base + value;
+    }
+    return addToBase;
+}
```
yash-kapila: 👍

Week3/homework/5-step3.js
```diff
@@ -1,2 +1,18 @@
...
+        for (let j = 0; j < arr[i].length; j++) {
+            product *= arr[i][j];
+        }
+    }
+    return product;
+    // Only change code above this line
+}
+
+// Modify values below to test your code
+console.log(multiplyAll([[1, 2], [3, 4], [5, 6, 7]]));
```
yash-kapila: Nice one. But would you like to try it with `Array.reduce`? 😉

Week3/homework/step4-bonus.js
yash-kapila: Otherwise, you can also try using `Set` in JavaScript. If you create a Set out of an Array, it is automatically going to trim out any duplicate values. However, you would have to create an Array back from the Set using the `Array.from` method. More information about `Set`s can be found here: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set

Week3/homework/4-step3.js
```diff
@@ -1,2 +1,10 @@
 'use strict';
 // paste your freeCodeCamp solutions in here
+
+function Dog() {
+    this.name = "Lili";
+    this.color = "white";
+    this.numLegs = 4;
+}
+
+const hound = new Dog();
```
yash-kapila: This would work just fine but only in case when we need just one instance of a dog. What happens if I need multiple dogs? You can achieve that scenario a little differently like this:
```
    this.name = name;
    this.color = color;
    this.numLegs = legs;
}

// new Dog("Lili", "white", 4);
// new Dog("Don", "black", 4);
```
Makes sense?

Week3/homework/step4-bonus.js
Wael-Alhomsi: const newValues = values.reduce((acc, value) => {
  if (acc.includes(value) === false) {
    acc.push(value);
  }
  return acc;
}, []);

console.log(newValues);

Week3/homework/3-step3.js
Wael-Alhomsi: Hi @yash-kapila ,
Regarding the first part of your review for Week3/homework/3-step3.js, doesn't the 'else' expression includes the num === 0 condition?
As for the second part thank you for that important information (I didn't know it before), but I was only trying to stick to the code they have already provided on freeCode.Camp.

PR #194, opened 2018-11-29
Yaacob class18 js2 wk3

SignosRk:
The commit is 29 nov called homework week3


Week3/homework/3-step3.js
```diff
@@ -2,24 +2,37 @@
 
 // use a 'for' loop
 function repeatStringNumTimesWithFor(str, num) {
-  // add your code here
-  return str;
+  
+  let answerStr = "";
+  for(let i = 0; i < num; i++) {
```
yash-kapila: Aahaan. This works just fine. But have you thought about what happens if num < 0? Your initial declaration `let answerStr = '''` handles scenario where num could be equal to 0 but not if it less than it. Would you like to give it a try?

Week3/homework/3-step3.js
```diff
@@ -2,24 +2,37 @@
...
-
 console.log('for', repeatStringNumTimesWithFor('abc', 3));
 
+
+
 // use a 'while' loop
 function repeatStringNumTimesWithWhile(str, num) {
-  // add your code here
-  return str;
+  let ansStr = "";
```
yash-kapila: Same as above. Also, I like that you are returning `ansStr` from this function which is a good example of immutability. 👍

Week3/homework/3-step3.js
```diff
@@ -2,24 +2,37 @@
...
 
 // use a 'do...while' loop
 function repeatStringNumTimesWithDoWhile(str, num) {
-  // add your code here
-  return str;
-}
-
-console.log('while', repeatStringNumTimesWithDoWhile('abc', 3));
+  let repeat = "";
+  if(num < 0) return ""; //I used this to pass the challenge on Freecodecamp
```
yash-kapila: Aahaan. I wonder why you didn't use this logic in for and while loop. Also, you can simply return `repeat` variable from the if block as it itself is an empty string.

Week3/homework/3-step3.js
SignosRk: if(num < 0) return "";  this would be the correct to do if the num is < 0... I have it already in the last loop and I forgot to add it in the first ones hahaha...

Week3/homework/4-step3.js
```diff
@@ -1,2 +1,20 @@
...
+// // second part
+
+// function Dog() {
+//     this.name = "Rupert";
+//     this.color = "brown";
+//     this.numLegs = 4;
+//   }
+//   // Add your code below this line
+  
+//   let hound = new Dog();
```
yash-kapila: Hi @SignosRk - what if I want to have 10 dogs. I would have to create 10 different functions in that case. In such a scenario, it is a good idea to provide arguments to your constructor function and then use that to create all objects. For example:

```
  this.name = name;
  this.color = color;
  this.legs = legs;
}

const dog1 = new Dog('Dante', 'black', 4);
const dog2 = new Dog('Rupert', 'brown', 4);
```

Makes sense?

Week3/homework/3-step3.js
SignosRk: if(num < 0) return "";   same as above xD... I am just a lazy sometimes xDDDDDDD will fix it soon!

Week3/homework/5-step3.js
```diff
@@ -1,2 +1,20 @@
 'use strict';
 // paste your freeCodeCamp solutions in here
+
+function multiplyAll(arr) {
+    var product = 1;
```
yash-kapila: How about using `let` here? 🙂

Week3/homework/4-step3.js
SignosRk: of course it makes sense, it would be the proper thing to do if you have more dogs! I just didn't think to add a function and pass those arguments.. Good thing man...   :+1:

Week3/homework/5-step3.js
SignosRk: crap.... silly me xD... I shall fix it

Week3/homework/3-step3.js
SignosRk: @yash-kapila  Yeah like I commented in the other loops.. I forgot to put them, this was a challenge on freecodecamp, and with the others it passed me the challenge, but with this last when I builded my do while loop it throw me the error of if num was < 0 so I came up with the empty string return and forgot to add it to the others

Week3/homework/6-step3.js
```diff
@@ -4,3 +4,12 @@ const arr2d = [[1, 2], [3, 4], [5, 6]];
...
 // add your solution here, or add a comment on how you would tackle this problem
+
+//I would tackle this problem by using a forEach() method, which execute a callback function for every element of the array, this way I could print how many nested elements there are. 
+
+function matrix(arr) {
+    arr.forEach(function(a, b) {
+      console.log(a, b);
+    });
+  }
+  matrix(arr3d);
```
yash-kapila: Hi @SignosRk - I am not sure if this is the correct way to solve this problem because forEach iterates over individual array element and applies the callback function. For example, in our case, it would call console log only twice and print `[[1, 2], [3, 4]]` & `[[5, 6], [7, 8]]` which I don't think is what we want.

This would be best handled by a pattern in programming which we call `recursion`. I would suggest that you give it a read on the internet. Briefly, recursion means calling the same function over and over again until it satisfies a condition.

SignosRk: @yash-kapila ... Yes you are right, I didn't had that concept clear... that's why I solved this in a wrong way. I will investigate a bit more and practice more about the recursions. I've had too much information to process in javascript and not much time to practice everything, I knew this sunday about those functions calling itself xD. I will practice more in this topic man.. thanks for the feedback!

Week3/homework/7-step3.js
```diff
@@ -23,3 +23,10 @@ console.log(y);
 
 // Add your explanation as a comment here
 
+/*
+
+The answer is a bit simple to be honest, when you log x, of course it will log a value which is number 9, but in the case of y is a bit different because the value is an object, this you can see by logging the typeof x and y,
+
+this happens because in Javascript, data types like numbers strings and booleans always pass by values, and objects pass by references
+
```
yash-kapila: Hi @SignosRk - your explanation is correct. However, there is one thing which I wanted to add. **JavaScript is always call by value. However, in case of objects, the value which is passed is the reference itself.** It doesn't change anything in the way you have learnt things as they behave according to your expectations but just wanted to clarify it.

Week3/homework/step4-bonus.js
```diff
@@ -1,10 +1,21 @@
...
+// // Add your function here. Try and come up with a good name for this function
+
+// console.log(values.filter(different));
+
+// ----------second way to do it with es6-------------
 
-console.log(uniqueValues);
+const values = ['a', 'b', 'c', 'd', 'a', 'e', 'f', 'c'];
+
+const newValues = [... new Set(values)]
```
yash-kapila: Nice use of the spread operator. 👍 Would you also like to try this out using `Array.reduce`?

Week3/homework/step4.js
```diff
@@ -2,7 +2,14 @@
 
 // Add your code here
 
-const addSix = createBase(6);
 
-addSix(10); // returns 16
-addSix(21); // returns 27
+function sumTen(num) {
```
yash-kapila: Aahaan. I think you slightly missed the point of the exercise. You need to first provide a base addition as a parameter to a function. That function would then return another function to which you can provide a new value as 10 or 21. For example, in this case, our function only adds 6 to a number. What if I want to add 60 instead. 😉 

Remember closures? Let me know if you still have any confusion.

Week3/homework/6-step3.js
yash-kapila: Let me know if you have any questions around it.

yash-kapila:
Also, I wonder why you had files from previous week's homework reflected here? 🤔

Week3/homework/step4.js
SignosRk: @yash-kapila  yupe, I could use more of your knowledge because.. Yeah, I Remember closures but to be honest not as much as I want to !

yash-kapila: Okay. So a simple way to remember closures is that `if an inner function remembers the variables from its outer function even when the outer function has returned is called a closure`. 

Now, in our case, we want our outer function to accept a base argument which is going to be 6 for now. And from this function, we would want to return another function which accepts a second argument and returns the sum of that and the outer function's parameter.

I would suggest checking the example of closures which we discussed in the project I had shared.

Week3/homework/step4-bonus.js
SignosRk: @yash-kapila How could I use reduce in this case???

yash-kapila: A good way to find a use-case for `Array.reduce` is that it can be used when we want to get a single value from an array and that value can be anything, a string, a number or an array itself.

So reduce accepts an accumulator which in our case is going to be our new array with no repetitions starting with an empty array `[]`. Then we iterate through each of elements in original array and then check if it exists inside our accumulator. If it does,  we don't do anything but if it doesn't, we simply push the new element to our accumulator. Something like this:

```
const uniqueArray = arr => arr.reduce((acc,elem) => {
  if (!acc.includes(elem)) {
    acc.push(elem);
  }
  return acc;
}, []);
```

I would agree though that using a Set is much simpler for this particular operation but with this way, you can perhaps improve your knowledge of `Array.reduce`. 🙂

PR #195, opened 2018-11-30
Gaffari class18 js2 wk3

Gaffari02:


Week3/homework/1-step3.js
```diff
@@ -2,7 +2,10 @@
 
 function foo(func) {
   // What to do here? 
+  //func();
+  func = func + 1
 }
+func();
```
yash-kapila: Hi @Gaffari02 - I am not sure what are you trying to do here. Let's go through the exercise again. It says `We saw that we can pass functions as arguments to other functions. Your task is to write a function that takes another function as an argument and runs it.`. Which means that `foo` should be expecting a function, possibly `bar` and execute it. `bar` here would be our callback.

Week3/homework/2-step3.js
```diff
@@ -3,6 +3,27 @@
 function threeFive(startIndex, stopIndex, threeCallback, fiveCallback) {
   const values = [];
   // Add your code here
+  for (let i = startIndex; i <= stopIndex; i++) {
+    values.push[i];
+  }
+  values.forEach(value => {
+    if (value % 3 === 0 && value % 5 === 0) {
+      threeCallback(value);
+      fiveCallback();
```
yash-kapila: A minor problem here. You missed sending `value` as a parameter to the function. Also, I think you can also write the if block as if (values[i] % 15 === 0). If a number is both divisible by 3 and 5, it would mean that it would always be divisible by 15. This way you can avoid the extra && condition. What do you think?

Week3/homework/3-step3.js
```diff
@@ -3,6 +3,13 @@
 // use a 'for' loop
 function repeatStringNumTimesWithFor(str, num) {
   // add your code here
+  if (num < 1) {
```
yash-kapila: 👍

Week3/homework/3-step3.js
```diff
@@ -3,6 +3,13 @@
...
+    return '';
+  }
+  let myStr = str;
+  for (let i = 1; i < num; i++) {
+    str += myStr;
+  }
```
yash-kapila: Umm.. I know it has not been introduced to the class in a detailed level but try remembering this. Always stride for having your functions pure. You can try a few things to make sure you adhere to it. For example, avoid using variables from outside a function's scope and also avoid manipulating a function's parameters. In this case, let's run the loop from 0 and declare `myStr` to be an empty string and then assign it further values and return `myStr` instead of `str`. Otherwise, the task is fine. 🙂

Week3/homework/3-step3.js
```diff
@@ -11,6 +18,15 @@ console.log('for', repeatStringNumTimesWithFor('abc', 3));
...
 function repeatStringNumTimesWithWhile(str, num) {
   // add your code here
+  if (num < 1) {
+    return '';
+  }
+  let myStr = str;
+  let i = 1;
+  while(i < num) {
+    str += myStr;
+    i++;
```
yash-kapila: Same observation for while and do-while loop as for loop. Please have a look.

Week3/homework/4-step3.js
```diff
@@ -1,2 +1,9 @@
 'use strict';
 // paste your freeCodeCamp solutions in here
+function Dog () {
+    this.name = 'Kara';
+    this.color = 'black';
+    this.numLegs = 4;
+}
+
+const hound = new Dog();
```
yash-kapila: Hi @Gaffari02 - what if I want to have 10 dogs. I would have to create 10 different functions in that case. In such a scenario, it is a good idea to provide arguments to your constructor function and then use that to create all objects. For example:
```
  this.name = name;
  this.color = color;
  this.legs = legs;
}

const dog1 = new Dog('Dante', 'black', 4);
const dog2 = new Dog('Rupert', 'brown', 4);
```

Makes sense?

Week3/homework/5-step3.js
```diff
@@ -1,2 +1,19 @@
 'use strict';
 // paste your freeCodeCamp solutions in here
+function multiplyAll(arr) {
+    var product = 1;
```
yash-kapila: I think you mean `let` here 😉

Week3/homework/6-step3.js
```diff
@@ -4,3 +4,11 @@ const arr2d = [[1, 2], [3, 4], [5, 6]];
 const arr3d = [[[1, 2], [3, 4]], [[5, 6], [7, 8]]];
 
 // add your solution here, or add a comment on how you would tackle this problem
+
+for ( var i = 0; i < arr.length; i++) {
+    for ( var j = 0; j < arr[i].length; j++) {
+        for ( var k = 0; k < arr[j].length; k++) {
+            product *= arr[i][j][k];
+        }
```
yash-kapila: Aahaan. I think this works just fine. However, what if I had a 4d array or a 5d array. Our code would break under such scenarios. Such scenarios would be best handled by a pattern in programming which we call **recursion**. I would suggest that you give it a read on the internet. Briefly, recursion means calling the same function over and over again until it satisfies a condition. 🙂

Week3/homework/7-step3.js
```diff
@@ -23,3 +23,6 @@ console.log(y);
 
 // Add your explanation as a comment here
 
+// in the first example because it is a primitive data type, it passes by value,
+// but in the second example it is an object and, objects pass by reference.
```
yash-kapila: 👍 
A quick point which I wanted to otherwise a correct answer though - **JavaScript is always call by value. However, in case of objects, the value which is passed is the reference itself.** It doesn't change anything in the way you have learnt things as they behave according to your expectations but just wanted to clarify it.

PR #199, opened 2018-12-02
My homework Week3

Kani103:
i did some steps, the other steps still working on it. sorry!!

### DIFF
--- a/Week3/homework/1-step3.js
+++ b/Week3/homework/1-step3.js
@@ -2,6 +2,7 @@
 
 function foo(func) {
   // What to do here? 
+  return func 
 }
 
 function bar() {
--- a/Week3/homework/3-step3.js
+++ b/Week3/homework/3-step3.js
@@ -3,7 +3,11 @@
 // use a 'for' loop
 function repeatStringNumTimesWithFor(str, num) {
   // add your code here
-  return str;
+  let string = "";
+  for (let i = 0; i < num; i++) {
+    string += str;
+  }
+  return string;
 }
 
 console.log('for', repeatStringNumTimesWithFor('abc', 3));
@@ -11,7 +15,11 @@ console.log('for', repeatStringNumTimesWithFor('abc', 3));
 // use a 'while' loop
 function repeatStringNumTimesWithWhile(str, num) {
   // add your code here
-  return str;
+  while (num > 0) {
+    string += str;
+    num--;
+  }
+  return string;
 }
 
 console.log('while', repeatStringNumTimesWithWhile('abc', 3));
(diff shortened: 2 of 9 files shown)


Week3/homework/1-step3.js
```diff
@@ -2,6 +2,7 @@
 
 function foo(func) {
   // What to do here? 
+  return func 
 }
```
rohni: Not quite.  All you will do here is return the function object.  You want to call `func` and return the value from calling `func`.
```
....
return func();
....
````

File: Week3/homework/3-step3.js [hunk 2]
```diff
@@ -3,23 +3,38 @@
 // use a 'for' loop
```
rohni: Where did 2-step3.js go to?

Week3/homework/3-step3.js
```diff
@@ -3,23 +3,38 @@
...
+  return string;
 }
 
 console.log('for', repeatStringNumTimesWithFor('abc', 3));
 
 // use a 'while' loop
 function repeatStringNumTimesWithWhile(str, num) {
   // add your code here
-  return str;
+  while (num > 0) {
```
rohni: You have to define string before you can use it.  
This is where your program errors out because it cannot find string.
It is very obvious that you have not *run* or *tested* your program, otherwise you would have seen the `error message`, and been able to correct the problem which is:
```
let string = '';
```

Week3/homework/4-step3.js
```diff
@@ -1,2 +1,7 @@
 'use strict';
 // paste your freeCodeCamp solutions in here
+function Dog () {
+    this.name = "Jumbo";
+    this.color = "Black";
+    this.numLegs = 4;
+    }
```
rohni: Correct. ✅ 
Just wondering why you removed the line that actually creates an instance of a `Dog`.

Week3/homework/5-step3.js
```diff
@@ -1,2 +1,10 @@
 'use strict';
 // paste your freeCodeCamp solutions in here
+function Dog() {
+    this.name = "Rupert";
+    this.color = "brown";
+    this.numLegs = 4;
+  }
+  // Add your code below this line
```
rohni: This question was about nested loops, no the `Dog`.

Week3/homework/step4-bonus.js
```diff
@@ -3,8 +3,14 @@
 const values = ['a', 'b', 'c', 'd', 'a', 'e', 'f', 'c'];
```
rohni: I seem to be missing, `6-step3.js`, `7-step3.js`, and `step4.js`.

Week3/homework/step4-bonus.js
```diff
@@ -3,8 +3,14 @@
...
 
 // Add your function here. Try and come up with a good name for this function
-
+function removeDoubles(foo) {
+    const uniqueLetters = [];
+    for (let i = 0; i < foo.length; i++){
+        uniqueLetters.push(foo[i]);
```
rohni: All you are doing here is copying the whole array. *Not* removing the duplicate values.
You need to check first whether the value already exists in `uniqueLetters` and if it does just skip to the next iteration of the loop, and if it does not, add it to `uniqueLetters`.

PR #200, opened 2018-12-02
Keyya class18 js2 week3

keyyaali:


Week3/homework/1-step3.js
```diff
@@ -1,11 +1,9 @@
 'use strict';
 
-function foo(func) {
-  // What to do here? 
-}
+(function foo(func) {
+  console.log("Hello, Foo");
```
daniela-olaru: Here you are supposed to call the function given as parameter, namely func. When foo bar is called ( foo(bar); in the initial file), the func parameter given to the funct function will become bar, such that when you call the func parameter, you actually call the bar function, which will to the console itself.

Week3/homework/3-step3.js
```diff
@@ -3,23 +3,53 @@
...
+  while (str < 3) {
+    str++;
+  }
+
+
+  // expected output: 3
+
 }
 
 console.log('while', repeatStringNumTimesWithWhile('abc', 3));
```
daniela-olaru: When run, it gives undefined :)

Week3/homework/3-step3.js
```diff
@@ -3,23 +3,53 @@
...
+
+  do {
+    i = i + 1;
+    str = str + i;
+  } while (i < 3);
+
+
 }
 
 console.log('while', repeatStringNumTimesWithDoWhile('abc', 3));
```
daniela-olaru: Same here, when run, it gives undefined. :)

Week3/homework/3-step3.js
```diff
@@ -3,23 +3,53 @@
...
+  }
+  return strOut;
 }
 
 console.log('for', repeatStringNumTimesWithFor('abc', 3));
 
 // use a 'while' loop
 function repeatStringNumTimesWithWhile(str, num) {
   // add your code here
-  return str;
```
daniela-olaru: The idea is to thing about it in the same way you did it for the _for_ loop.
First, you initialise the variable to be returned and the index used to loop.

_var strOut = "";
var i = 0;_

Then, while index is smaller then the maximum number given as param, namely _num_, you add the _str_ given as param to the current strOut.

 _while (i < num) {
    strOut = strOut + str;
    i++;
  }_

 and then return the strOut (which now will have the given string _str_ repeated _num_ times)
 
_return strOut;_

Week3/homework/3-step3.js
```diff
@@ -3,23 +3,53 @@
...
+  var str = "";
+  var i = 0;
+
+  do {
+    i = i + 1;
+    str = str + i;
+  } while (i < 3);
+
```
daniela-olaru: First, you didn't return anything, you should return, as in the "for" example, which you did correctly, the new strOut variable created using the loop.
Secondly, "str = str + i;" this makes no sense, for several reasons: first, you don't want to add the index itself to the string. Secondly, you don't want to alter the parameter _str_ given. :) You just want to create a new string, as you did correctly in the "for" example, and add the given string _str_ several times. _i_ should be used just as a way to loop, never used in creating the string itself.
In conclusion, the code should be really similar to the previous one, just adding the condition to stop at the end:
Initialize _i_ and _strOut_.
Loop:
_do {
    i = i + 1;
    strOut = strOut + str;
  } while (i < num);_

And, of course, return the _strOut_.

Week3/homework/3-step3.js
daniela-olaru: Generally, I notice you didn't test your code. This is something that might have a really bad impact on the team, when you will be working in a real team environment. Please make sure you test your code always before handing it. In most of the cases, no matter how experienced you are, the code will not run as expected. It is our responsibility to ALWAYS test the code, no matter how confident we are.

Week3/homework/4-step3.js
```diff
@@ -1,2 +1,19 @@
...
+
+// Use a Constructor to Create Objects
+function Dog() {
+  this.name = "Rupert";
+  this.color = "brown";
+  this.numLegs = 4;
+}
+var hound = new Dog();
+// Add your code below this line
+
```
daniela-olaru: Seems all good!

Week3/homework/5-step3.js
```diff
@@ -1,2 +1,17 @@
...
+    for (var j = 0; j < arr[i].length; j++) {
+      product = product * arr[i][j];
+    }
+  }
+  // Only change code above this line
+  return product;
+}
+
+// Modify values below to test your code
+multiplyAll([[1, 2], [3, 4], [5, 6, 7]]);
```
daniela-olaru: Well done!

Week3/homework/6-step3.js
```diff
@@ -1,6 +1,13 @@
...
+function printThreeDimensionalArray(arr) {
+  for (let i = 0; i < arr.length; i++) {
+    for (let i2 = 0; i2 < arr[i].length; i2++) {
+      for (let i3 = 0; i3 < arr[i][i2].length; i3++) {
+        console.log(arr[i][i2][i3]);
+      }
+    }
+  }
+}
+printThreeDimensionalArray(arr3d);
```
daniela-olaru: Well done, also!

Week3/homework/7-step3.js
```diff
@@ -22,4 +22,8 @@ f2(y);
 console.log(y);
 
 // Add your explanation as a comment here
-
+/* 
+In Javascript all types of data except objects define immutable values "primitive values".
+In the case of objects, JavaScript always uses the Object by reference when it passes as an argument, so if we pass objects to a function and call the Function, the function updates the value of the object.
```
daniela-olaru: Yes, exactly!

Week3/homework/step4.js
```diff
@@ -1,8 +1,14 @@
...
+  return addToBase;
 
-const addSix = createBase(6);
+  const addSix = createBase(6);
 
-addSix(10); // returns 16
-addSix(21); // returns 27
+  addSix(10); // returns 16
+  addSix(21); // returns 27
+}
```
daniela-olaru: Again, it is not tested. :) The idea is really good, though! Just that you should have closed the createBase function after _return addToBase;_ 
If you really understood the idea, it is great! But please keep in mind to test, closing the function body in a strange way, after calling those function might give the impression that you didn't really understand the logic behind, which I hope it is not the case. :)

Week3/homework/step4-bonus.js
```diff
@@ -4,7 +4,15 @@ const values = ['a', 'b', 'c', 'd', 'a', 'e', 'f', 'c'];
...
+      newValues.push(element);
+    }
+  }
+  return newValues;
+}
 // Replace `yourFunction` with the name of the function you just created
-const uniqueValues = yourFunction(values);
-
+const uniqueValues = constructArrayWithUniqueValues(values);
 console.log(uniqueValues);
```
daniela-olaru: Really well done the code! Especially I appreciate you payed attention to naming the function and the variables also! 👍

Week2/homework/maartjes_work.js
```diff
@@ -1,24 +1,22 @@
-'use strict';
-
-const monday = [
```
daniela-olaru: It seems you have deleted ```const moday = [``` and replaced it by ```const tasks = monday.concat(tuesday);``` by mistake, i'm sure, which generates a syntax error in the file. If you add this back it works!

Week2/homework/maartjes_work.js
```diff
@@ -41,7 +39,10 @@ const tuesday = [
     duration: 40
   }
 ];
-
 const tasks = monday.concat(tuesday);
-
-// Add your code here
+let twoDaysIncome = tasks
```
daniela-olaru: Nicely done! 👏

Week2/homework/maartjes_work.js
```diff
@@ -41,7 +39,10 @@ const tuesday = [
...
+  .map(task => task.duration / 60)
+  .filter(task => task >= 2)
+  .map(total => total * 60)
+  .reduce((a, b) => a + b, 0);
+console.log("Maartje earned " + twoDaysIncome + ".00 Euro on Monday and Tuesday!");
```
daniela-olaru: For this last part, you were supposed to round the amount to 2 decimals. Let's imagine that Maartje makes $59/hr instead of 60. You cannot be sure that it will always be an exact amount. Javascript has a functionality just for this use-case: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/toFixed 

Sample:
```javascript
console.log("Maartje earned " + twoDaysIncome.toFixed(2) + " Euro on Monday and Tuesday!"); 
```

Week2/homework/map_filter.js
```diff
@@ -1,5 +1,65 @@
 'use strict';
```
daniela-olaru: Well done on this assignment!

PR #209, opened 2019-01-10
Homework week1

YaseirAlkhwalda:


YaseirAlkhwalda:
Homework Week 1

YaseirAlkhwalda:
update my app.js file

Week1/homework/app.js
```diff
@@ -1,11 +1,182 @@
...
+
+  */
+
+  const information = {
+    home_and_world: {
+      author: 'Rabindranath Tagore',
+      name: 'The Home and the World',
+      alt: 'This is the cover of the book called The Home and the World',
+      description:
+        "Originally written in Bengali and published in 1916, The Home and the World is the vision of Rabindranath Tagore, Asia's first Nobel laureate."
```
yash-kapila: Hi @YaseirAlkhwalda - As an alternative to double quotes here, you might learn about a new feature introduced in ES6 which is called template strings using back ticks i.e (`). There are also a few other benefits of using back ticks like having variables inside strings and multiline strings which you can find out here - https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals

Week1/homework/app.js
```diff
@@ -1,11 +1,182 @@
...
+    },
+    the_rescue: {
+      author: 'Joseph Conrad',
+      name: 'The Rescue',
```
yash-kapila: Something which Jim pointed out in the Slack channel. The requirements states: 

`1.4 Make an object (not an array!) containing information for each book. Each property of this object should be another (i.e., nested) object with the book ID you thought up in step 1.1 as a key, and at least the following properties: title, language and author.`

However, the properties which you are using are different except for `author`. It might seem very minor right now as you are developing alone but imagine a situation where you work with a Backend developer. You agree on a contract related to naming conventions but now that you use different names for properties, you are in a breach of that contract which would eventually lead to bugs in your application. Please make sure that following the requirements is priority number 1. 🙂

YaseirAlkhwalda: yes sure. I will focus on this thing.
thank you very much ^_^

Week1/homework/app.js
```diff
@@ -1,11 +1,182 @@
...
+    robinson_crusoe: './images/robinson_crusoe.jpeg',
+    the_odyssey: './images/the_odyssey.jpeg',
+    death_in_venice: './images/death_in_venice.jpeg',
+    adventures_of_huckleberry_finn: './images/adventures_of_huckleberry_finn.jpeg',
+    oliver_twist: './images/oliver_twist.jpeg',
+    frankenstein: './images/frankenstein.jpeg',
+    return_of_native: './images/return_of_native.jpeg'
+  };
+
+  const div = document.getElementById('content');
```
yash-kapila: Hi @YaseirAlkhwalda - I think you missed something listed in the requirements. This may not be functionally important but I believe such things would help you becoming a better developer. For example, **1.5** and **1.8**  state that you modify/create functions to achieve the desired functionality.

The current piece of code written works just fine but it is a good practice to separate out your code into smaller functions. Functions provide a lot of benefits like re-usable code and more semantic understanding of the code. 

For example, you can have a function(say `createAppLayout`) to create the layout of your application with the `content`, `h1`, `div` and `ul` tags. Similarly, you can create a function(say `createBookItems`) for **1.5** where you add all the books from your object as li elements to the `ul` tag. And at the end, have another function(say `addBookImages`) for **1.8** where you use your `images` object to add images to all books that you created in the previous function.

I would suggest that you give those two requirements another go to split out your code into smaller functions. 🙂

Week1/homework/app.js
```diff
@@ -1,11 +1,182 @@
...
+    liTag.className = 'book';
+
+    const div2 = document.createElement('div');
+    div2.className = 'section';
+
+    const writer = document.createElement('h3');
+    writer.innerHTML = information[id].author;
+    writer.className = 'writer';
+
+    const title = document.createElement('h2');
```
yash-kapila: I see 9 places in total where you do:

```
const var = document.createElement('<tag>');
var.className = '.....';
```

That is in total 18 lines of code where we are repeating ourselves. There is a principle in software development named Don't Repeat Yourself (DRY). That means, if you see repeating yourself like in our case, we should take out the common pieces and move those to a separate function.

Thus, you can have a function say `setElement` which accepts some arguments and performs the same operation as many times as you want it to. Would you like to try it that way? You will find your code to be much cleaner and easier to read. 🙂

YaseirAlkhwalda: I will try to make my home work look better :") 
thank you very much for review my homework .

YaseirAlkhwalda: Hi
excuse me, I want to update my homework. But it give All checks have failed.
I don't know the problem exactly.
thanks

PR #210, opened 2019-01-10
Homework week1

ahmetbibi:


Week1/homework/app.js
```diff
@@ -1,11 +1,143 @@
...
+      liTag2.appendChild(headings);
+      liTag2.appendChild(aboutLanguage);
+      liTag2.appendChild(aboutAuthor);
+      ulTag2.appendChild(liTag2);
+
+      const idOfLi = document.getElementsByTagName('li')[i].getAttribute('id');
+      const keysOfBookCovers = Object.keys(bookCovers);
+      const valuesOfBookCovers = Object.values(bookCovers);
+
+      for (let j = 0; j < keysOfBookCovers.length; j++) {
```
remarcmij: See my slack message mentioned earlier how to do add images in a separate function.

Week1/homework/app.js
```diff
@@ -103,60 +103,45 @@
...
+      addNewTag('h2', liTag2, { text: book.title });
+      addLabel('Language', book.language, liTag2);
+      addLabel('Author', book.author, liTag2);
     }
   }
 
   function addImage() {
     const keysOfBookCovers = Object.keys(bookCovers);
-
     for (const keys of keysOfBookCovers) {
```
remarcmij: The `bookIds` array is what you need here, as you have done in line 129:

```js
for (const bookId of bookIds) {
```

In this way the `bookIds` array totally 'controls' which books are to be shown in the page and in which order. If you make changes to this array (remove items, reorder items) all else will follow along.

Week1/homework/app.js
```diff
@@ -103,60 +103,45 @@
...
       const book = bookCovers[keys];
```
remarcmij: An element of a `bookCovers` array is a `bookCover`, not a `book`.

```js
const bookCover = bookCovers[bookId];
```

Week1/homework/app.js
ahmetbibi: Now I really got the logic of ordering the images ( before, here is a little dark in my mind but now it becomes clear)

PR #212, opened 2019-01-10
JS2_Week1_HomeWork.

BassamHajar:
Great thanks for your efforts.


Week1/homework/app.js
```diff
@@ -2,10 +2,112 @@
...
+
+  // 1.8
+  const keysBooksCoversObj = Object.keys(booksCoversObj);
+  const valuesCovers = Object.values(booksCoversObj);
+  function displayBooksCovers() {
+    keysBooksCoversObj.forEach(book => {
+      const li = document.createElement('li');
+      li.textContent = book;
+      document.body.appendChild(li);
+      valuesCovers.forEach(cover => {
```
yash-kapila: The requirement asks you to use the list element created in the previous steps. You need not create new list elements here again. Please go through the requirement 1.8 again.

Also, you need not use Object.values here and the logic to add image can simply be like:
```
const img = document.createElement('img');
img.setAttribute('src', booksCoversObj[book]);
.
.
.
```
i.e you use the key of the property(defined as `book`) to fetch the value by looking inside the actual object itself(https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Property_accessors) 🙂

Week1/homework/app.js
```diff
@@ -2,10 +2,112 @@
...
+    the_craftsman: './imgs/the_craftsman.jpg'
+  };
+
+  // 1.8
+  const keysBooksCoversObj = Object.keys(booksCoversObj);
+  const valuesCovers = Object.values(booksCoversObj);
+  function displayBooksCovers() {
+    keysBooksCoversObj.forEach(book => {
+      const li = document.createElement('li');
+      li.textContent = book;
```
yash-kapila: This adds the `key` i.e `the_brief_history_of_the_dead`, `the_craftsman` etc as the content of the `li` elements. Instead, what we want is to have the `title` defined in the above object to be displayed. 🙂

Week1/homework/app.js
```diff
@@ -1,11 +1,132 @@
...
+  function createList() {
+    const ul = document.createElement('ul');
+    bookTitles.forEach(title => {
+      const li = document.createElement('li');
+      li.textContent = title;
+      ul.appendChild(li);
+    });
+    document.body.appendChild(ul);
+  }
+  window.onload = createList;
```
remarcmij: You can assign only one event handler function to `window.onload`. Since you assign it in three different places in your code, the last one (in line 131) wins. Consequently your  `createList` and `displayBooksInfo()` functions are never called.

PR #213, opened 2019-01-10
JavaScript2/Week1 homework

Ayh0o0m:


Ayh0o0m:
Homework Week 1

Week1/homework/app.js
```diff
@@ -1,11 +1,149 @@
...
+
+//  1.5------1.8
+function getBooksInfo() {
+  const ulDom = document.createElement('ul');
+  document.body.appendChild(ulDom);
+  ulDom.setAttribute('class', 'books');
+  for (let i = 0; i < bookTitles.length; i++) {
+    // const divDom1 = document.createElement('div');
+    const liDom = document.createElement('li');
+    const divDom2 = document.createElement('div');
```
remarcmij: What does the word `Dom` add to the meaning of this variable? Or is it just noise?

Week1/homework/app.js
```diff
@@ -1,11 +1,149 @@
...
+    divDom3.setAttribute('class', 'book_cover');
+    liDom.setAttribute('id', bookTitles[i] + '_container');
+    liDom.setAttribute('class', 'book');
+    h1Dom.setAttribute('class', 'books_titles');
+    h4Dom.setAttribute('class', 'book_author');
+    h4Dom.setAttribute('id', bookTitles[i] + '_author');
+    h5Dom.setAttribute('class', 'book_language');
+    h5Dom.setAttribute('id', bookTitles[i] + '_language');
+    img.setAttribute('src', 'images/' + bookTitles[i] + '.jpg');
+    img.setAttribute('alt', bookListObj[`${bookTitles[i]}`] + ' cover');
```
remarcmij: `bookListObj[`${bookTitles[i]}`]` is a JavaScript object, not a string. If you look at the generated HTML for this image you will see the string `[object Object]` which is not what your want:

```html
<img src="images/harry_potter_chamber_secrets.jpg" alt="[object Object] cover" id="harry_potter_chamber_secrets">
```

Week1/homework/app.js
```diff
@@ -1,11 +1,149 @@
...
+    img.setAttribute('id', bookTitles[i]);
+
+    h1Dom.textContent = bookListObj[`${bookTitles[i]}`].title;
```
remarcmij: There is not need to use a template literal here -> `bookListObj[bookTitles[i][` will do fine. Or, you can just use the convenience variable `bookDetails` mentioned before instead:

```js
h1Dom.textContent = bookDetails.title;
```

Week1/homework/app.js
```diff
@@ -1,11 +1,149 @@
...
+  },
+  another_bullshit_night_in_suck_city: {
+    title: 'Another Bullshit Night in Suck City',
+    language: 'English',
+    author: 'Nick Flynn'
+  }
+};
+
+//  1.5------1.8
+function getBooksInfo() {
```
remarcmij: When a function name starts with the word `get` most developers expect such a function to returns something, the `thing` your are getting. In this case your function doesn't return anything. So the verb `get` is misleading. A better name would for instance be `renderBookDetails()` or something like that.

Week1/homework/app.js
```diff
@@ -1,11 +1,149 @@
...
+  const ulDom = document.createElement('ul');
+  document.body.appendChild(ulDom);
+  ulDom.setAttribute('class', 'books');
+  for (let i = 0; i < bookTitles.length; i++) {
+    // const divDom1 = document.createElement('div');
```
remarcmij: 1. Remove commented-out code when submitting a pull request. The reviewer is not interested in 'dead code'.
2. You can make your code a bit easier to read if you introduce two 'convenience' variables at the top of your loop body:

```js
const bookTitle = bookTitles[i];
const bookDetails = bookListObj[bookTitle];
```

Or, better still, use a `for...of` loop:

```js
for (const bookTitle of bookTitles) {
  const bookDetails = bookListObj[bookTitle];
  // etc
}
```


Now, whenever you use `bookTitles[i]` in your loop body, replace it by `bookTitle`. And where you use `bookListObj[bookTitle]` or even `bookListObj[bookTitles[i]], ` just use `bookDetails`. Less to type, smaller chance to make typing errors.

Week1/homework/app.js
```diff
@@ -1,11 +1,149 @@
...
+function changeBgOriginal() {
+  for (let i = 0; i < bookTitles.length; i++) {
+    const img = document.getElementById(bookTitles[i]);
+    img.addEventListener('mouseout', function() {
+      img.setAttribute('src', 'images/' + bookTitles[i] + '.jpg');
+      img.setAttribute('alt', bookListObj[`${bookTitles[i]}`] + ' Cover');
+    });
+  }
 }
+changeBgOriginal();
```
remarcmij: Nice use of event listeners to handle the `mousover` and `mouseout` events. But what if you don't have a mouse, e.g. on a mobile phone or iPad?

The two functions `changeBg` and `changeBgOriginal` both loop through the same `bookTitles` array. You could combine them in one function. Also, it is best to execute all code when the page is fully loaded, using `window.onload`, for example:


```js
// ...

function addEventListeners() {
  for (const bookTitle of bookTitles) {
    const img = document.getElementById(bookTitle);
    img.addEventListener('mouseover', function() {
      img.setAttribute('src', 'images/' + bookTitle + '_author.jpg');
      img.setAttribute('alt', bookTitle + ' author');
    });
    img.addEventListener('mouseout', function() {
      img.setAttribute('src', 'images/' + bookTitle + '.jpg');
      img.setAttribute('alt', bookTitle + ' Cover');
    });
  }
}

function main() {
  getBooksInfo();
  addEventListeners();
}

window.onload = main;
```

Week1/homework/app.js
```diff
@@ -1,11 +1,149 @@
...
+    divDom2.setAttribute('class', 'book_info');
+    divDom3.setAttribute('class', 'book_cover');
+    liDom.setAttribute('id', bookTitles[i] + '_container');
+    liDom.setAttribute('class', 'book');
+    h1Dom.setAttribute('class', 'books_titles');
+    h4Dom.setAttribute('class', 'book_author');
+    h4Dom.setAttribute('id', bookTitles[i] + '_author');
+    h5Dom.setAttribute('class', 'book_language');
+    h5Dom.setAttribute('id', bookTitles[i] + '_language');
+    img.setAttribute('src', 'images/' + bookTitles[i] + '.jpg');
```
remarcmij: You have chosen to use the book title as the file name for the image. Although the assignment asked you to create a separate object with the image urls, your approach will work, but only when you can choose those file names yourself. If you use images hosted somewhere else, you don't control their file names or urls and you will need some translation between the ids you use and the image urls.

Week1/homework/app.js
```diff
@@ -1,11 +1,149 @@
...
+  'pride_and_prejudice',
+  'to_sir_with_love',
+  'to_miss_with_love',
+  'where_the_wild_things_are',
+  'me_talk_pretty_one_day',
+  'another_bullshit_night_in_suck_city'
+];
+
+//  1.2
+/*  function createBookList() {
```
remarcmij: See my message in slack why this function should be present in your code (and not commented out).

Week1/homework/app.js
Ayh0o0m: it supposed to be indicating that this variable creates element called <div> in the DOM. that's what i was thinking about

Week1/homework/app.js
Ayh0o0m: oh i thought this will let me print the value in bookTitles. thans for clarification.

Week1/homework/app.js
Ayh0o0m: i actully still don't feel comfortable using for ...of loops. but i will give it a try for sure

PR #215, opened 2019-01-10
Utku / Class 19 / Homework Week 1

u61:


u61:
Good Evening Sander,
I editted my js file as you reviewed.
I am ready for second check.
Thanks in advance.

PR #218, opened 2019-01-11
2nd update: Homework-Week1 / Fatih

kfth:
This is the first homework of JavaScript2


Week1/homework/app.js
```diff
@@ -3,9 +3,162 @@
...
+    'war_and_peace',
+    'anna_karenina',
+    'animal_farm',
+    'a_tale_of_two_cities',
+    'the_alchemist'
   ];
 
-  // Replace with your own code
-  console.log(bookTitles);
+  const Books = {
```
yash-kapila: In JavaScript, the variable naming convention is to make use of camelCasing as listed [here](https://github.com/HackYourFuture/fundamentals/blob/master/fundamentals/naming_conventions.md). Quite similar to how you have named the `bookTitles` variable above.

Week1/homework/app.js
```diff
@@ -3,9 +3,162 @@
...
+    moby_dick: {
+      title: 'Moby-Dick',
+      author: 'Herman Melville',
```
yash-kapila: I would like to appreciate that you followed the instructions while naming properties for your `books` object and not deviating from it.

Would like to add one small thing here which you might learn a little later. ES6 has introduced a new feature called [template literals](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals). This makes use of back-ticks and a couple of advantages of using them is that you can also include variables inside your strings and split really long lines as we have in `info` property in multi-line strings which are then easier to read. 🙂

Week1/homework/app.js
yash-kapila: In JavaScript, the variable naming convention is to make use of camelCasing as listed [here](https://github.com/HackYourFuture/fundamentals/blob/master/fundamentals/naming_conventions.md). Quite similar to how you have named the `bookTitles` variable above.

Week1/homework/app.js
```diff
@@ -3,9 +3,162 @@
...
+      language: 'English',
+      length: 720,
+      info:
+        'In Herman Melvilles Moby- Dick, Captain Ahab is an eerily compelling madman who focuses his distilled hatred and suffering(and that of generations before him) into the pursuit of a creature as vast, dangerous and unknowable as the sea itself.More than just a novel of adventure, this is a haunting social commentary populated with some of the most enduring characters in literature.Written with wonderfully redemptive humour, Moby Dick is a profound and timeless inquiry into character, faith and the nature of perception.'
```
yash-kapila: I would like to appreciate that you followed the instructions regarding naming conventions of the properties of the books object. I would like to add one thing here though.

ES6 introduced a new way of dealing with strings called [template strings](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals). It makes use of back-ticks(`) and a couple of advantages of using them over plain strings are:
1. You can make use of variable names inside the strings which you can't do if you are using single or double quotes.
2. You can split really long lines as in the case of our `info` property into multi-line strings which are then easier to read.

I would suggest you give it a read. 🙂

Week1/homework/app.js
```diff
@@ -3,9 +3,162 @@
...
+    wuthering_heights: 'img/wuthering_heights.jpg',
+    war_and_peace: 'img/war_and_peace.jpg',
+    anna_karenina: 'img/anna_karenina.jpg',
+    animal_farm: 'img/animal_farm.jpg',
+    a_tale_of_two_cities: 'img/a_tale_of_two_cities.jpg',
+    the_alchemist: 'img/the_alchemist.jpg'
+  };
+
+  function generateBookCards(container) {
+    for (let i = 0; i < bookTitles.length + 1; i++) {
```
yash-kapila: You need not loop till `array length + 1` as the last element created would have its `id = undefined` i.e bookTitles is defined only from indices 0 to 9. `bookTitles[10] = undefined` 🙂

yash-kapila: Alternatively, you can make use of the `for...of` loop as you are using in the below functions.

Week1/homework/app.js
```diff
@@ -3,9 +3,162 @@
...
+    for (const key of bookList) {
+      const img = document.createElement('img');
+      img.src = BookCovers[key];
+      img.alt = key;
+      const li = document.getElementById(key);
+      li.appendChild(img);
+    }
+  }
+
+  function getActualBookInfo() {
```
yash-kapila: Usually, if one sees a function named like `get....`, one expects the function to return something at the end. However, in this function, we aren't returning anything. Instead, we are inserting all the book related information to our list elements. A more appropriate name for this function could have been `insertBookDetails` or something similar to how you named the above function `insertBookCovers`. Not a big thing but it always helps in understanding your code. 🙂

Week1/homework/app.js
kfth: Hi @yash-kapila,
Thank you so much for your review.  For the naming part, I mixed up Classes and objects' naming, I grasped the issue, thanks

Week1/homework/app.js
kfth: I've searched template strings and learn to use multine line  and many others,  thanks for the suggestion.

Week1/homework/app.js
kfth: I realize that that issue after your review and fixed it

Week1/homework/app.js
kfth: I've fixed all the issues upon your review.

I also implemented another function for repeated "create and append  element tasks" in the code as covered in the  last lesson. Should I push and make a new PR for the updated version of homework week1?

Week1/homework/app_new.js
```diff
@@ -0,0 +1,159 @@
...
+      img.src = BookCovers[key];
+      img.alt = key;
+    }
+  }
+
+  function insertBookDetails() {
+    const bookList = Object.keys(books);
+    for (const key of bookList) {
+      const bookCard = document.getElementById(key);
+      createAndAppendElement('h2', bookCard, books[key].title);
```
remarcmij: This would become:

```js
createAndAppendElement('h2', bookCard, {text: books[key].title });
```

Week1/homework/app_new.js
```diff
@@ -0,0 +1,159 @@
...
+    the_alchemist: {
+      title: 'The Alchemist',
+      author: 'Paulo Coelho',
+      language: 'English',
+      length: 197,
+      info:
+        'Paulo Coelhos masterpiece tells the mystical story of Santiago, an Andalusian shepherd boy who yearns to travel in search of a worldly treasure.His quest will lead him to riches far different—and far more satisfying—than he ever imagined.Santiagos journey teaches us about the essential wisdom of listening to our hearts, of recognizing opportunity and learning to read the omens strewn along lifes path, and, most importantly, to follow our dreams.'
+    }
+  };
+  const BookCovers = {
```
remarcmij: Variable, function and parameter names should start with a lowercase letter.

Week1/homework/app_new.js
```diff
@@ -0,0 +1,159 @@
...
+    }
+  }
+
+  function insertBookCovers() {
+    const bookList = Object.keys(BookCovers);
+    for (const key of bookList) {
+      const li = document.getElementById(key);
+      const img = createAndAppendElement('img', li);
+      img.src = BookCovers[key];
+      img.alt = key;
```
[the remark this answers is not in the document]
kfth: Thank you so much @remarcmij . I wanted to implement the easy way of create and append function but i got the point with set attribute and will update it

Week1/homework/app_new.js
kfth: I was mixed-up Classes and Objects naming conventions. I realized and fixed that issue thanks to review of @yash-kapila . I've missed out "BookCovers" and I fixed it.

Week1/homework/app_new.js
```diff
@@ -0,0 +1,159 @@
...
+  };
+
+  function createAndAppendElement(tag, parent, elementInnerText) {
+    const element = document.createElement(tag);
+    if (elementInnerText) {
+      element.innerText = elementInnerText;
+    }
+    parent.appendChild(element);
+    return element;
+  }
```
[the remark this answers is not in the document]
kfth: It looks pretty good after I've updated the function 🥇 👍

PR #221, opened 2019-01-11
Soolaf-js2-Week1

soolaf:
homework Week1


waxs:
Hi Soolaf,

Good to see you finished you assignment for the week, I see that you already picked up a thing or two about the way Javascript works and it's always nice to see good progression! As always I review the work as if I've been developing it myself. So some of my comments are more a personal preference instead of a clear mistake or wrongdoing on your side. Please be aware that some of these "preferences" are considered a best practice. I highly encourage you to make the necessary changes to get a better understanding of the code! So without futher ado my comments about your assignment.

You started of declaring a curly bracket, that was unnecessary for the script to work, so I would recommend removing it from the code. You nested the information about the books such as the `listBooks` and `information` inside the `main()` function. What would happen if a new function is added, let's say, a function that saves a book and stores it into a shopping cart for instance. Now the information is _scoped_ inside the `main()` function and inaccessible for use. So despite it functioning, it might be better to declare these variables outside the function and let the function handle only the logic of displaying the books within the DOM (or HTML document).  

Using a variable name such as `information` can be tricky, since the information is relating to the books I would suggest using something to the likes of `bookInformation`, `bookImages` etc. This makes sure that these variables are not reserved (since your only able to declare them once), I would personally only use such variable names if these are generic for global sets of information, let's say for instance information about the development environment (such as browser type).

Now without getting to much in detail about the variable names, the classnames used within the HTML should not be presented as camelcase. So `divTag.className = 'myBooks';` could be rewritten to something like `divTag.className = 'my-books';`. The HTML and related CSS classes have a slightly different approach, as discussed within the HTML classes. So despite the fact you are declaring it inside the JS file, you would still refer to these classnames inside your CSS file, it makes more sense to camelcase related to JS only.

You have used the `forEach` method to declare a loop. It's a relative new function your able to use, however the performance of the `forEach` method is quiet slow and has some other known issues witch you should be aware of using it in the future, please read this discussion (comment section) https://coderwall.com/p/kvzbpa/don-t-use-array-foreach-use-for-instead. 

So I would suggest for now using the good old `for()` loop. Something like `for (let i = 0; i < example.length; i++)` should do! It's also possible to `map()` an array of values, but that might be something we should discuss during future classes.

I liked that you contained certain blocks of code, such as first declaring the element, then declaring a title and other attributes, it reads very nice and it's considered a good thing to do. Overall you did good, but there are somethings you should refactor. Hope you find my comments useful!

See you in class tomorrow!

Greets,

Sander

soolaf:
Hi Sander,
thanks for your feedback ,you have alerted me to many important things ,I will do it in the right way, I benefit greatly from your feedback.
greetings,
Soolaf

PR #226, opened 2019-01-14
Ammar Vehbi - JS2 Homework Week2

ammarvehbi:


Week2/homework/maartjes-work.js
```diff
@@ -46,16 +46,29 @@ const maartjesTasks = monday.concat(tuesday);
 const maartjesHourlyRate = 20;
 
 function computeEarnings(tasks, hourlyRate) {
-  // Replace this comment and the next line with your code
-  console.log(tasks, hourlyRate);
+  const result =
+    // step 1.2.4
+    Math.floor(
```
remarcmij: Rounding this number to two decimals is only required for presentation purpose (i.e. showing the result to the user). By rounding at this stage the result this function can no longer be used for accurate subsequent calculations (because of loss of precision). It is best to delay the rounding until the result must be presented (i.e. outside of this function).

Week2/homework/maartjes-work.js
```diff
@@ -46,16 +46,29 @@ const maartjesTasks = monday.concat(tuesday);
...
+      100 *
+        tasks
+          // step 1.2.1
+          .map(x => x.duration / 60)
```
remarcmij: Do not use name `x` unless you actually mean the algebraic 'unknown'. Using `x` here requires the reader of your code to mentally 'map' `x` to mean `task`. This is a burden for the reader. Your code becomes easier to understand if you are explicit in your naming, If you are mapping `tasks` then a single element is a `task`:

```js
```

Week2/homework/maartjes-work.js
```diff
@@ -46,16 +46,29 @@ const maartjesTasks = monday.concat(tuesday);
...
+          // step 1.2.2
+          .filter(x => x >= 2)
```
remarcmij: Since you have mapped `tasks` to `hours`, this line could now read:

```js
.filter(hour => hour >= 2)
```

Week2/homework/maartjes-work.js
```diff
@@ -46,16 +46,29 @@ const maartjesTasks = monday.concat(tuesday);
...
+          // step 1.2.3
+          .map(x => x * hourlyRate)
```
remarcmij: `x` is still an `hour`:

```js
```

Week2/homework/maartjes-work.js
```diff
@@ -46,16 +46,29 @@ const maartjesTasks = monday.concat(tuesday);
...
+          .reduce((accumulator, number) => accumulator + number)
```
remarcmij: The `number` here is an `amount`. And I would choose `total` for the accumulator, because you are computing a running total. I have made it a habit to always explicitly add the optional, second argument. 

```js
.reduce((total, amount) => total + amount, 0)
```

Week2/homework/maartjes-work.js
```diff
@@ -46,16 +46,29 @@ const maartjesTasks = monday.concat(tuesday);
...
+    ) / 100;
+  return result;
 }
 
 // eslint-disable-next-line no-unused-vars
 const earnings = computeEarnings(maartjesTasks, maartjesHourlyRate);
+console.log(earnings);
```
remarcmij: This is a 'debug' type console.log which should not present in final code (i.e. when you submit a PR).

Week2/homework/maartjes-work.js
```diff
@@ -46,16 +46,29 @@ const maartjesTasks = monday.concat(tuesday);
...
 
 // add code to convert `earnings` to a string rounded to two decimals (euro cents)
 
-console.log(`Maartje has earned €${'replace this string with the earnings rounded to euro cents'}`);
+console.log(`Maartje has earned €${earnings}`);
```
remarcmij: Here, at presentation, is the correct point to round of the earnings. Rather than using `Math.round()` you can simply use `Number.toFixed(2)`:

```js
console.log(`Maartje has earned €${earnings.toFixed(2)}`);
```

Week2/homework/map-filter.js
```diff
@@ -1,8 +1,8 @@
 'use strict';
 
 function doubleOddNumbers(numbers) {
-  // Replace this comment and the next line with your code
-  console.log(numbers);
+  // step 1.1
+  return numbers.filter(x => x % 2 !== 0).map(x => x * 2);
```
remarcmij: Since this is mostly an algebraic operation without giving the numbers any meaning, the name `x` here is entirely appropriate.

ammarvehbi:
Hi @remarcmij , 

Thank you so so much for your valuable feedback. 

I thought that it would be better to use x to shorten the code but yes I realized that using x makes the code less understandable. I updated the code. The new changes can be seen in the new commit. Thanks again.

Kind regards

PR #227, opened 2019-01-14
week2 homework

ahmetbibi:


Week2/homework/maartjes-work.js
```diff
@@ -46,16 +46,16 @@ const maartjesTasks = monday.concat(tuesday);
 const maartjesHourlyRate = 20;
 
 function computeEarnings(tasks, hourlyRate) {
-  // Replace this comment and the next line with your code
-  console.log(tasks, hourlyRate);
+  const durationInHours = tasks.map(minute => minute.duration / 60);
```
remarcmij: If you map an array of `tasks` then an individual element of that array is a `task`, not a `minute`. So this line would be more accurate if it read like this:

```js
const durationsInHours = tasks.map(task => task.duration / 60);
```

Note that I have added an 's` after 'duration' in the name `durationsInHours` to make explicit that we get back an array.

Week2/homework/maartjes-work.js
```diff
@@ -46,16 +46,16 @@ const maartjesTasks = monday.concat(tuesday);
...
+  const moreThan2Hours = durationInHours.filter(newArray => newArray >= 2);
```
remarcmij: Since we are filtering the array `durationsInHours` a single element is a `durationInHours`, not a `newArray`. Therefore this line would be more accurate as it read as:

```js
const moreThan2Hours = durationsInHours.filter(durationInHours => durationInHours >= 2);
```

The variable name `moreThan2Hours` is incorrect, because we also include durations that are equal to two. Therefore you should rename this to a more accurate name, e.g.: `durationsEqualOrMoreThan2Hours` os something similar.

Week2/homework/maartjes-work.js
```diff
@@ -46,16 +46,16 @@ const maartjesTasks = monday.concat(tuesday);
...
+  const wageForTotalHours = moreThan2Hours.map(wage => wage * hourlyRate);
```
remarcmij: We not mapping wages here, we are mapping durations. So the parameter name `wage` is incorrect.

Week2/homework/maartjes-work.js
```diff
@@ -46,16 +46,16 @@ const maartjesTasks = monday.concat(tuesday);
...
+  const totalWage = wageForTotalHours.reduce((total, num) => total + num);
```
remarcmij: As you can see it is a hard job trying to invent accurate names for the variables holding the intermediate results. We can escape from this burden by using a method chain:

```js
  return tasks
    .filter(hour => hour >= 2)
    .reduce((total, amount) => total + amount, 0);
}
```

Now we have a concise, readable piece of code, with accurate names.

Note that storing the intermediate results into separate variables is a good approach if you are debugging, because once you have them in a method chain it becomes harder to debug.

Week2/homework/maartjes-work.js
```diff
@@ -46,16 +46,16 @@ const maartjesTasks = monday.concat(tuesday);
...
+  return totalWage;
 }
 
-// eslint-disable-next-line no-unused-vars
 const earnings = computeEarnings(maartjesTasks, maartjesHourlyRate);
-
-// add code to convert `earnings` to a string rounded to two decimals (euro cents)
-
-console.log(`Maartje has earned €${'replace this string with the earnings rounded to euro cents'}`);
+const roundedEarnings = Number(earnings.toFixed(2));
```
remarcmij: Excellent use of `.toFixed(2)` to round the number just prior to presenting the result to the user.

Week2/homework/map-filter.js
```diff
@@ -1,8 +1,7 @@
 'use strict';
 
 function doubleOddNumbers(numbers) {
-  // Replace this comment and the next line with your code
-  console.log(numbers);
+  return numbers.filter(n => n % 2 !== 0).map(n => n * 2);
```
remarcmij: I don't mind using single letters here for the parameter names, as this is sort of an algebraic operation where single letters are customary.

Week2/homework/maartjes-work.js
ahmetbibi: That's a good idea that we can use all of these functions (map, filter, map, reduce) in a chain without using a variable, and I learned how to use it now. Thank you very much for your useful and instructive comments.

PR #228, opened 2019-01-14
Bassam___JS2_Week2_HomeWork.

BassamHajar:
Thanks & Regards,,,

### DIFF
--- a/Week2/homework/maartjes-work.js
+++ b/Week2/homework/maartjes-work.js
@@ -48,8 +48,8 @@ const maartjesHourlyRate = 20;
 function computeEarnings(tasks, hourlyRate) {
   const totalEarnings = tasks
     .map(task => task.duration / 60)
-    .filter(rate => rate >= 2)
-    .map(rate => rate * hourlyRate)
+    .filter(hours => hours >= 2)
+    .map(hours => hours * hourlyRate)
     .reduce((acc, curVal) => Math.round((acc + curVal) * 100) / 100);
   return totalEarnings;
 }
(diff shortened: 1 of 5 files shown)


Week2/homework/maartjes-work.js
```diff
@@ -46,16 +46,21 @@ const maartjesTasks = monday.concat(tuesday);
...
 function computeEarnings(tasks, hourlyRate) {
-  // Replace this comment and the next line with your code
-  console.log(tasks, hourlyRate);
+  const totalEarnings = tasks
+    .map(task => task.duration / 60)
+    .filter(rate => rate >= 2)
+    .map(rate => rate * hourlyRate)
+    .reduce((acc, curVal) => Math.round((acc + curVal) * 100) / 100);
+  return totalEarnings;
 }
```
yash-kapila: Hi @BassamHajar - I liked how you used method chaining here to accomplish the desired results. I would like to add one thing here though. Although it might seem a small thing but giving appropriate names to variables inside methods like `map`, `filter` and `reduce` is quite important. I see that you have done a mix of both.

For example, the `map` method uses `task` as an internal variable correctly. However, `filter` method uses `rate` as the variable name when `hour` might have been more appropriate because we are filtering out hours here. Similarly, inside `reduce`, you have used `acc` and `curVal` which might have been coming from the APIs of the reduce method but having `total`, `amount` would have been more appropriate. Don't you think? 🙂 

Please be mindful of variable naming conventions. Even though they are small things but they happen to make you lot better.

Week2/homework/map-filter.js
```diff
@@ -1,8 +1,7 @@
 'use strict';
 
 function doubleOddNumbers(numbers) {
-  // Replace this comment and the next line with your code
-  console.log(numbers);
+  return numbers.filter(num => num % 2 !== 0).map(num => num * 2);
```
yash-kapila: Just FYI - you can also use arrow functions here to replace the function definition using `function` keyword with function expressions i.e
```
const doubleOddNumbers = numbers => numbers
  .filter(num => num % 2 !== 0)
  .map(num => num * 2);
```

PR #232, opened 2019-01-15
Week2

Ayh0o0m:


Week2/homework/maartjes-work.js
```diff
@@ -46,16 +46,18 @@ const maartjesTasks = monday.concat(tuesday);
 const maartjesHourlyRate = 20;
 
 function computeEarnings(tasks, hourlyRate) {
-  // Replace this comment and the next line with your code
-  console.log(tasks, hourlyRate);
+  const totalEarnings = tasks
+    .map(x => x.duration / 60)
```
remarcmij: When you call `.map()` on an array with `tasks`, the callback function that you pass as an argument to the `.map()` method gets a single `task` as its argument. Therefore this parameter should be called `task` and not `x` (which in algebra is equivalent for 'unknown'). If you use `x` the reader of your code must mentally map `x` to mean `task` to understand what is happening. You should minimise the burden for the reader to use mental mapping like that.

Week2/homework/maartjes-work.js
```diff
@@ -46,16 +46,18 @@ const maartjesTasks = monday.concat(tuesday);
...
+    .filter(x => x >= 2)
```
remarcmij: Since the mapping operation mapped tasks to hours, the `x` of of the filter callback function is now an `hour`.

Week2/homework/maartjes-work.js
```diff
@@ -46,16 +46,18 @@ const maartjesTasks = monday.concat(tuesday);
...
+    .map(x => x * hourlyRate)
```
remarcmij: `x` here is still an `hour`.

Week2/homework/maartjes-work.js
```diff
@@ -46,16 +46,18 @@ const maartjesTasks = monday.concat(tuesday);
...
+    .reduce((accumulator, number) => accumulator + number);
+  return totalEarnings;
 }
```
remarcmij: You could write the function like this, to show through the proper naming of parameters how the data is transformed and filter while it travels through the chain, without requiring the reader to do mental mapping:

```js
  return tasks
    .filter(hour => hour >= 2)
    .reduce((total, amount) => total + amount, 0);
}
```

Week2/homework/maartjes-work.js
```diff
@@ -46,16 +46,18 @@ const maartjesTasks = monday.concat(tuesday);
...
 
 // eslint-disable-next-line no-unused-vars
 const earnings = computeEarnings(maartjesTasks, maartjesHourlyRate);
 
-// add code to convert `earnings` to a string rounded to two decimals (euro cents)
-
-console.log(`Maartje has earned €${'replace this string with the earnings rounded to euro cents'}`);
+console.log(`Maartje has earned €${earnings.toFixed(2)}`);
```
remarcmij: Good find of the `.toFixed(2)` method to round to two decimals.

Week2/homework/map-filter.js
```diff
@@ -1,8 +1,8 @@
 'use strict';
 
 function doubleOddNumbers(numbers) {
-  // Replace this comment and the next line with your code
-  console.log(numbers);
+  numbers = numbers.filter(x => x % 2).map(x => x * 2);
```
remarcmij: Here, we are just working with numbers that have no higher meaning than being numbers. So the use of the algebraic name `x` is entirely appropriate here.

PR #237, opened 2019-01-16
week1 homework

AhmadAbdulhaleem:


Week1/homework/app.js
```diff
@@ -1,11 +1,149 @@
...
+  },
+  the_stranger: {
+    title: 'The-Stranger',
+    language: 'French',
+    author: 'Albert Camus'
+  }
+};
+
+// Create append function
+function modifyAppendChildFunc(parent = 'document.body', child) {
```
remarcmij: Setting a default value for a parameter is only useful at the end of the parameter list. If I want to use the default parameter as you have used it I would need to specify `undefined` for that parameter, e.g.:

```
modifyAppendChildFunc(undefined, headerElement);
```

So what would be the point?

But apart from that, nowhere in the rest of your code your are making use of this 'default' value. Therefore it is just noise.

With regard to what this function does: it only appends a child to a parent. But that is exactly what `parent.appendChild(...)` already does. So what is the point of this function?

Week1/homework/app.js
```diff
@@ -1,11 +1,149 @@
...
+  the_book_of_job: './img/220px-Aleppo_Codex_Joshua_1_1.jpg',
+  one_thousand_and_one_nights: './img/Ali_Baba,_by_Maxfield_Parrish.jpg',
+  pride_and_prejudice: './img/220px-PrideAndPrejudiceTitlePage.jpg',
+  the_decameron: './img/Boccaccio.jpg',
+  ficciones: './img/Ficciones.jpg',
+  the_stranger: './img/The_Strange.jpg'
+};
+
+// 1.8
+function coverPhoto() {
```
remarcmij: Function names should have a verb as the main word to indicate that they perform some action. In this case, you could call it `addCoverPhotos` or something similar.

PR #238, opened 2019-01-16
homework Week2

soolaf:


soolaf:
Hi @remarcmij 
Thank you for all your feedback , these points will be very important to me

PR #239, opened 2019-01-16
week2 homework

AhmadAbdulhaleem:


Week2/homework/maartjes-work.js
```diff
@@ -43,19 +43,28 @@ const tuesday = [
 ];
 
 const maartjesTasks = monday.concat(tuesday);
+
 const maartjesHourlyRate = 20;
 
 function computeEarnings(tasks, hourlyRate) {
   // Replace this comment and the next line with your code
-  console.log(tasks, hourlyRate);
+  const mappedTasks = tasks.map(task => task.duration / 60);
```
remarcmij: The result of mapping the tasks to hours produces an array of hours, not of tasks. So the name `mappedTasks` is incorrect and misleading. A better name would be `hours`.

```js
let hours = tasks.map(task => task.duration / 60);
```

Week2/homework/maartjes-work.js
```diff
@@ -43,19 +43,28 @@ const tuesday = [
...
+  const filteredTasks = mappedTasks.filter(task => task >= 2);
```
remarcmij: This operations should filter the `hours`, such that only hours >= 2 remain. The parameter name `task` is incorrect and misleading here. A correct line would be:

```js
hours = hours.filter(hour => hour >= 2);
```

Week2/homework/maartjes-work.js
```diff
@@ -43,19 +43,28 @@ const tuesday = [
...
+  const multiplyTasks = filteredTasks.map(task => task * hourlyRate);
```
remarcmij: In this step we are mapping hours to a an 'earning', by multiplying the filtered hours by the hourly rate. Again, the variable and parameter names must tell this story:

```js
const earnings = hours.map(hour => hour * hourlyRate);
```

Week2/homework/maartjes-work.js
```diff
@@ -43,19 +43,28 @@ const tuesday = [
...
+  const roundedTasks = multiplyTasks.map(task => task);
```
remarcmij: This line is complete useless. The `roundedTasks` array has the same values as the `multiplyTasks` array, because you are mapping each element onto itself. Furthermore, these names are incorrect because what you call `multiplyTasks` are actually `hours`.

Week2/homework/maartjes-work.js
```diff
@@ -43,19 +43,28 @@ const tuesday = [
...
+  const sumTasks = roundedTasks.reduce((a, b) => a + b);
```
remarcmij: You are summing up the earnings here, so the name `sumTasks` is not entirely wrong but nor very informative either. A better name would be `totalEarnings`.

Week2/homework/maartjes-work.js
```diff
@@ -43,19 +43,28 @@ const tuesday = [
...
+  // console.log(tasks, hourlyRate);
 }
 
 // eslint-disable-next-line no-unused-vars
 const earnings = computeEarnings(maartjesTasks, maartjesHourlyRate);
 
 // add code to convert `earnings` to a string rounded to two decimals (euro cents)
 
-console.log(`Maartje has earned €${'replace this string with the earnings rounded to euro cents'}`);
+console.log(`Maartje has earned €${earnings.toFixed(2)}`);
```
remarcmij: Excellent use of `.toFixed(2)` to round off to two decimals.

Week2/homework/map-filter.js
```diff
@@ -1,8 +1,8 @@
 'use strict';
 
 function doubleOddNumbers(numbers) {
-  // Replace this comment and the next line with your code
-  console.log(numbers);
+  const oddNumbers = numbers.filter(number => number % 2 !== 0).map(number => number * 2);
+  return oddNumbers;
```
remarcmij: Introducing the variable `oddNumbers` here does not add value. You can simply return the result straightaway, e.g.:

```js
return numbers.filter(number => number % 2 !== 0).map(number => number * 2);
```

remarcmij:
Hi @AhmadAbdulhaleem, I think you still need to follow-up on the feedback in this PR.

PR #240, opened 2019-01-16
homework week2

alquinex:
homework2

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -59,3 +59,4 @@ typings/
 
 .netlify
 dist/
+live
--- a/Week2/homework/maartjes-work.js
+++ b/Week2/homework/maartjes-work.js
@@ -46,16 +46,21 @@ const maartjesTasks = monday.concat(tuesday);
 const maartjesHourlyRate = 20;
 
 function computeEarnings(tasks, hourlyRate) {
-  // Replace this comment and the next line with your code
+  const durationInHours = tasks.map(x => x.duration / 60);
+  const filterHours = durationInHours.filter(x => x >= 2);
+  for (let i = 0; i < filterHours.length; i++) {
+    filterHours[i] = filterHours[i] * hourlyRate;
+  }
   console.log(tasks, hourlyRate);
+
+  const sumEarnings = filterHours.reduce((total, amount) => total + amount);
+  const roundSumEarnings = Math.round(sumEarnings * 100) / 100;
+  return roundSumEarnings;
 }
 
-// eslint-disable-next-line no-unused-vars
 const earnings = computeEarnings(maartjesTasks, maartjesHourlyRate);
 
-// add code to convert `earnings` to a string rounded to two decimals (euro cents)
-
-console.log(`Maartje has earned €${'replace this string with the earnings rounded to euro cents'}`);
+console.log(`Maartje has earned €${earnings}`);
 
 // Do not change or remove anything below this line
 module.exports = {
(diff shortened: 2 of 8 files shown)


remarcmij:
This comment is still outstanding:

>The only issue with this PR is that it contains the Live folder which is not part of the homework and shouldn't be in this PR. Can you please remove that?

.gitignore
```diff
@@ -59,3 +59,4 @@ typings/
 
 .netlify
 dist/
+live
```
remarcmij: The .gitignore file only works for files and directories that are not already tracked by Git. Adding an entry to .gitignore afterwards has no effect. Consequently, your PR still contains the `Live` folder. You can just delete that folder in VSCode from your week2 branch, commit and push to update the PR.

Week2/homework/maartjes-work.js
```diff
@@ -46,16 +46,21 @@ const maartjesTasks = monday.concat(tuesday);
 const maartjesHourlyRate = 20;
 
 function computeEarnings(tasks, hourlyRate) {
-  // Replace this comment and the next line with your code
+  const durationInHours = tasks.map(x => x.duration / 60);
+  const filterHours = durationInHours.filter(x => x >= 2);
+  for (let i = 0; i < filterHours.length; i++) {
```
remarcmij: Can you think of an array method (`forEach`, `map`, `filter` or `reduce`) that can replace this `for` loop?

Week2/homework/maartjes-work.js
```diff
@@ -46,16 +46,21 @@ const maartjesTasks = monday.concat(tuesday);
 const maartjesHourlyRate = 20;
 
 function computeEarnings(tasks, hourlyRate) {
-  // Replace this comment and the next line with your code
+  const durationInHours = tasks.map(x => x.duration / 60);
```
remarcmij: I covered in the week3 lecture my reservations regarding the use of meaningless names such as `x`.

Week2/homework/maartjes-work.js
```diff
@@ -46,16 +46,21 @@ const maartjesTasks = monday.concat(tuesday);
...
+  const filterHours = durationInHours.filter(x => x >= 2);
+  for (let i = 0; i < filterHours.length; i++) {
+    filterHours[i] = filterHours[i] * hourlyRate;
+  }
   console.log(tasks, hourlyRate);
```
remarcmij: You should take out all debug-type console.log statements when you submit a pull request.

Week2/homework/maartjes-work.js
```diff
@@ -46,16 +46,21 @@ const maartjesTasks = monday.concat(tuesday);
...
+
+  const sumEarnings = filterHours.reduce((total, amount) => total + amount);
+  const roundSumEarnings = Math.round(sumEarnings * 100) / 100;
```
remarcmij: This rounding inside the `computeEarnings` function causes a loss of precision, which make this function unsuitable for use in further calculations. You should round only at the point where the result needs to be displayed to the user, which in this case in, or just before the console.log statement of line 63. The `.toFixed(2)` method can be used to fix the number of decimals to 2. No need for the `Math.round()` calculation.

Week2/homework/map-filter.js
```diff
@@ -1,14 +1,13 @@
 'use strict';
 
 function doubleOddNumbers(numbers) {
-  // Replace this comment and the next line with your code
-  console.log(numbers);
+  const newNumbers = numbers.filter(x => x % 2 === 1).map(y => y * 2);
```
remarcmij: In this case, I have no objection against using names such as `x` or `y`, since we are dealing with numbers that in this context have no higher meaning other than being numbers.

PR #241, opened 2019-01-16
Week1

alquinex:
homework1


Week1/homework/app.js
```diff
@@ -2,10 +2,128 @@
...
-  console.log(bookTitles);
+  function createList() {
+    const ul = document.createElement('ul');
+    for (let i = 0; i < bookTitles.length; i++) {
+      const li = document.createElement('li');
+      ul.appendChild(li);
+      li.setAttribute('id', bookTitles[i]);
+    }
+  }
+  window.onload = createList;
```
remarcmij: You can assign only a single value to `window.onload`. The last value you assign overrides earlier values. Therefore the only function that gets executed is the `displayBooks` function, because the last assignment to `window.onload` is in line 128.

remarcmij:
This PR is still awaiting your follow-up.

PR #245, opened 2019-01-21
Ammar Vehbi - JS2 Homework Week3

ammarvehbi:


Week3/homework/step2-2.js
```diff
@@ -2,19 +2,27 @@
...
+    if (number % 5 === 0) {
+      fiveCallback(number);
+    }
+  }
 }
 
 function sayThree(number) {
-  // Replace this comment and the next line with your code
-  console.log(number);
+  return console.log(`${number} is divisible by three.`);
```
remarcmij: 👍 for using a template literal.

Week3/homework/step2-3.js
```diff
@@ -18,8 +19,9 @@ function repeatStringNumTimesWithWhile(str, num) {
   // eslint-disable-next-line
   let result = '';
 
-  // Replace this comment and the next line with your code
-  console.log(str, num, result);
+  while (result.length < str.length * num) {
```
remarcmij: The loop condition requires computation (property access, multiplication) for each iteration. It would be more efficient to introduce a loop index variable and simplify the loop condition, e.g.:

```js
let i = 0;
while (i < num) {
  //...
  i++;
}
```

This mimics how the `for` loop works.

Week3/homework/step2-3.js
```diff
@@ -31,8 +33,9 @@ function repeatStringNumTimesWithDoWhile(str, num) {
   // eslint-disable-next-line
   let result = '';
 
-  // Replace this comment and the next line with your code
-  console.log(str, num, result);
+  do {
+    result += str;
+  } while (result.length < str.length * num);
```
remarcmij: See my comment on the `while` loop with respect to the loop condition.

Week3/homework/step2-3.js
```diff
@@ -31,8 +33,9 @@ function repeatStringNumTimesWithDoWhile(str, num) {
   // eslint-disable-next-line
   let result = '';
 
-  // Replace this comment and the next line with your code
-  console.log(str, num, result);
+  do {
```
remarcmij: Because the `do-while` loop test the loop condition at the end of the loop (and therefore executes the loop body at least once), you must test yourself, before entering the loop, whether `num <=0`.

Week3/homework/step2-6.js
```diff
@@ -4,13 +4,21 @@ const arr2d = [[1, 2], [3, 4], [5, 6]];
...
 
 function printArray2d(arr) {
-  // Replace this comment and the next line with your code
-  console.log(arr);
+  let result = '';
+  for (let i = 0; i < arr.length; i++) {
+    result += arr[i].join('');
+  }
+  return console.log(result);
 }
```
remarcmij: Perhaps you peeked at the unit test to see what it expected, however, you could just console.log each array element directly:

```js
function printArray2d(arr) {
      console.log(arr[i][j]);
    }
  }
}
```

Having said that, your solution passes the test.

Week3/homework/step2-6.js
```diff
@@ -4,13 +4,21 @@ const arr2d = [[1, 2], [3, 4], [5, 6]];
...
-  // Replace this comment and the next line with your code
-  console.log(arr);
+  let result = '';
+  for (let i = 0; i < arr.length; i++) {
+    for (let j = 0; j < arr[i].length; j++) {
+      result += arr[i][j].join('');
+    }
+  }
+  return console.log(result);
 }
```
remarcmij: Ditto here.

Week3/homework/step2-7.js
```diff
@@ -20,4 +20,4 @@ f2(y);
 
 console.log(y);
 
-// Add your explanation as a comment here
+// in both examples, the actual value of x changes. Because both of the functions use the arguments which are passed as reference, not as value.
```
remarcmij: As you will have seen, the value of `x` doesn't change in the first case. Therefore your explanation is incorrect.

ammarvehbi: Hi @remarcmij ,
It is my bad. I did it very quicklky to study for the exam. I realized that my answer is incorrect. The new explanation could be seen in my new commit. Thanks for your valuable feedback. :pray:

Week3/homework/step2-3.js
```diff
@@ -31,8 +35,13 @@ function repeatStringNumTimesWithDoWhile(str, num) {
   // eslint-disable-next-line
   let result = '';
 
-  // Replace this comment and the next line with your code
-  console.log(str, num, result);
+  let i = 0;
+  do {
+    if (num > 0) {
```
remarcmij: By placing the `if` statement inside the loop it is executed for each loop iteration. However, it needs to be tested once only, before the loop executes.

Week3/homework/step2-7.js
```diff
@@ -20,4 +20,5 @@ f2(y);
 
 console.log(y);
 
-// Add your explanation as a comment here
+// Since in the first example the parameter passed by value, the value of x does not change.
+// As for the second example, since the parameter passed by reference, the value of y changes.
```
remarcmij: 👍

Week3/homework/step2-3.js
```diff
@@ -36,12 +36,12 @@ function repeatStringNumTimesWithDoWhile(str, num) {
   let result = '';
 
   let i = 0;
-  do {
-    if (num > 0) {
+  if (num > 0) {
```
remarcmij: 👍

ammarvehbi: :pray:

Week3/homework/step2-7.js
ammarvehbi: :pray:

Week3/homework/step2-2.js
ammarvehbi: :pray:

Week3/homework/step2-2.js
```diff
@@ -2,19 +2,27 @@
...
   const numbers = [];
-
-  // Replace this comment and the next line with your code
-  console.log(startIndex, stopIndex, threeCallback, fiveCallback, numbers);
+  for (let number = startIndex; number <= stopIndex; number++) {
+    numbers.push(number);
+  }
+  console.log(numbers);
+  for (let i = 0; i < numbers.length; i++) {
+    const number = numbers[i];
```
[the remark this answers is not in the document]
ammarvehbi: :pray:

ammarvehbi:
> LGTM (Looks Good To Me)

Thanks to you @remarcmij :pray:

PR #248, opened 2019-01-21
week3 homework

ahmetbibi:


Week3/homework/step2-2.js
```diff
@@ -3,18 +3,29 @@
...
 
-  // Replace this comment and the next line with your code
-  console.log(startIndex, stopIndex, threeCallback, fiveCallback, numbers);
+  for (let i = startIndex; i <= stopIndex; i++) {
+    numbers.push(i);
+  }
+
+  console.log(numbers);
+
+  numbers.map(number => {
```
remarcmij: The purpose of the `.map()` method is to create a new array by transforming (i.e. _mapping_) each element of the original through some transformation callback function. Here, you are just iterating through the array. You have no use for the array that `.map()` produces. A simple `.forEach()` will do here. In that case you also can get rid of line 19: there is no need to return anything.

Using a `.map()` method like you do here is confusing to the reader of your code as there will be some expectation that some mapping is required, while there is no such requirement.

Week3/homework/step2-3.js
```diff
@@ -5,8 +5,13 @@ function repeatStringNumTimesWithFor(str, num) {
...
 
-  // Replace this comment and the next line with your code
-  console.log(str, num, result);
+  if (num <= 0) {
+    result = '';
+  } else {
+    for (let i = 0; i < num; i++) {
+      result += str;
+    }
+  }
```
remarcmij: Guess what: your code still works fine when I comment out lines 8-10 and line 14. Can you explain why?

Week3/homework/step2-3.js
```diff
@@ -18,8 +23,15 @@ function repeatStringNumTimesWithWhile(str, num) {
...
-  console.log(str, num, result);
+  if (num <= 0) {
+    result = '';
+  } else {
+    let i = 0;
+    while (i < num) {
+      result += str;
+      i++;
+    }
+  }
```
remarcmij: See my previous comment on redundant code.

Week3/homework/step2-3.js
```diff
@@ -31,8 +43,15 @@ function repeatStringNumTimesWithDoWhile(str, num) {
   // eslint-disable-next-line
   let result = '';
 
-  // Replace this comment and the next line with your code
-  console.log(str, num, result);
+  if (num <= 0) {
```
remarcmij: Here, you need to test the condition of `num` before entering the loop as the `do-while` loop condition is tested at the end of the loop. This means that the loop body is executed at least once, without taking the loop condition into account. However, you do not need to re-initialise the `result` variable to the empty string as you already did that in line 44.

The fact that the `do-while` condition is tested at the end of the loop makes this construct rather useless. A `do-while` loop can always be rewritten as regular `while` loop. Using a `do-while` loop is like having unsafe sex on your first date 😁.

Week3/homework/step2-6.js
```diff
@@ -4,15 +4,27 @@ const arr2d = [[1, 2], [3, 4], [5, 6]];
...
-  // Replace this comment and the next line with your code
-  console.log(arr);
+  let matrix = '';
+  for (let i = 0; i < arr.length; i++) {
+    for (let j = 0; j < arr[i].length; j++) {
+      matrix += arr[i][j];
+    }
+  }
+  console.log(matrix);
 }
```
remarcmij: You could console.log each element in the inner `for` loop directly. This would have passed the unit tests too:

```js
function printArray2d(arr) {
      console.log(arr[i][j]);
    }
  }
}
```

Week3/homework/step2-6.js
```diff
@@ -4,15 +4,27 @@ const arr2d = [[1, 2], [3, 4], [5, 6]];
...
-  console.log(arr);
+  let matrix = '';
+  for (let i = 0; i < arr.length; i++) {
+    for (let j = 0; j < arr[i].length; j++) {
+      for (let k = 0; k < arr[j].length; k++) {
+        matrix += arr[i][j][k];
+      }
+    }
+  }
+  console.log(matrix);
```
remarcmij: The comment I made for the 2d version applies here too.

Week3/homework/step2-7.js
```diff
@@ -20,4 +20,9 @@ f2(y);
 
 console.log(y);
 
-// Add your explanation as a comment here
+// In f1 function if we use console.log(f1(x)) instead of console.log(x) it writes 10 on the log.
+// This is why f1 function returns the new value of x.
+// But outside the function f1, the primitive value of x is still keeping its first value 9 because of
+// f1 function's local scope.
+// Objects are not like primitive and you can change the objects' first value with functions
+// even if you call it outside the function.
```
remarcmij: I like your explanation.

Week3/homework/step3-bonus.js
```diff
@@ -3,8 +3,13 @@
...
-  // Replace this comment and the next line with your code
-  console.log(arr);
+  const reducer = (accumulator, currentValue) => {
+    if (!accumulator.includes(currentValue)) {
+      accumulator.push(currentValue);
+    }
+    return accumulator;
+  };
+  return arr.reduce(reducer, []);
 }
```
remarcmij: There are multiple ways of doing this but this is a perfectly acceptable solution.

Week3/homework/step3.js
```diff
@@ -1,8 +1,8 @@
 'use strict';
 
 function createBase(base) {
-  // Replace this comment and the next line with your code
-  console.log(base);
+  base = a => a + 6;
+  return base;
```
remarcmij: The `createBase` function should accept any base (not just 6). The output from the code below will be 16 and not 20 as I would expect.

```js
const addTen = createBase(10);
console.log(addTen(10)); 
```

Week3/homework/step2-3.js
ahmetbibi: Because in this case `num` variable is always positive and if you comment ou that lines code will still continue to work :) I delete these parts, thanks.

Week3/homework/step2-3.js
```diff
@@ -42,17 +30,11 @@ console.log('while', repeatStringNumTimesWithWhile('abc', 3));
...
-  } else {
-    let i = 0;
-    do {
-      result += str;
-      i++;
-    } while (i < num);
-  }
-
+  let i = 0;
+  do {
```
remarcmij: As the loop condition for the `do-while` loop is tested at the end, the loop body will always be executed at least once. In this case therefore you do need an `if` statement to ensure that the loop body is not executed if `num <= 0`. This illustrates why a `do-while` loop is pretty useless in practice. And it can always be rewritten using a (safe) regular `while` loop.

ahmetbibi: I see now dear Jim sorry for late understand.

PR #249, opened 2019-01-21
homework Week3

soolaf:


Week3/homework/step2-6.js
```diff
@@ -4,13 +4,17 @@ const arr2d = [[1, 2], [3, 4], [5, 6]];
 const arr3d = [[[1, 2], [3, 4]], [[5, 6], [7, 8]]];
 
 function printArray2d(arr) {
-  // Replace this comment and the next line with your code
-  console.log(arr);
+  for (let i = 0; i < arr.length; i++) {
+    console.log(arr[i].join(''));
+  }
```
remarcmij: The code we expected you to write is:

```js
function printArray2d(arr) {
      console.log(arr[i][j]);
    }
  }
}
```

However, your code passes the unit tests, so it would be unfair to reject it here.

Week3/homework/step2-6.js
```diff
@@ -4,13 +4,17 @@ const arr2d = [[1, 2], [3, 4], [5, 6]];
...
 
 function printArray3d(arr) {
-  // Replace this comment and the next line with your code
-  console.log(arr);
+  for (let i = 0; i < arr.length; i++) {
+    for (let j = 0; j < arr[i].length; j++) {
+      console.log(arr[i][j].join(''));
+    }
+  }
 }
```
remarcmij: Similar to the 2d example, you could three nested `for` loops here.

Week3/homework/step2-2.js
```diff
@@ -2,19 +2,23 @@
...
   const numbers = [];
-
-  // Replace this comment and the next line with your code
-  console.log(startIndex, stopIndex, threeCallback, fiveCallback, numbers);
+  for (let i = startIndex; i <= stopIndex; i++) {
+    numbers.push(i);
+  }
+  console.log(numbers);
+  for (const number of numbers) {
+    if (number % 3 === 0) threeCallback(number);
```
remarcmij: I would recommend that you always use curly braces around the `if` block, even it contains one statement only. This makes it easy to add more statements later. But if you nevertheless leave the curly braces out then I prefer that you do it the way you have done. The example below is particularly dangerous:

```js
if (number % 3 === 0)
  threeCallback(number);
```
This is because you might be tempted to add something, believing that is part of the `if` block, as suggested by the indentation:

 ```js
if (number % 3 === 0)
  console.log(number);
  threeCallback(number);
```

File: Week3/homework/step2-2.js [hunk 4]
```diff
@@ -2,19 +2,23 @@
 
 function threeFive(startIndex, stopIndex, threeCallback, fiveCallback) {
   const numbers = [];
-
-  // Replace this comment and the next line with your code
-  console.log(startIndex, stopIndex, threeCallback, fiveCallback, numbers);
+  for (let i = startIndex; i <= stopIndex; i++) {
+    numbers.push(i);
+  }
+  console.log(numbers);
```
remarcmij: This console.log (used for debugging purposes) should be removed when you submit your code in a pull request.

Week3/homework/step2-3.js
```diff
@@ -30,10 +33,10 @@ console.log('while', repeatStringNumTimesWithWhile('abc', 3));
 function repeatStringNumTimesWithDoWhile(str, num) {
   // eslint-disable-next-line
   let result = '';
-
-  // Replace this comment and the next line with your code
-  console.log(str, num, result);
-
+  do {
```
remarcmij: If `num <=0` the loop body of the `do-while` loop will be executed, where it shouldn't. This is because with a `do-while` loop the loop condition is tested at the end of the loop, making `do-while` rather useless in practice. You can always rewrite a `do-while` as a regular `while` loop. But here, where you are asked to use a `do-while` loop you must add an `if` statement to check whether the `do-while` loop should be executed at all.

Week3/homework/step2-7.js
```diff
@@ -21,3 +21,5 @@ f2(y);
 console.log(y);
 
 // Add your explanation as a comment here
+// in the first example :number passed to function by value so change the argument inside the function.
+// in the second example :objects passed by reference and changing the argument inside the function affect the variable passed from outside the function.
```
remarcmij: 👍 for your explanation.

soolaf: Hi @remarcmij  ,
Thanks for your feedback ,I will try to fix it

soolaf:
> LGTM (Looks Good To Me)

Thank you very much @remarcmij

PR #251, opened 2019-01-23
week3 homework

Ayh0o0m:


Week3/homework/step2-2.js
```diff
@@ -3,18 +3,26 @@
 function threeFive(startIndex, stopIndex, threeCallback, fiveCallback) {
   const numbers = [];
 
-  // Replace this comment and the next line with your code
-  console.log(startIndex, stopIndex, threeCallback, fiveCallback, numbers);
+  for (let number = startIndex; number <= stopIndex; number++) {
+    numbers.push(number);
+  }
+  for (let i = 0; i < numbers.length; i++) {
+    const number = numbers[i];
```
remarcmij: 👍 for creating the 'convenience' variable so that you can use it instead of having to repeat `numbers[i]` throughout the loop body.

Week3/homework/step2-2.js
```diff
@@ -3,18 +3,26 @@
...
+    if (number % 5 === 0) {
+      fiveCallback(number);
+    }
+  }
 }
 
 function sayThree(number) {
-  // Replace this comment and the next line with your code
-  console.log(number);
+  console.log(`${number} is divisible by three.`);
```
remarcmij: Template literal: 👍

Week3/homework/step2-3.js
```diff
@@ -17,9 +17,9 @@ console.log('for', repeatStringNumTimesWithFor('abc', 3));
 function repeatStringNumTimesWithWhile(str, num) {
   // eslint-disable-next-line
   let result = '';
-
-  // Replace this comment and the next line with your code
-  console.log(str, num, result);
+  while (result.length < str.length * num) {
```
remarcmij: This loop condition looks overly complex. For every loop iteration a computation must be performed. Here, it would be more efficient and easier to understand if you just mimic the `for` loop, by introducing a loop index variable:

```js
let i = 0;
while (i < num) {
  //...
  i++;
}
```

Week3/homework/step2-3.js
```diff
@@ -31,8 +31,9 @@ function repeatStringNumTimesWithDoWhile(str, num) {
   // eslint-disable-next-line
   let result = '';
 
-  // Replace this comment and the next line with your code
-  console.log(str, num, result);
+  do {
```
remarcmij: What happens if `num <=0`? The `do-while` loop executes at least once because the loop condition is only tested at the end of the loop. This makes the `do-while` loop rather useless: it can always be rewritten with a regular (safe) `while` loop.

But since we asked you here to use a `do-while` loop, what code can you add to take care of `num <= 0`?

(My previous comment about the complex loop condition applies here too.)

Week3/homework/step2-6.js
```diff
@@ -4,13 +4,21 @@ const arr2d = [[1, 2], [3, 4], [5, 6]];
 const arr3d = [[[1, 2], [3, 4]], [[5, 6], [7, 8]]];
 
 function printArray2d(arr) {
-  // Replace this comment and the next line with your code
-  console.log(arr);
+  let arrContent = '';
+  for (let i = 0; i < arr.length; i++) {
+    arrContent += arr[i].join('');
+  }
```
remarcmij: While this `for` loop passes the unit test, we had a nested `for` loop in mind:

```js
    console.log(arr[i][j]);
  }
}
```

Week3/homework/step2-6.js
```diff
@@ -4,13 +4,21 @@ const arr2d = [[1, 2], [3, 4], [5, 6]];
...
+  return console.log(arrContent);
```
remarcmij: There is not much point in trying to return anything from this function as the return value is unused. Besides that, you are returning the value that is returned by the console.log function. But console.log doesn't return anything (its return value is `undefined`), making this `return` statement doubly useless.

Week3/homework/step2-6.js
```diff
@@ -4,13 +4,21 @@ const arr2d = [[1, 2], [3, 4], [5, 6]];
...
-  // Replace this comment and the next line with your code
-  console.log(arr);
+  let arrContent = '';
+  for (let i = 0; i < arr.length; i++) {
+    for (let j = 0; j < arr[i].length; j++) {
+      arrContent += arr[i][j].join('');
+    }
+  }
+  return console.log(arrContent);
 }
```
remarcmij: The above comments apply to the 3d version too.

Week3/homework/step2-7.js
```diff
@@ -20,4 +20,8 @@ f2(y);
 
 console.log(y);
 
-// Add your explanation as a comment here
+// when we pass a variable in a function and that variable holds a number, javascript will take a copy of mentioned
+// variable and it passes the copy in that function. that means: the value of x in the first example doesn't change
+// because we passes a copy of x....
+// While in the second example we pass a variable that holds an object and here is the difference. WHen we pass an
+// object in a function, javascript takes that object a REFERENCE and modifies it.
```
remarcmij: 👍 for your explanation.

Week3/homework/step3-bonus.js
```diff
@@ -3,8 +3,7 @@
 const values = ['a', 'b', 'c', 'd', 'a', 'e', 'f', 'c'];
 
 function makeUnique(arr) {
-  // Replace this comment and the next line with your code
-  console.log(arr);
+  return [...new Set(arr)];
```
remarcmij: 👍 for the ES6 syntax

PR #254, opened 2019-01-23
Week3 Homework

AhmadAbdulhaleem:


Week3/homework/step2-6.js
```diff
@@ -4,13 +4,21 @@ const arr2d = [[1, 2], [3, 4], [5, 6]];
...
-  // Replace this comment and the next line with your code
-  console.log(arr);
+  for (let i = 0; i < arr.length; i++) {
+    if (arr[i] instanceof Array) {
+      printArray3d(arr[i]);
+    } else {
+      console.log(arr[i]);
+    }
+  }
 }
```
remarcmij: Excellent use of a recursive function (which we didn't cover in class). Since this recursive function can deal with arrays of any dimension the name `printArray3d` is too limiting. Can you rewrite step 2-6 such that you still have two functions `printArray2d` and `printArray3d` (as required to pass the unit tests) but that both make use of your recursive function and thus make your code DRY?

Week3/homework/step2-7.js
```diff
@@ -21,3 +21,5 @@ f2(y);
 console.log(y);
 
 // Add your explanation as a comment here
+// When I call the Function directly I can pass the value as a parameter and if I want to change the value of the argument which is inside the function then it will not affect the variable which is exist outside the scope
+// By reference way the inside value of the scope will affect the object or the array which is outside the scop
```
remarcmij: I think you sort of grasped the idea. It is the difference between 'passing by value' and 'passing by reference'.

Week3/homework/step3-bonus.js
```diff
@@ -3,8 +3,7 @@
 const values = ['a', 'b', 'c', 'd', 'a', 'e', 'f', 'c'];
 
 function makeUnique(arr) {
-  // Replace this comment and the next line with your code
-  console.log(arr);
+  return [...new Set(arr)];
```
remarcmij: 👍 for the ES6 syntax.

PR #256, opened 2019-01-23
Correct week3 branch

alquinex:


Week3/homework/step2-3.js
```diff
@@ -28,11 +27,11 @@ console.log('while', repeatStringNumTimesWithWhile('abc', 3));
 
 // Use a 'do...while' loop
 function repeatStringNumTimesWithDoWhile(str, num) {
-  // eslint-disable-next-line
   let result = '';
-
-  // Replace this comment and the next line with your code
-  console.log(str, num, result);
+  do {
```
remarcmij: What if `num <= 0`? Then the `do` loop body will run once anyway because the loop condition will only be executed at the end of the loop. This makes the `do-while` loop unsafe and is best replaced by a regular `while` loop which tests the condition before entering the loop body. We asked to use the `do-while` loop here to demonstrate this problem. What code can you add to prevent the `do-while` loop from running if `num <= 0`?

Week3/homework/step2-5.js
```diff
@@ -1,12 +1,12 @@
...
 function multiplyAll(arr) {
-  // eslint-disable-next-line
   let product = 1;
-
-  // Replace this comment and the next line with your code
-  console.log(arr, product);
-
+  for (let i = 0; i < arr.length; i++) {
+    for (let j = 0; j < arr[i].length; j++) {
+      product = product * arr[i][j];
```
remarcmij: You could use the short-hand notation ` product *= arr[i][j];` here.

Week3/homework/step2-6.js
```diff
@@ -4,13 +4,21 @@ const arr2d = [[1, 2], [3, 4], [5, 6]];
 const arr3d = [[[1, 2], [3, 4]], [[5, 6], [7, 8]]];
 
 function printArray2d(arr) {
-  // Replace this comment and the next line with your code
-  console.log(arr);
+  for (let i = 0; i < arr.length; i++) {
+    for (let b = 0; b < arr[i].length; b++) {
```
remarcmij: It is customary to use the letters `i`, `j`, `k`, `m` and `n` for nested loops (as you would in algebra). Note the letter `l` is to be avoided as it easily confused with the digit `1`.

Week3/homework/step2-7.js
```diff
@@ -20,4 +20,15 @@ f2(y);
...
+/** Pass by value is whereby a function is called by directly passing the value of
+ * the variable as the argument. The variable passed from outside the function is not
+ * affected as a result of changing the argument inside the function. In this case,
+ * the value of 'x' is not changed in the f1 function.
+ *
+ * While pass by reference is whereby a function is called directly by passing the
+ * reference or address of the variable as the argument. Changing the argument inside
+ * the function affects the variable passed in from outside the function as in the case
+ * of f2 function above.
+ */
```
remarcmij: Okay, the key point is passing by value as opposed to passing by reference.

alquinex: Hi Jim, 
All corrections have been made. It should be fine now

PR #257, opened 2019-01-23
week3-homework - fatih

kfth:


kfth:
I think I've missed out step 8 unintentionally that's why this problem occurs, sorry.

Week3/homework/step2-6.js
```diff
@@ -3,14 +3,23 @@
...
-  // Replace this comment and the next line with your code
-  console.log(arr);
+  printArrayNd(arr);
 }
 
 function printArray3d(arr) {
-  // Replace this comment and the next line with your code
-  console.log(arr);
+  printArrayNd(arr);
 }
```
remarcmij: Even simpler would be:

```js
const printArray2d = printArrayNd;
const printArray3d = printArrayNd;
```

These two variables are now just aliases for your `printArrayNd` function.

Week3/homework/step2-7.js
```diff
@@ -20,4 +20,7 @@ f2(y);
 
 console.log(y);
 
-// Add your explanation as a comment here
+/*  While dealing with primitives(strings, numbers etc.) Javascript always passes by value,
+that we can see in the first example original variable value (const x = 9) did not change.
+On the other hand while dealing with objects (objects, arrays) Javascript passes by reference.
+Therefore changing the object changes the original one. */
```
remarcmij: 👍 for your explanation

Week3/homework/step3-bonus.js
```diff
@@ -3,8 +3,12 @@
 const values = ['a', 'b', 'c', 'd', 'a', 'e', 'f', 'c'];
 
 function makeUnique(arr) {
-  // Replace this comment and the next line with your code
-  console.log(arr);
+  return arr.reduce((uniqueArray, value) => {
```
remarcmij: There are a couple of ways to solve this assignment. But this is a good one.

PR #260, opened 2019-01-25
week3HomeWork

Aya92syr:


Aya92syr:
week3 update step3

Aya92syr:
Last update week3 Aya

.eslintrc.json
```diff
@@ -6,7 +6,6 @@
     "jest": true
   },
   "rules": {
-    "prettier/prettier": ["error"],
```
remarcmij: You are not allowed to remove this line.

Aya92syr:
LastUpdateWeek3

Week3/homework/step2-2.js
```diff
@@ -2,19 +2,26 @@
...
+    if (number % 5 === 0) {
+      fiveCallback(number);
+    }
+  }
 }
 
 function sayThree(number) {
-  // Replace this comment and the next line with your code
-  console.log(number);
+  return console.log(`${number} is divisible by three.`);
```
remarcmij: 👍 for using a template literal

Week3/homework/step2-3.js
```diff
@@ -31,8 +35,13 @@ function repeatStringNumTimesWithDoWhile(str, num) {
   // eslint-disable-next-line
   let result = '';
 
-  // Replace this comment and the next line with your code
-  console.log(str, num, result);
+  let i = 0;
+  if (num > 0) {
```
remarcmij: 👍 for checking if the `do-while` loop should be entered at all. Because the `do-while` loop does not check the loop condition upfront, it is unsafe to use. It almost always better to use a regular `while` loop that checks the loop condition before entering the loop body. The purpose of this exercise was to demonstrate this.

Week3/homework/step2-6.js
```diff
@@ -4,13 +4,21 @@ const arr2d = [[1, 2], [3, 4], [5, 6]];
...
 function printArray3d(arr) {
-  // Replace this comment and the next line with your code
-  console.log(arr);
+  let result = '';
+  for (let i = 0; i < arr.length; i++) {
+    for (let j = 0; j < arr[i].length; j++) {
+      result += arr[i][j].join('');
+    }
+  }
+  return console.log(result);
```
remarcmij: There is not much point in return the result of a `console.log` statement: console.log returns `undefined`, i.e. it doesn't return anything. If you leave the `return` keyword everything still works.

Note that the version for the 2d array, with a simple console.log inside the inner loop, is sufficient to pass the test. When given a choice between a simpler solution and a more complex one that both achieve the same result, the simpler solution should take precedence.

Week3/homework/step2-7.js
```diff
@@ -20,4 +20,5 @@ f2(y);
 
 console.log(y);
 
-// Add your explanation as a comment here
+// E.g.1 :the number passed to function so that change the argument inside the function only (function's local scope).
+// E.g.2 : changing the argument inside the function affect the property of the object that passed in outside the function so object is deferent.
```
remarcmij: The difference is that primitive values (e.g. number, string and booleans) are passed by value and objects (including arrays and functions) are passed by reference.

PR #271, opened 2019-03-19
Ramazan Ince JS2 Hw 1 Week 1

RamazanInce:


Week1/homework/app.js
```diff
@@ -1,11 +1,144 @@
...
+
+      li.textContent = listOfBooks;
+      h2.textContent = listOfAuthors;
+      p.textContent = listOfLanguage;
+      img.src = bookImage;
+      img.alt = listAltImage;
+
+      ul.appendChild(li);
+      li.appendChild(h2);
+      h2.appendChild(p);
```
paulbremer: With `appendChild` you put the element inside the element you specify at the beginning, so in this case you put the `p` tag inside the `h2` title. Right now you have the `p` tag and the image `img` inside the title. Better would be to create a container `div` and put all the elements inside that container of in this case you could append everything to the `li` element since it is already the container for the elements.

rincemath: Hi @paulbremer 
I noticed that one after finishing the homework then I decide to search and find a way to resolve the issue but that week there were some additional task which I have to work on it then I left it like that.  I will work on it very soon.
Thanks million times for your great efforts.

PR #275, opened 2019-03-21
javascript2 week 1

alaashhadat:


rohni:
I see that you have eslint errors.  Could you please fix the errors and then I can review the code.

Week1/homework/app.js
```diff
@@ -1,11 +1,121 @@
...
+    const div = document.getElementById('root');
+    const head = document.createElement('h1');
+    head.innerText = 'My favorites books';
+    div.appendChild(head);
+    div.appendChild(ul);
+    for (let i = 0; i < bookTitles.length; i++) {
+      const li = document.createElement('li');
+      ul.appendChild(li);
+      const img = document.createElement("img");
+      li.appendChild(img);
```
rohni: Nice job, but according to the makeme.md, the adding of the bookcover must happen after the whole 'ul' is finished and attached.

Week1/homework/app.js
```diff
@@ -1,11 +1,121 @@
...
+  };
+  function listingBooks() {
+    const ul = document.createElement('ul');
+    const div = document.getElementById('root');
+    const head = document.createElement('h1');
+    head.innerText = 'My favorites books';
+    div.appendChild(head);
+    div.appendChild(ul);
+    for (let i = 0; i < bookTitles.length; i++) {
+      const li = document.createElement('li');
```
rohni: Pretty much everything in here would be more readable if they were broken up into different functions that are then called.  But this will get the job done.  The only problem is that it is less readable.

PR #276, opened 2019-03-21
JS2 Week1 Assignment-Semih

sigma21:


apps.js
```diff
@@ -0,0 +1,113 @@
...
+
+const root = document.querySelector('body');
+const bookList = document.createElement('ul');
+const header = document.createElement('h1');
+header.innerHTML = 'The Best Books of 2018 by Amazon';
+root.appendChild(bookList);
+bookList.appendChild(header);
+
+for (let i = 0; i < books.length; i++) {
+  const listItem = document.createElement('li');
```
rohni: I would suggest putting all the <li> generating code in its own function, and then just call the function in the loop.

apps.js
```diff
@@ -0,0 +1,113 @@
...
+  const cover = document.createElement('img');
+  const title = document.createElement('h2');
+  const author = document.createElement('h3');
+  const language = document.createElement('h4');
+
+  cover.setAttribute('src', bookCovers[books[i]]);
+  title.innerHTML = bookObj[books[i]].title;
+  author.innerHTML = bookObj[books[i]].author;
+  language.innerHTML = bookObj[books[i]].language;
+  listItem.style.border = '2px solid #950740';
```
rohni: This should be done in a css class.

sigma21: Noted! Will separate the css.

apps.js
sigma21: Sounds good, I'll try to implement that. Thanks for the idea!

sigma21:
Thanks a lot @rohni for the feedback!

PR #277, opened 2019-03-21
Orhan_JS2_Week1_Homework

OrhanGoc:


Orhan_JS2_Week1_Homework/apps.js
```diff
@@ -0,0 +1,140 @@
...
+  // }
+
+  function createBookList() {
+    const root = document.getElementById("root");
+    const ul = document.createElement("ul");
+    root.appendChild(ul);
+    const values = Object.keys(bookInfo);
+
+    for (let i = 0; i < values.length; i++) {
+      const li = document.createElement("li");
```
rohni: I would have put this contents of the loop in a function and then call that within the loop, but this works as well.

Orhan_JS2_Week1_Homework/apps.js
```diff
@@ -0,0 +1,140 @@
...
+      h2.innerText = bookInfo[values[i]].author;
+      li.appendChild(h2);
+
+      const h3 = document.createElement("h3");
+      h3.innerText = bookInfo[values[i]].language;
+      li.appendChild(h3);
+    }
+  }
+
+  function addImage() {
```
rohni: Excellent.

Orhan_JS2_Week1_Homework/apps.js
```diff
@@ -0,0 +1,140 @@
...
+    const values = Object.keys(bookCovers);
+    for (let i = 0; i < values.length; i++) {
+      const imageElem = document.createElement("img");
+      imageElem.setAttribute("src", bookCovers[values[i]]);
+      const liNewId = document.getElementById(values[i]);
+      liNewId.appendChild(imageElem);
+
+      console.log(imageElem);
```
rohni: Once you are done developing/debugging, unless you have a specific reason to leave it, it is a good idea to remove your logging.

Orhan_JS2_Week1_Homework/style.css
```diff
@@ -0,0 +1,30 @@
+
+.book_list ul {
```
rohni: Nice use of classes here.

Orhan_JS2_Week1_Homework/style.css
```diff
@@ -0,0 +1,30 @@
...
+}
+.book_list li {
+  float: left;
+  position: relative;
+  list-style: none;
+  color: blue;
+  border: 1px solid black;
+  margin: 10px;
+  padding: 10px;
+  width: 400px;;
```
rohni: Hard-coding width in pixels works here, but means that it is less flexible if you are looking at this page with a phone or tablet.

Orhan_JS2_Week1_Homework/apps.js
OrhanGoc: I have forgotten to remove logging:))

PR #279, opened 2019-03-21
Mohammad-Darwesh homework js2 week1

mohammadnaefaldarwesh:


Week1/Mohammad-Darwesh-JS2-week1/app.js
```diff
@@ -0,0 +1,126 @@
...
+      'https://images-na.ssl-images-amazon.com/images/I/5116zAISjzL._SX329_BO1,204,203,200_.jpg',
+    church_of_marvels:
+      'https://i.pinimg.com/originals/b8/19/0d/b8190d0fb7bcb1ced1bc4d69c3637c22.jpg',
+    after_the_rain:
+      'https://vicsmediaroom.files.wordpress.com/2016/04/after-the-rain-book-cover.jpg?w=308',
+  };
+  const div = document.createElement('div');
+  document.body.appendChild(div);
+  const header = document.createElement('h1');
+  const addHeader = document.body.getElementsByTagName('div')[0];
```
rohni: Just to remember if another div is added to the page somewhere, this code will break.  For this sort of thing I would have given the div an `id`.

Week1/Mohammad-Darwesh-JS2-week1/app.js
```diff
@@ -0,0 +1,126 @@
...
+  addHeader.appendChild(header);
+  header.innerHTML = 'My favorite books';
+  const list = document.createElement('ul');
+  div.appendChild(list);
+  list.setAttribute('class', 'books_list');
+  function makingList() {
```
rohni: Nice that you broke this into a function

Week1/Mohammad-Darwesh-JS2-week1/app.js
```diff
@@ -0,0 +1,126 @@
...
+    for (let item of bookTitles) {
```
rohni: I would have taken the contents of the loop and created another function where you create one complete list item, and then just call that function in the loop and pass in the book data.

Week1/Mohammad-Darwesh-JS2-week1/app.js
```diff
@@ -0,0 +1,126 @@
...
+      const booksList = document.createElement('li');
+      list.appendChild(booksList);
+      booksList.setAttribute('id', item);
```
rohni: You have a function later that inserts the cover (as should happen from the spec) why are you adding it here as well?

mohammadnaefaldarwesh: Hello Rohni ,
Thank you for your  feedback and sure i will consider your useful advises.
about insert the cover ......my idea from tow function that one for HTML structures(elements,tags,attributes) and the second one  for insert stuff in the elements.

PR #283, opened 2019-03-25
Ramazan ince js2 hw2

RamazanInce:


Week2/homework/maartjes-work.js
```diff
@@ -45,17 +45,21 @@ const tuesday = [
 const maartjesTasks = monday.concat(tuesday);
 const maartjesHourlyRate = 20;
 
-function computeEarnings(tasks, hourlyRate) {
-  // Replace this comment and the next line with your code
-  console.log(tasks, hourlyRate);
+function computeEarnings() {
```
paulbremer: You should use the `tasks` and `hourlyRate` as parameters for the function. It works fine right now but what if you want to run the function again with `paulsTasks`, the function is still looking at `maartjesTasks`. If the function looks only at the parameter `tasks` it will still work.

`function computeEarnings(tasks, hourlyRate) {`

This will work fine for the function with different parameters like these:
`computeEarnings(maartjesTasks, maartjesHourlyRate);`
`computeEarnings(paulsTasks, maartjesHourlyRate);`

rincemath: Hi @paulbremer thanks for your feedback and sacrificing your time for us., I understood what you mean. I will correct myself.

PR #284, opened 2019-03-25
Enes JavaScript2 Week 2 Homework

eneserdek:
npm run test-week2 COULD NOT realised.

### DIFF
--- a/Week2/enes-javascript2-homework.js
+++ b/Week2/enes-javascript2-homework.js
@@ -4,7 +4,7 @@
 
 const myNumbers = [1, 2, 3, 4];
 
-const doubleOddNumberss = myNumbers.filter(number => number % 2 === 1).map(number => number * 2);
+const doubleOddNumbers = myNumbers.filter(number => number % 2 === 1).map(number => number * 2);
 
 console.log(doubleOddNumbers);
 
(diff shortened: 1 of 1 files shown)


Week2/enes-javascript2-homework.js
```diff
@@ -0,0 +1,97 @@
+{
```
paulbremer: Hi Enes! You should have changed the files in the `Week2/homework` folder, that way running the test will also work.  

The first should be written as 

```
  const doubleNumbers = numbers.filter(number => number % 2 === 1).map(number => number * 2);
  return doubleNumbers;
}
```

If you run `npm run test-week2` you will see 1 test passed.

Week2/enes-javascript2-homework.js
```diff
@@ -0,0 +1,97 @@
...
+    .map(duration => duration * 20)
+    .reduce((sum, elem) => (sum += elem), 0);
+  const billing = totalHoursOfWork
+    .map(duration => duration * 20)
+    .reduce((sum, elem) => (sum += elem), 0);
+  const roundedBilling = Math.round(billing * 100) / 100;
+  return `€${roundedBilling};`;
+};
+
+console.log(finalBill(monday, tuesday /*,thursday*/));
```
paulbremer: While this function works it is not what is asked in the homework. Open `homework/maartjes-work.js` and implement the code over there. You should not edit the line with `const earnings = computeEarnings(maartjesTasks, maartjesHourlyRate);`

If you add the correct code in that file running the test will also work again.

PR #285, opened 2019-03-25
JS2-Week1-HW

saidcetinkaya:


Week1/homework/app.js
```diff
@@ -1,11 +1,179 @@
...
+// Question 1.8
+
+function createUl(ul) {
+  for (const key of Object.keys(bookSpecified)) {
+    const titleText = bookSpecified[key].title;
+    const languageText = bookSpecified[key].language;
+    const authorText = bookSpecified[key].author;
+    const titleElem = document.createElement('h1');
+    const imageElem = document.createElement('img');
+    imageElem.setAttribute('src', `${images.key}`);
```
paulbremer: You are searching for the `.key` but you don't want the actual string key but the value of the key, in that case you can use the bracket notion `images[key]`. The bracket notation can contain variables. 

Also, since it's already a string there is no need to wrap it in quotes; `'${images.key}'` can be written as `images[key]`.

Week1/homework/app.js
```diff
@@ -1,11 +1,179 @@
...
+    imageElem.setAttribute('alt', `${key}`);
```
paulbremer: Same here; `imageElem.setAttribute('alt', key);`

Week1/homework/style.css
```diff
@@ -1 +1,20 @@
-/* add your styling here */
\ No newline at end of file
+#root ul {
+  display:flex;
```
paulbremer: Good use of `flex`!

Week1/homework/style.css
```diff
@@ -1 +1,20 @@
...
+  flex-flow: row wrap;
+  list-style-type: none;
+  padding-inline-start: 0px;
```
paulbremer: The `padding-inline-start` is a new property and isn't supported in Edge and IE, just using `padding` would work better is all browsers.

Week1/homework/app.js
```diff
@@ -1,11 +1,179 @@
...
+}
+
+window.addEventListener('load', main);
+*/
+
+// Question 1.4
+
+const bookSpecified = {
+  booknumber0: {
+    title: 'title0',
```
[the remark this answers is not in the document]
saidcetinkaya: :)

Week1/homework/app.js
saidcetinkaya: Got it

Week1/homework/app.js
saidcetinkaya: 👍

PR #289, opened 2019-03-26
the homework of js 2 week 1

Nadoshka:


Week1/homework/app.js
```diff
@@ -1,11 +1,187 @@
-'use strict';
+/* eslint-disable no-unused-expressions */
```
paulbremer: `eslint-disable`? That is cheating haha, since it's not in the homework for week 1 its fine :)

Week1/homework/app.js
```diff
@@ -1,11 +1,187 @@
...
+    'forty_rules',
+    'memory_in_flesh',
+    'dust_of_promises',
+    'art_of_forgetting',
+    'black_suits_you',
+    'Chaos_Of_Senses',
+    'bed_hopper',
+    'game_of_thrones',
+  ];
+  console.log(bookTitles);
```
paulbremer: Remember to remove `console.log` if you don't need them in the pull request.

Week1/homework/app.js
```diff
@@ -1,11 +1,187 @@
...
+
+  const listElement = document.createElement('ul');
+  function listElements(ul, itemsArray) {
```
paulbremer: You don't use this function anymore, remember to clean up your code when you make a pull request.

Week1/homework/app.js
```diff
@@ -1,11 +1,187 @@
...
+  }
+  // elements();
+}
+// const load = () => {
+//   console.log('load event detected!');
+//   elements();
+// };
+// window.onload = load;
+{
+  ('use strict');
```
paulbremer: You only need to use `('use strict')` once at the top of the file.

Week1/homework/app.js
```diff
@@ -1,11 +1,187 @@
...
+      div.setAttribute('class', 'item-container');
+      const imageContainer = document.createElement('div');
+      div.appendChild(imageContainer);
+      imageContainer.setAttribute('class', 'image_container');
+      const contentContainer = document.createElement('div');
+      contentContainer.setAttribute('class', 'content_container');
+      div.appendChild(contentContainer);
+      const bookImage = document.createElement('img');
+      bookImage.setAttribute('src', booksPath[key]);
+      bookImage.setAttribute('alt', liElem.imageAlt);
```
paulbremer: The key should be `ImageAlt`, mind the capital `I`.

Week1/homework/app.js
Nadoshka: haha, No cheating! :) but I think my Eslint has technical problem, It gave me some messages and I treid to solve it but it did not work! I will try on sunday also to understand what is happen.

Week1/homework/app.js
Nadoshka: Thanks for your tip! I just want to practice the q on another way, and keep it to remember it later  when I need it, I thought if I did not call the function then it will not work, Is that right?

Week1/homework/app.js
Nadoshka: Good to know!

Week1/homework/app.js
Nadoshka: Oops! I understood from Naming Convention it should be camelCase. Am I right? 
Thanks in advance! :)

Week1/homework/app.js
paulbremer: Not sure if I understand your question, the `console.log` will always be outputted since it is not inside a function.

Week1/homework/app.js
paulbremer: Take a look at line 73, `ImageAlt: 'book cover Deep Work',` you have capital `I` which should be lowercase ;)

PR #293, opened 2019-03-27
Semih Week2

sigma21:


semih-week2.js
```diff
@@ -0,0 +1,69 @@
+'use strict';
```
paulbremer: Instead of putting the homework in a separate file you should edit the files in the homework folder. Look at the files at `Week2/homework/maartjes-work.js` and `Week2/homework/map-filter.js`. Look for this comment `// Replace this comment and the next line with your code`.  If you put your code in these files and run `npm run test-week2` in your terminal you can check if the test passes or not.

semih-week2.js
```diff
@@ -0,0 +1,69 @@
...
+
+const numbers = [1, 2, 3, 4, 5, 6, 7, 8, 9];
+const isOdd = x => x % 2 !== 0;
+const double = x => x * 2;
+const doubleOdd = numbers.filter(isOdd).map(double);
```
paulbremer: While this code works fine it is more common to put the different functions in different `const`. So instead of a const with `x => x % 2 !== 0` and using that const in the `filter()` you could also write it together like this

`const isOdd = numbers.filter(x => x % 2 !== 0);`

For the doubleOdd you can then use the isOdd value to do the `map()`. Again, not wrong but it is more readable to put the things together.

semih-week2.js
```diff
@@ -0,0 +1,69 @@
...
+  },
+];
+
+const minToHour = x => x.duration / 60; // map
+const moreThanTwoHours = x => x > 2; // filter
+const perHourRate = x => x * 20; // map
+const sum = (x, y) => x + y; // reduce
+
+const mondayHours = monday.map(minToHour);
+const tuesdayHours = tuesday.map(minToHour);
```
paulbremer: Take a look at the `maartjes-work.js` file, you will find the following line `const maartjesTasks = monday.concat(tuesday);` which will add the tasks together so you only need to `map()` once.

semih-week2.js
```diff
@@ -0,0 +1,69 @@
...
+    name: 'Talk to a lot of people',
+    duration: 200,
+  },
+  {
+    name: 'Look at application assignments new students',
+    duration: 40,
+  },
+];
+
+const minToHour = x => x.duration / 60; // map
```
paulbremer: Using `x` works but like these lines it is getting harder to grasp what you are working with. Try using more descriptive names, `task.duration` for example.

semih-week2.js
```diff
@@ -0,0 +1,69 @@
...
+const tuesdayHours = tuesday.map(minToHour);
+
+const mondayHoursFiltered = mondayHours.filter(moreThanTwoHours);
+const tuesdayHoursFiltered = tuesdayHours.filter(moreThanTwoHours);
+
+const mondayPerHourRates = mondayHoursFiltered.map(perHourRate);
+const tuesdayPerHourRates = tuesdayHoursFiltered.map(perHourRate);
+
+const mondayRateSum = mondayPerHourRates.reduce(sum, 0);
+const tuesdayRateSum = tuesdayPerHourRates.reduce(sum, 0);
```
paulbremer: You can already notice that you are repeating the same stuff for monday and tuesday. This is always a good time to look at your code again and see if you can refactor it. If you use the `concat()` function I mentioned earlier you don't need to do the same thing for the two days.

semih-week2.js
```diff
@@ -0,0 +1,69 @@
...
+
+const rateSumCombined = Math.round((mondayRateSum + tuesdayRateSum) * 100) / 100;
+console.log(`Total sum is €${rateSumCombined}`);
```
paulbremer: Your console.log outputs `Total sum is €333.33`, while the expected output is €373.33. Maartje is getting underpaid! Try to figure out where the error in the code is, hint; sometimes Maartje only works 2 hours.

PR #294, opened 2019-03-27
javascript2week2

alaashhadat:


Week2/homework/kalb3/maartjes-work.js
```diff
@@ -0,0 +1,68 @@
...
+
+const maartjesTasks = monday.concat(tuesday);
+const maartjesHourlyRate = 20;
+
+function computeEarnings(tasks, hourlyRate) {
+  tasks = tasks
+    .map(durations => durations.duration / 60)
+    .filter(durations => durations >= 2)
+    .map(durations => durations * hourlyRate)
+    .reduce((x, y) => x + y);
```
paulbremer: While using `x` and `y` works fine it easier for other developers to know what is going on if you use more descriptive names, `sum`, `taskDuration` for example.

Week2/homework/kalb3/maartjes-work.js
```diff
@@ -0,0 +1,68 @@
...
+  return tasks;
+}
+
+const earnings = computeEarnings(maartjesTasks, maartjesHourlyRate);
+let earn = Math.floor(earnings * 100) / 100.0;
+earn = new Intl.NumberFormat('nl-NL', { style: 'currency', currency: 'EUR' }).format(earnings);
```
paulbremer: Cool use of the `Intl.NumberFormat` function! You are not using the output from the line above (`let earn = Math.floor(earnings * 100) / 100.0;`), since you are using the `earnings` value here. You could skip that line and make this a `const`.

Week2/homework/kalb3/map-filter (2).js
```diff
@@ -0,0 +1,18 @@
+'use strict';
+
+/* cSpell:disable */
+
+function doubleOddNumbers(numbers) {
+  numbers = numbers.filter(number => number % 2 !== 0).map(number => number * 2);
```
paulbremer: Nice use of chaining! Keep in mind that your code should also still be readable, if one line gets to long it sometimes is better to do the second function on a new line.

PR #295, opened 2019-03-27
Week2

ghifo:


ghufran/map-filter.js
```diff
@@ -0,0 +1,15 @@
+'use strict';
+
+function doubleOddNumbers(numbers) {
+  const newNumbers = numbers.filter(number => number % 2 !== 0); // filter the odd nrs
+  numbers = newNumbers.map(number => number * 2); // double the odd nrs
+  console.log(numbers);
+}
+
+const myNumbers = [1, 2, 3, 4];
+console.log(doubleOddNumbers(myNumbers)); // [2, 6] the double of [1,3]
```
paulbremer: Nice! I like the comments in your code, makes it easier for other developers to know what is going on.

ghufran/maartjes-work.js
```diff
@@ -0,0 +1,78 @@
...
+/* cSpell: disable */
+const maartjesTasks = monday.concat(tuesday);
+const maartjesHourlyRate = 20;
+
+function computeEarnings(tasks, hourlyRate) {
+  const duration = tasks.map(task => {
+    // turn mins to hours
+    const hours = task.duration / 60;
+    return { ...task, duration: hours };
+  });
```
paulbremer: This code works fine but can be written more efficiently. The `map()` function by default will return a new array with each element being the result of the callback function.

`const duration = tasks.map(task => task.duration / 60);`

Take a look at the docs for some more info on `map()`, https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map

PR #302, opened 2019-03-29
saidcetinkaya JS2 Week2 HW

saidcetinkaya:


Week2/homework/maartjes-work.js
```diff
@@ -46,16 +46,17 @@ const maartjesTasks = monday.concat(tuesday);
 const maartjesHourlyRate = 20;
 
 function computeEarnings(tasks, hourlyRate) {
-  // Replace this comment and the next line with your code
-  console.log(tasks, hourlyRate);
+  const hourTasks = tasks.map(elem => elem.duration / 60);
+  const moreThan2Hour = hourTasks.filter(elem => (elem >= 2 ? elem : false));
```
paulbremer: You already are filtering things so it will only return an element if the element is equal or greater than 2 (`elem >= 2`). There is no need the return false because the filter function will just not return that element. 

`hourTasks.filter(elem => elem >= 2);`

Week2/homework/map-filter.js
```diff
@@ -1,8 +1,7 @@
 'use strict';
 
 function doubleOddNumbers(numbers) {
-  // Replace this comment and the next line with your code
-  console.log(numbers);
+  return numbers.filter(elem => (elem % 2 !== 0 ? elem : false)).map(elem => elem * 2);
```
paulbremer: Same here, take one more look at the docs to refresh your memory.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/filter

PR #305, opened 2019-04-03
Ramazan ince js2 hw3

RamazanInce:


ArcoMul:
Hi @rincemath 

Looking very good! All tests are passing successfully, including the bonus! :-)

I only have a comment about **2.7**, your explanation there does not describe correctly what's going on. Did you do a Google search for "javascript pass by value pass by reference" as advised in the MAKEME.md?

rincemath:
Hi @ArcoMul thanks for your feedback, i will check and get back to you again.

rincemath:
Hi @ArcoMul thanks for your feedback, i will check and get back to you again.
 
 image
 
 in pass value function value in the initial value will remain same and f2(y) will use the initial value and update the value for the each execution with this will value for f2(y) will change but not for f1(x).

PR #307, opened 2019-04-03
Muhannad js2 week3

Muhannad1989:


ArcoMul:
Hi @Muhannad1989 

Great work, all tests are passing smoothly! ⚡️ 

I noticed two things:
- Step 2-2 could be written more efficiently, you could reduce the if-else statement to two separate if-statements, which would result in fewer lines of code. Could you give it a try?
- And on Step 2-7 great Google find ;-) Stack Overflow always has good answers, could you try to explain in more detail though what exactly happens in the sample code? You answer was already correct regarding value vs. reference types, but I'd like to hear how it works in the context of this sample code.

Good luck!

PR #310, opened 2019-04-03
Mohammad darwesh3

mohammadnaefaldarwesh:


ArcoMul:
Hi @mohammadnaefaldarwesh 

Great work, all tests are passing now :)

Two things I noticed:

- Why do you use a setTimeout in Step 2-1?
- On step 2-7, great find, Stack Overflow always helps. Could you explain in more detail though what happens in the sample code? You answer is correct, but I'm mainly curious if you also understand how it works in the sample code. So describe step by step what happens in the code.

Good luck!

mohammadnaefaldarwesh:
> Hi @mohammadnaefaldarwesh
> 
> Great work, all tests are passing now :)
> 
> Two things I noticed:
> 
> * Why do you use a setTimeout in Step 2-1?
> * On step 2-7, great find, Stack Overflow always helps. Could you explain in more detail though what happens in the sample code? You answer is correct, but I'm mainly curious if you also understand how it works in the sample code. So describe step by step what happens in the code.
> 
> Good luck!

Hello @ArcoMul ,

Thank you for your feedback .
about setTimeOut i use it just for fun .

about 2-7:
In first function we are changing a  value of x and it is a number.
We are assigning new value.so it is pass by value.
In second function we are changing a property value of old object.
We are not assigning new object or array to old one.so it is pass by reference.
I hope that is enough to explain what happening.

Best regards,

PR #311, opened 2019-04-03
Orhan js2 week3 homework

OrhanGoc:


Week3/homework/step2-2.js
```diff
@@ -3,18 +3,26 @@
 function threeFive(startIndex, stopIndex, threeCallback, fiveCallback) {
   const numbers = [];
 
-  // Replace this comment and the next line with your code
-  console.log(startIndex, stopIndex, threeCallback, fiveCallback, numbers);
+  for (let i = startIndex; i <= stopIndex; i++) {
+    numbers.push(i);
+  }
+
+  numbers.forEach(number => {
```
paulbremer: Since you already use a for loop on line 6 you could move the if statements inside the for loop. The `i` corresponds to the number, you could do `if (i % 3 === 0) {}` inside the for loop. Then you don't need the `forEach` to loop over the array again.

Week3/homework/step2-7.js
```diff
@@ -20,4 +22,18 @@ f2(y);
...
+// different values in different locations in memory.
+// But, objects, arrays and functions pass by reference which means if we use
+// these variables in a function we address the same location.
+// So if we assign a new value to the variable, we also change the value of the variable.
+// In the first example variable x is a number and numbers are primitive
+// which passes by value. Since we have two locations in the memory for x, we get
+// 9 for the value of x outside the function, but in the function the value of x is 10.
+// In the second example y variable is an object and passes by reference. So
+// we have one location in the memory for y value. So we have the same result for y
+// in the function and outside the function.
```
paulbremer: Very detailed explanation 👍

OrhanGoc:
Hi @paulbremer, 
Thank you for your feedback. 
I reviewed assignment 2.2 as in your feedback and also added step 3 .

paulbremer:
Great job @OrhanGoc !

PR #312, opened 2019-04-03
Enes Week3 Homework with Week2 Homeworks

eneserdek:


Week3/homework/step2-2.js
```diff
@@ -2,19 +2,22 @@
...
+    return arr;
+  }
+  threeCallback(inner(numbers));
+  fiveCallback(inner(numbers));
 }
 
-function sayThree(number) {
-  // Replace this comment and the next line with your code
-  console.log(number);
+function sayThree(arr) {
```
paulbremer: So the function should actually use a number as argument, not the whole array.

Week3/homework/step2-4.js
```diff
@@ -2,6 +2,7 @@
 
 function Dog() {
   // add your code here
+  (this.name = "Da Kin'"), (this.color = 'blue'), (this.numLegs = 3);
```
paulbremer: Good, you could also write it like this;
```
this.name = "Da Kin'";
this.color = "blue";
this.numLegs = 3;
```

Looks a bit more readable in my opinion.

PR #313, opened 2019-04-04
saidcetinkaya JS2 Week3 Homework

saidcetinkaya:


Week3/homework/step2-2.js
```diff
@@ -3,18 +3,25 @@
 function threeFive(startIndex, stopIndex, threeCallback, fiveCallback) {
   const numbers = [];
 
-  // Replace this comment and the next line with your code
-  console.log(startIndex, stopIndex, threeCallback, fiveCallback, numbers);
+  for (let number = startIndex; number < stopIndex + 1; number++) {
+    numbers[number - startIndex] = number;
```
paulbremer: What does this line do?

saidcetinkaya: It’s creating index for array and push a value to respective index.

paulbremer: But you are not using that array (`numbers`) anymore so no need to push the value there.

saidcetinkaya: yes, you're right. I fixed it now. Thank you! :)

paulbremer: After some more reviewing I now understand that you were doing the following: 'The function should first generate an array containing values from start value to end value (inclusive).'

My mistake, sorry, the `numbers` does contain the correct numbers for the array! 🙈

saidcetinkaya: still, it was unnecessary 🙈

saidcetinkaya: FYI: and also I changed the 2.7

PR #314, opened 2019-04-04
Homework w3

ghifo:


Week3/homework/step2-7.js
```diff
@@ -20,4 +20,15 @@ f2(y);
...
+Changing the value of a variable never changes the underlying primitive or object. 
+However, changing a property of an object referenced by a variable does change the underlying object.
+*/
+
+/* 
+in my terms:
+the variable == x == is defined with (const). when the value of such a variable is a num or str, so it doesn't accept any changes.
+the variable == y == has object as a value. when the value of a variable is an arr or an obj, 
+so it accepts changes accessing it through keys for obj and [i] for arr.
+*/
```
paulbremer: I think you understand what the difference is, make sure you really know the difference between a copy by value and a copy by reference and which is used in which situation.

Week3/homework/step3-bonus.js
```diff
@@ -3,7 +3,7 @@
 const values = ['a', 'b', 'c', 'd', 'a', 'e', 'f', 'c'];
 
 function makeUnique(arr) {
-  // Replace this comment and the next line with your code
+  arr = [...new Set(arr)];
```
paulbremer: Nice!

Week3/homework/step2-2.js
```diff
@@ -1,23 +1,39 @@
...
+    numbers.push(i); // push all nrs between the startIndex & the stopIndex
+  }
 
-  // Replace this comment and the next line with your code
-  console.log(startIndex, stopIndex, threeCallback, fiveCallback, numbers);
+  for (let i = 0; i < numbers.length; i++) {
+    if (numbers[i] % 3 === 0) {
+      // apply a condition to check if the [i] is divided by 3
+      threeCallback(numbers[i]);
+    } else if (numbers[i] % 5 === 0) {
```
paulbremer: One small thing, because you chain the if statement now the function outputs '10 can be divided by 5.', but not the '10 can be divided by 3.'. You can easily solve this by not using the `} else if` but putting the `if` on a new line, to make it two separate if statements.

ghifo:
OK, I'll do it.
Thank you

PR #315, opened 2019-04-04
homework week3 js2

Nadoshka:


Week3/homework/step2-4.js
```diff
@@ -1,7 +1,9 @@
 'use strict';
 
 function Dog() {
-  // add your code here
+  this.name = 'dog';
```
paulbremer: The dogs name is dog? 😛

Week3/homework/step2-5.js
```diff
@@ -3,10 +3,13 @@
 function multiplyAll(arr) {
   // eslint-disable-next-line
   let product = 1;
-
-  // Replace this comment and the next line with your code
-  console.log(arr, product);
-
+  let i = 0;
+  let j = 0;
+  for (i = 0; i < arr.length; i++) {
```
paulbremer: You could just add the `let` here and skip the lines above. 

`for (let i = 0; i < ...)`

Week3/homework/step2-6.js
```diff
@@ -4,18 +4,19 @@ const arr2d = [[1, 2], [3, 4], [5, 6]];
 const arr3d = [[[1, 2], [3, 4]], [[5, 6], [7, 8]]];
 
 function flattenArray2d(arr) {
-  // Replace this comment and the next line with your code
-  console.log(arr);
+  return arr.flat();
```
paulbremer: Very smart to look for functions that do all the work 👍 But this function won't work on Internet Explorer and Edge, can you find a way to do this yourself and that works in all browsers?

Week3/homework/step2-7.js
```diff
@@ -20,4 +24,9 @@ f2(y);
 
 console.log(y);
 
-// Add your explanation as a comment here
+/* the value of the object is stored in a reference on memory 
+so when we give a new value to the same object, 
+the new value would be stored on the same reference
+(the same reference of the original object)
+so the original value of the object will change 
+so F2(Y) === Y */
```
paulbremer: Looks like you understand the difference, one is copy by value and the ofter is a copy by reference.

Nadoshka: Yes, that what I meant! 💯 Thanks for making shorter! :) 👍

Week3/homework/step2-6.js
Nadoshka: Thanks for your feedback! I will try to do it in another way! :) 👍

Week3/homework/step2-5.js
Nadoshka: Thanks! you are right!  I will update it 👍 :)

Week3/homework/step2-4.js
Nadoshka: hahaha it can be Poodle ;)
Thanks 👍

PR #316, opened 2019-04-06
Akif js2 w3

Akif2709:


Week3/homework/step2-3.js
```diff
@@ -4,10 +4,13 @@
...
   let result = '';
-
-  // Replace this comment and the next line with your code
-  console.log(str, num, result);
-
+  if (num > 0) {
+    for (let i = 0; i < num; i++) {
+      result += str;
+    }
+  } else {
```
paulbremer: You are not really returning anything useful so you could just skip the `else` here.

Week3/homework/step2-7.js
```diff
@@ -21,3 +21,11 @@ f2(y);
 console.log(y);
 
 // Add your explanation as a comment here
+/* Primitive variables can not be changed by function, but array or object can be changed. Because when we write array or object
+as a argument, it is approached with its reference/address. And when we change object in  function, we are changing 
+the referenced/original object. Dont ask why, it is Javascript : )
```
paulbremer: But we are asking why! I think you get the general idea, try to explain a bit more by using the code from the task.

Akif2709: I added more explanation using codes as much as possible. Thanks for review Paul

paulbremer: 👍

PR #326, opened 2019-05-19
JS2/Week1/Shvan

Shvan1:


app.component.js
```diff
@@ -0,0 +1,137 @@
...
+  'arms_and_the_man',
+  'merchant_of_venice',
+  'crime_and_punishment',
+  'romeo_and_juliet',
+  'a_tale_of_two_cities',
+];
+//console.log(bookIds);
+
+//3.
+function books(bookIds) {
```
NoerGitKat: It's better to give a function a more 'active' name, that includes a verb. This is because a function is a tool you're using to get a certain job done. In this case, you could call it "displayBookList" or "renderBookTitles".

app.component.js
```diff
@@ -0,0 +1,137 @@
...
+  for (let index = 0; index < bookIds.length; index++) {
+    const li = document.createElement('li');
+    (li.textContent = 'List of Books:'), bookIds[index];
+    ul.appendChild(li);
+  }
+}
+window.addEventListener('load', books);
+
+//4.
+const DifferentBooks = {
```
NoerGitKat: The name of this variable can be more meaningful, like "bookDetails".

app.component.js
```diff
@@ -0,0 +1,137 @@
...
+  the_merchant_of_venice: './img/MerchantOfVenice.jpg',
+  crime_and_punishment: './img/CrimeAndPunishment.jpg',
+  romeo_and_juliet: './img/RomeoAndJuliet.jpg',
+  a_tale_of_two_cities: './img/a-tale-of-two-cities.jpg',
+};
+
+//8.
+function covers() {
+  const div = document.getElementById('items');
+  const div2 = document.createElement('h1');
```
NoerGitKat: Same advice here: naming variables and functions to be meaningful is **very** important. This is because you'll be reading your code over and over again, so a meaningful name helps you much quicker understand what's happening.

app.component.js
```diff
@@ -0,0 +1,137 @@
...
+  div2.innerHTML = 'My Favorite Books';
+  div.appendChild(div2);
+  div2.className = 'moreItems';
+
+  const divTag = document.createElement('div');
+  divTag.className = 'my-books';
+  const ul = document.createElement('ul');
+  ul.className = 'books';
+  bookIds.forEach(function(index) {
```
NoerGitKat: For consistency you could've used a for-loop again. It might look different, but both constructs (a for-loop and forEach) essentially do the same.

PR #327, opened 2019-05-20
7alip - Homework 1

7alip:


ArcoMul:
Hey Talip!

Really good work! I'm impressed! You are ready for the future with the light/dark switch 😁 

Two things which I noticed:
- You're using all capitals for some functions name, namely `CREATE_ELEMENT` and `ADD_CHILD`, in Javascript there is the convention to write function names in camelCase, meaning that function and variable names almost always start with a lowercase letter and as soon as you hit another word you capitalize the first letter. So in your case it would be: `createElement` and `addChild`
- In the homework is requested to add an `alt` attribute to your images. Read more about the alt attribute here: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img#Attributes It's important to define the alt attribute since some people (visually impaired people for example) need this attribute to navigate the web.

Besides these two things, very well done!

Arco

7alip:
> Hey Talip!
> 
> Really good work! I'm impressed! You are ready for the future with the light/dark switch 😁
> 
> Two things which I noticed:
> 
> * You're using all capitals for some functions name, namely `CREATE_ELEMENT` and `ADD_CHILD`, in Javascript there is the convention to write function names in camelCase, meaning that function and variable names almost always start with a lowercase letter and as soon as you hit another word you capitalize the first letter. So in your case it would be: `createElement` and `addChild`
> * In the homework is requested to add an `alt` attribute to your images. Read more about the alt attribute here: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img#Attributes It's important to define the alt attribute since some people (visually impaired people for example) need this attribute to navigate the web.
> 
> Besides these two things, very well done!
> 
> Arco

Thanks for reviewing my code. 
> * I knew it's really important to add `alt` attribute to `img` element but I just forgot to set it. 
> * I corrected the naming convention and added more helper functions. It might make the code more confused to force many custom functions but I just wanted to play with functions :)

ArcoMul:
Super Talip! Nice helpers functions as well 👏

fedefu:
@7alip Great job man!

PR #329, opened 2019-05-20
Suheyb - Week1 Homework

suh3yb:


ArcoMul:
Hi @suh3yb,

Very nice work! Clear code and everything works as expected :-) I like how you made the menu and that the page jumps to the right book when clicking it 👌 

Your homework is completely fine, just two small things you could improve upon:
- You current get the `navigator` element using `getElementsByClassName`, which returns an array. Then in your code you use `navigator[0]`, in my opinion it would be better to straight away take the first item of the array when getting the element, so: `const navigator = document.getElementsByClassName('navigator')[0];`
- As an extra challenge, it would be nice if the book titles instead of the book ids would show up in the menu. Try it if you find some extra time ;-)

See you tomorrow!

suh3yb:
Hi @ArcoMul, 

Thank you very much for your time and feedback!

I have made some improvements as you suggested and changed navigator's texts as book titles.

Best regards

PR #337, opened 2019-05-23
JS2 homework week1

alquinex:


Week1/homework/app.js
```diff
@@ -1,11 +1,133 @@
...
+    },
+    installing_and_configuring_windows_server_2012: {
+      title: 'installing_and_configuring_windows_server_2012',
+      language: 'English',
+      author: 'Craig Zacker',
+    },
+  };
+
+  // Task 1.3 - 1.5
+  function actualBooksInfo() {
```
NoerGitKat: For the naming of functions it's better to use a verb, to show it's an action. In this case you could rename it into "renderBookList" or "displayBookDetails".

Week1/homework/app.js
```diff
@@ -1,11 +1,133 @@
...
+    things_fall_apart: './img/things_fall_apart.jpg',
+    chike_and_the_river: './img/chike_and_the_river.jpg',
+    a_man_for_all_seasons: './img/a_man_for_all_seasons.jpg',
+    installing_and_configuring_windows_10: './img/installing_and_configuring_windows_10.jpg',
+    installing_and_configuring_windows_server_2012:
+      './img/installing_and_configuring_windows_server_2012.jpg',
+  };
+
+  // 1.7 - 1.8
+  function displayBooks() {
```
NoerGitKat: The naming is great in this case! Can you see why?

Week1/homework/app.js
```diff
@@ -1,11 +1,133 @@
...
+    for (const key of Object.keys(bookCovers)) {
```
NoerGitKat: It would be cleaner (in other words, more readable) if you put `Object.keys(bookCovers)` into a separate variable. This way you'll separate the two operations, thus creating more clarity.

alquinex: Hi Noer, 
Thanks for the feedback. 
I was actually going to improve on the homework again if I have the time.
But I have made the required changes.

PR #338, opened 2019-05-24
week1-homework

irfankaradeniz:


NoerGitKat:
Hey @wilgert and @ArcoMul ! Could anyone of you give Irfan a review of his week 1 homework?

ArcoMul:
Hi @irfankaradeniz 

Good job on the homework! Some things I noted:

- You created a file called "DOM property.url" in the homework folder, you don't use it though, and it is not clear to me what it's use is. Better to delete it if it is not used.
- Currently you add CSS styling to your HTML elements in Javascript. It's better to do this in a CSS file. There is already a style.css where you can fill in the CSS, you just have to link in the HTML to get it to work.

Besides that it good code, clean and readable. Could you fix those two small points though?

Thanks!

PR #340, opened 2019-05-24
Rename bookTitles to bookIDs

wilgert:
This will hopefully make it more clear that the entries in this array should be used as the keys in the object.

### DIFF
--- a/Week1/homework/app.js
+++ b/Week1/homework/app.js
@@ -1,11 +1,11 @@
 'use strict';
 
 {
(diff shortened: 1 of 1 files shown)


wilgert:
The problem with the current name is that it does not match up with the text in MAKEME.md.
For example in point 1.4 of MAKEME.md:

> 1.4 ... Each property of this object should be another (i.e., nested) object with the **book ID** you thought up in step 1.1 as a key ...

PR #343, opened 2019-05-26
Murad-Week2 Homework

Hdvndgr:


Week2/homework/maartjes-work.js
```diff
@@ -46,16 +46,24 @@ const maartjesTasks = monday.concat(tuesday);
 const maartjesHourlyRate = 20;
 
 function computeEarnings(tasks, hourlyRate) {
-  // Replace this comment and the next line with your code
+  const result = tasks
+    .map(item => item.duration / 60)
```
wilgert: I think I used `item` as the name of a variable when doing `map/filter/reduce` during the lesson as well. Nevertheless, I think the name `task` would be more clear in this case!

Week2/homework/maartjes-work.js
```diff
@@ -46,16 +46,24 @@ const maartjesTasks = monday.concat(tuesday);
...
 // eslint-disable-next-line no-unused-vars
 const earnings = computeEarnings(maartjesTasks, maartjesHourlyRate);
 
 // add code to convert `earnings` to a string rounded to two decimals (euro cents)
-
-console.log(`Maartje has earned €${'replace this string with the earnings rounded to euro cents'}`);
+const roundedEarnings = earnings
+  .toFixed(2)
+  .toString()
+  .replace(/(\d)(?=(\d{3})+(?!\d))/g, '$1.');
```
wilgert: What is this line doing?

Week2/homework/maartjes-work.js
```diff
@@ -46,16 +46,24 @@ const maartjesTasks = monday.concat(tuesday);
...
 
 // eslint-disable-next-line no-unused-vars
 const earnings = computeEarnings(maartjesTasks, maartjesHourlyRate);
 
 // add code to convert `earnings` to a string rounded to two decimals (euro cents)
-
-console.log(`Maartje has earned €${'replace this string with the earnings rounded to euro cents'}`);
+const roundedEarnings = earnings
+  .toFixed(2)
+  .toString()
```
wilgert: One more thing: `.toFixed()` already returns a string so you don't need the `.toString()` here. See: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/toFixed

Week2/homework/maartjes-work.js
Hdvndgr: This line is formatting the number as currency. To do that I wrote "Currency Formating in javascript" and found this website : https://blog.abelotech.com/posts/number-currency-formatting-javascript/
But i replaced '$1,' with '$1.' But this maybe make the code confusing.

wilgert: Well it is not needed because line 63 already rounds it to 2 decimals and line 66 already puts a € before it.

PR #349, opened 2019-05-27
run tests for week2 automatically

wilgert:
so we don't need to checkout the Pull Requests locally to verify if the code works, but instead have travis do that work.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,3 +4,4 @@ node_js:
 
 script:
   - npm run lint
+  - npm run test-week2


wilgert:
Interesting catch22 here...

wilgert:
Yes I figured that. Maybe we need to create separate repositories per week, so we can really use unit tests and travis like they're supposed to.

remarcmij:
I should mention that @NoerGitKat and @wouterkleijn have pushed back a bit on using unit tests at this stage as they report seeing students becoming confused by the extra tooling. Let's see how this class is doing.

NoerGitKat:
This is what we have observed so far. I must say, however, that increasingly I have come to understand that students are confused by the homework in general (irrespective of the unit tests). I hope rewriting the curriculum will clear this up more. I'll keep you all up to date about that!

PR #352, opened 2019-05-28
zehra--homework week 2

zehrakopar:
There are some warnings because the spelling of 'maartjes'.

### DIFF
--- a/Week2/homework/maartjes-work.js
+++ b/Week2/homework/maartjes-work.js
@@ -46,16 +46,19 @@ const maartjesTasks = monday.concat(tuesday);
 const maartjesHourlyRate = 20;
 
 function computeEarnings(tasks, hourlyRate) {
-  // Replace this comment and the next line with your code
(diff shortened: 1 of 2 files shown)


Week2/homework/maartjes-work.js
```diff
@@ -46,16 +46,19 @@ const maartjesTasks = monday.concat(tuesday);
 const maartjesHourlyRate = 20;
 
 function computeEarnings(tasks, hourlyRate) {
-  // Replace this comment and the next line with your code
-  console.log(tasks, hourlyRate);
+  return tasks
+    .filter(task => task.duration >= 120) // when i do the filter after the map i get an empty array
```
wilgert: You could do the filter after the map. But you have to compare with a different number. Can you think of what number that should be?

zehrakopar: yes it will be 2. because of the hourly expression. I try it but i get an empty array

PR #353, opened 2019-05-28
Veli_JavaScript2_Week2

veliataseven:


Week2/homework/maartjes-work.js
```diff
@@ -46,16 +46,19 @@ const maartjesTasks = monday.concat(tuesday);
 const maartjesHourlyRate = 20;
 
 function computeEarnings(tasks, hourlyRate) {
-  // Replace this comment and the next line with your code
-  console.log(tasks, hourlyRate);
+  return tasks
+    .map(task => task.duration / 60)
```
wilgert: In this case it does not really matter. But in general it is advisable to do `.filter` before `.map`. Unless you really need the result of te map to do the filtering. Otherwise your code is wasting time mapping items it will filter out later.

Week2/homework/maartjes-work.js
```diff
@@ -46,16 +46,19 @@ const maartjesTasks = monday.concat(tuesday);
...
+    .filter(hour => hour >= 2)
+    .reduce((sum, hour) => sum + hour * hourlyRate, 0);
 }
 
 // eslint-disable-next-line no-unused-vars
 const earnings = computeEarnings(maartjesTasks, maartjesHourlyRate);
 
 // add code to convert `earnings` to a string rounded to two decimals (euro cents)
+const stringTypeOfEarning = earnings.toFixed(2).toString();
```
wilgert: `.toString()` is not needed here `.toFixed()` already produces a string.
Although the name `stringTypeOfEarning` is clear it's also a bit longer than necessary. I would call it something like `formattedEarnings`.

Week2/homework/maartjes-work.js
veliataseven: Okay Got it. I did these according to instructions order.

Week2/homework/maartjes-work.js
veliataseven: Okay thanks for feedback. I didn't know that .toFixed already produces a string. And I thought that in the homework  We need to convert specified result to string.

fedefu:
Great job @veliataseven !

PR #358, opened 2019-05-30
irfan - week2 - test is passed

irfankaradeniz:


Week2/homework/maartjes-work.js
```diff
@@ -46,8 +46,10 @@ const maartjesTasks = monday.concat(tuesday);
 const maartjesHourlyRate = 20;
 
 function computeEarnings(tasks, hourlyRate) {
-  // Replace this comment and the next line with your code
-  console.log(tasks, hourlyRate);
+  return tasks
+    .map(durations => durations.duration)
```
hasan-sh: `Multiply each duration by a per-hour rate for billing (use €20/hour) and sum it all up.` => first convert the duration to hour rate!
It's not wrong, but doing so makes it more clear!

Week2/homework/map-filter.js
```diff
@@ -1,8 +1,7 @@
 'use strict';
 
 function doubleOddNumbers(numbers) {
-  // Replace this comment and the next line with your code
-  console.log(numbers);
+  return numbers.filter(number => number % 2 !== 0).map(number => number * 2);
```
hasan-sh: Is the name `number` in the map method the best name you could give??

>Think about what numbers you're filtering out in the first place!

NoerGitKat:
Thanks for giving feedback @hasan-sh ! Hey @irfankaradeniz , could you take a look and implement the changes?

Week2/homework/maartjes-work.js
irfankaradeniz: I got that, thanks!

Week2/homework/map-filter.js
irfankaradeniz: I think it's not. Should it be oddNumber?

PR #359, opened 2019-05-30
Hanijs2w2

hanialkudsi:


Week2/homework/maartjes-work.js
```diff
@@ -46,16 +46,18 @@ const maartjesTasks = monday.concat(tuesday);
 const maartjesHourlyRate = 20;
 
 function computeEarnings(tasks, hourlyRate) {
-  // Replace this comment and the next line with your code
-  console.log(tasks, hourlyRate);
-}
+  const hours = tasks.map(task => task.duration / 60);
```
hasan-sh: Very nice namings :100:

Week2/homework/map-filter.js
```diff
@@ -1,8 +1,9 @@
 'use strict';
 
 function doubleOddNumbers(numbers) {
-  // Replace this comment and the next line with your code
-  console.log(numbers);
+  const doubleOddNumber = numbers.filter(number => number % 2 !== 0).map(number => number * 2);
```
hasan-sh: Is the name `number` in the `map` method the best name you could give??

>Think about what numbers you're filtering out in the first place!

PR #362, opened 2019-06-03
7alip | Homework3

7alip:


Week3/homework/step2-3.js
```diff
@@ -30,9 +33,12 @@ console.log('while', repeatStringNumTimesWithWhile('abc', 3));
 function repeatStringNumTimesWithDoWhile(str, num) {
   // eslint-disable-next-line prefer-const
   let result = '';
+  let i = 0;
 
-  // Replace this comment and the next line with your code
-  console.log(str, num, result);
+  do {
+    result = num <= 0 ? result : result + str;
```
remarcmij: You are correct in that you need to test the value of `num` as a `do-while` loop only tests the loop condition at the end of the loop. However, since the value of `num` does not change inside the loop (it is 'loop invariant') it is better and more efficient to test the value of `num` once only before entering the loop and execute the loop only if `num > 0`:

```js
if (num > 0) {
  let i = 0;
  do {
    result += str;
    i++;
  } while (i < num);
}
```

Week3/homework/step2-4.js
```diff
@@ -1,7 +1,9 @@
 'use strict';
 
 function Dog() {
-  // add your code here
+  this.name = '';
+  this.color = '';
+  this.numLegs = 4;
 }
 
 const hound = new Dog();
```
remarcmij: 👍

Week3/homework/step2-5.js
```diff
@@ -4,8 +4,11 @@ function multiplyAll(arr) {
   // eslint-disable-next-line
   let product = 1;
 
-  // Replace this comment and the next line with your code
-  console.log(arr, product);
+  for (let i = 0; i < arr.length; i++) {
+    for (let j = 0; j < arr[i].length; j++) {
+      product *= arr[i][j];
```
remarcmij: Nice use of the `*=` short-hand. 👍

Week3/homework/step2-7.js
```diff
@@ -20,4 +20,28 @@ f2(y);
...
+ * //////////////////////////////
+ */
+
+/**
+ * Reference types (object/array) are different from primitive types.
+ * Each object/array has a reference address in disk. If we reassign an object, its address will be changed.
+ * Then it will never be reassigned and its address will never be changed.
+ * But if we change a property of an object, then the object will be manipulated. But the address still will be the same.
+ * No matter in which scope we change the property, we get the manipulated object.
+ * Even if we assign our object to a new object, changing the property of the new object will effect the original object.
```
remarcmij: Overall, I think you have a good understanding of the concepts of call by value and call by reference, but I'm not sure if I understand this line: 'Even if we...'

Week3/homework/step3.js
```diff
@@ -1,8 +1,7 @@
 'use strict';
 
 function createBase(base) {
-  // Replace this comment and the next line with your code
-  console.log(base);
+  return addition => base + addition;
```
remarcmij: Nice use of an arrow function. As we saw in class, sometimes you this as:

```js
const createBase = base => addition => base + addition;
```

That takes some getting used to.
```

File: Week3/homework/step3-bonus.js [hunk 6]
```diff
@@ -3,8 +3,15 @@
 const values = ['a', 'b', 'c', 'd', 'a', 'e', 'f', 'c'];
 
 function makeUnique(arr) {
-  // Replace this comment and the next line with your code
-  console.log(arr);
+  const uniqueObject = {};
```
remarcmij: Here you are exploiting the fact that object keys are unique. You are just setting the property value to an empty string, but have no actual use for the property value. That is fine. Alternatively, you could use an ES6 Set, the members of which are also unique but have no associated value part.

fedefu:
Thanks for all the constructive feedback @remarcmij! Super clear right @7alip ?

Week1/homework/app.js
```diff
@@ -1,11 +1,191 @@
+/**
```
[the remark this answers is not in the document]
7alip: Sorry, it was because of my inattention. I will never do the same mistake again, I hoop.
Thank you!

Week3/homework/step2-1.js
```diff
@@ -2,8 +2,7 @@
 
 function foo(func) {
   // What to do here?
-  // Replace this comment and the next line with your code
-  console.log(func);
+  return func();
```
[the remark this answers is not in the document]
7alip: I was stuck it in my mind that closures basicly mean even if the outer function returns, the inner functions will still use the variables which are created in outer function. But it's not the single explanation for closures that the inner function must be returned. Now I know better thanks to you, especially with ovChipCard and wallet examples.

Week3/homework/step2-2.js
```diff
@@ -3,18 +3,31 @@
...
+
+  numbers.map(number => {
+    if (!(number % 3) && number % 15) {
+      threeCallback(number);
+    } else if (!(number % 5) && number % 15) {
+      fiveCallback(number);
+    } else if (!(number % 15)) {
+      threeCallback(number);
+      fiveCallback(number);
+    }
```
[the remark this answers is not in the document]
7alip: For these kind of if statements I was never thinking this approach which is seperated if statements. I got the point in the class while you were explaining this. I will try to think about it when I solve another algorithms problems.

Besides, thanks you so much for these impressive comments.

Week3/homework/step2-7.js
7alip: > but I'm not sure if I understand this line: 'Even if we...'

I tried to explain this:

```javascript
  const objX = { name: "Frank" }
  cons objY = objX;

   objY.age = 20;
  console.log(objX); // { name: "Frank", age: 20 }
```

Week3/homework/step3-bonus.js
7alip: I need to solve more algorithms to acquire alternative solution skills. After I google for some solutions, it can be forgotten if I don't try to solve without looking at the solutions.

PR #363, opened 2019-06-03
hamza js2-week3 homework

lizozo:
hamza js2-week3 homework

### DIFF
--- a/Week3/homework/step2-5.js
+++ b/Week3/homework/step2-5.js
@@ -4,9 +4,11 @@ function multiplyAll(arr) {
   // eslint-disable-next-line
   let product = 1;
 
-  // Replace this comment and the next line with your code
-  console.log(arr, product);
-
+  for (let i = 0; i < arr.length; i++) {
+    for (let j = 0; j < arr[i].length; j++) {
+      product *= arr[i][j];
+    }
+  }
   return product;
 }
 
--- a/Week3/homework/step2-6.js
+++ b/Week3/homework/step2-6.js
@@ -4,19 +4,20 @@ const arr2d = [[1, 2], [3, 4], [5, 6]];
 const arr3d = [[[1, 2], [3, 4]], [[5, 6], [7, 8]]];
 
 function flattenArray2d(arr) {
-  // Replace this comment and the next line with your code
-  console.log(arr);
+  while (arr.find(item => typeof item === 'object') !== undefined) arr = [].concat(...arr);
+  return arr;
 }
 
 function flattenArray3d(arr) {
-  // Replace this comment and the next line with your code
-  console.log(arr);
+  while (arr.find(item => typeof item === 'object') !== undefined) arr = [].concat(...arr);
+  return arr;
 }
 
 console.log(flattenArray2d(arr2d)); // -> [1, 2, 3, 4, 5, 6]
 console.log(flattenArray3d(arr3d)); // -> [1, 2, 3, 4, 5, 6, 7, 8]
 
-// Do not change or remove anything below this line
+/* I just wrote one function for flattening those two arrays because it works for both arrays. Moreover as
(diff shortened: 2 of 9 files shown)


Week3/homework/step2-5.js
```diff
@@ -4,9 +4,11 @@ function multiplyAll(arr) {
...
-  // Replace this comment and the next line with your code
-  console.log(arr, product);
-
+  for (let i = 0; i < arr.length; i++) {
+    for (let j = 0; j < arr[i].length; j++) {
+      product *= arr[i][j];
+    }
+  }
   return product;
 }
```
remarcmij: 👍

Week3/homework/step2-6.js
```diff
@@ -4,19 +4,20 @@ const arr2d = [[1, 2], [3, 4], [5, 6]];
 const arr3d = [[[1, 2], [3, 4]], [[5, 6], [7, 8]]];
 
 function flattenArray2d(arr) {
-  // Replace this comment and the next line with your code
-  console.log(arr);
+  while (arr.find(item => typeof item === 'object') !== undefined) arr = [].concat(...arr);
```
remarcmij: While your code is clever and passes the test, the solution that you have chosen is fragile. It works for the array provided, but if I add an object literal as an array element your code loops forever:

```js
const arr2d = [[1, 2], [3, 4], [5, 6], {}];
```

This is because an array is an object, for sure, but so is an actual object. If you want to test whether a value is an array (and not just any object) you should use `Array.isArray()`:

```js
while (arr.find(item => Array.isArray(item))) arr = [].concat(...arr);
```

Week3/homework/step2-7.js
```diff
@@ -20,4 +20,8 @@ f2(y);
 
 console.log(y);
 
-// Add your explanation as a comment here
+/* In first function we only copy the value of our"x" variable to the "val". 
+they do not point to same memory adress. In second function, since our variable 
+is an object, instead of copying its value to "val" variable, it assigns the same 
+memory adress to the "val" variable. so whenever we change the value of an object,
+other objects which points to same adress also changes. */
```
remarcmij: I'm completely satisfied with your explanation. 👍

Week3/homework/step3-bonus.js
```diff
@@ -3,8 +3,8 @@
 const values = ['a', 'b', 'c', 'd', 'a', 'e', 'f', 'c'];
 
 function makeUnique(arr) {
-  // Replace this comment and the next line with your code
-  console.log(arr);
+  const set1 = new Set(arr);
+  return [...set1];
```
remarcmij: Nice! We haven't covered the spread operator, but you have put it to good use here. Alternatively, you could use the `Array.from()` method:

```js
function makeUnique(arr) {
  return Array.from(new Set(arr));
}
```

Week3/homework/step3.js
```diff
@@ -1,8 +1,9 @@
 'use strict';
 
 function createBase(base) {
-  // Replace this comment and the next line with your code
-  console.log(base);
+  return function(number) {
+    return number + base;
+  };
 }
```
remarcmij: 👍

lizozo:
> 
> 
> Hi @lizozo , pretty good work here. Please look at the comments for step2-2 specifically. But I'm happy enough to approve your pull request. Good luck with the test!

Hi @remarcmij !!! First of all I would like thank you for your valuable comments and feedback. They are very helpful and insightful. I can easily say that they contribute a lot to my learning. as students of HFY, we are very lucky to have you.

When it comes to 2-2, I think i over concentrated on passing the test. A text in test 2.2 says "15 divisible by 3 and 5"  and it got me a little bit confused.  But i totally grasp the idea thanks to your comments.

thank you very much again.

PR #364, opened 2019-06-04
irfan-week3-homework

irfankaradeniz:


Week3/homework/step2-1.js
```diff
@@ -4,6 +4,7 @@ function foo(func) {
   // What to do here?
   // Replace this comment and the next line with your code
   console.log(func);
```
remarcmij: This `console.log` was supposed to be removed.

Week3/homework/step2-1.js
```diff
@@ -4,6 +4,7 @@ function foo(func) {
...
+  return func();
```
remarcmij: You are right that you should call `func` (which is, in this case, is a reference to `bar`. It is not necessary to return the return value of `func/bar` (``bar` doesn't return anything and therefore implicitly returns `undefined`):

```js

function foo(func) {
  func();
}
```

Week3/homework/step2-2.js
```diff
@@ -2,19 +2,34 @@
 
 function threeFive(startIndex, stopIndex, threeCallback, fiveCallback) {
   const numbers = [];
-
   // Replace this comment and the next line with your code
   console.log(startIndex, stopIndex, threeCallback, fiveCallback, numbers);
```
remarcmij: This `console.log` was supposed to be deleted.

Week3/homework/step2-2.js
```diff
@@ -2,19 +2,34 @@
...
+  for (let i = startIndex; i <= stopIndex; i++) {
+    numbers.push(i);
+  }
+  numbers.map(number => {
```
remarcmij: The `.map()` array method returns a new array of 'mapped' or transformed values of the original array. But here you are not interested in creating a new array, you just want to iterate through the elements of the array, in which case a `.forEach()` or a traditional `for` loop is more appropriate. While `.map()` works here, it will raise eyebrows from other developers (including me) trying to figure out what you are 'mapping' here (nothing).

Week3/homework/step2-2.js
```diff
@@ -2,19 +2,34 @@
...
+      threeCallback(number);
+    } else if (number % 5 === 0 && !(number % 3 === 0)) {
+      fiveCallback(number);
+    } else if (number % 3 === 0 && number % 5 === 0) {
+      threeCallback(number);
+      fiveCallback(number);
+    }
+    return number;
+  });
 }
```
remarcmij: You could simplify your `if-then-else` block by changing the order a bit:

```js
if (number % 3 === 0 && number % 5 === 0) {
  threeCallback(number);
  fiveCallback(number);
} else if (number % 3 === 0) {
  threeCallback(number);
} else if (number % 5 === 0) {
  fiveCallback(number);
}
```
If you test for the combined condition first, the second and the third conditions can be simpler.

However, still simpler and preferred is using two independent `if` statements, like this:

```js
numbers.forEach(number => {
  if (number % 3 === 0) {
    threeCallback(number);
  }
  if (number % 5 === 0) {
    fiveCallback(number);
  }
});
```

If you use `.forEach` instead of `.map` you can also get rid of the `return` statement without getting an ESLint error.

Week3/homework/step2-2.js
```diff
@@ -2,19 +2,34 @@
...
 
 function sayThree(number) {
   // Replace this comment and the next line with your code
   console.log(number);
+  return number + ' is divisible by three';
```
remarcmij: There is not much point in returning this string concatenation, as there is no caller that uses this value. It would be far better to `console.log` this. Then, instead of this rather uninformative output:

```
10
12
15
15
```

we would get the very informative output:

```
10 is divisible by five
12 is divisible by three
15 is divisible by three
15 is divisible by five
```

```js
function sayThree(number) {
  console.log(number + ' is divisible by three');
}

function sayFive(number) {
  console.log(number + ' is divisible by five');
}
```

Week3/homework/step2-3.js
```diff
@@ -4,10 +4,11 @@
 function repeatStringNumTimesWithFor(str, num) {
   // eslint-disable-next-line prefer-const
   let result = '';
-
   // Replace this comment and the next line with your code
   console.log(str, num, result);
```
remarcmij: This `console.log` and similar ones in the rest of this step were supposed to be deleted.

Week3/homework/step2-3.js
```diff
@@ -30,9 +34,13 @@ console.log('while', repeatStringNumTimesWithWhile('abc', 3));
 function repeatStringNumTimesWithDoWhile(str, num) {
   // eslint-disable-next-line prefer-const
   let result = '';
-
+  let i = 0;
   // Replace this comment and the next line with your code
   console.log(str, num, result);
+  do {
+    result = num <= 0 ? result : result + str;
```
remarcmij: You are correct that you must test for `num <= 0`,  however, you need not do this for every loop iteration. It is best and more efficient to just test it once, before entering the `do-while` loop and just skip the loop altogether if the loop condition is not met.

```js
if (num > 0) {
  do {
    result += str;
    i++;
  } while (i < num);
}
```

Week3/homework/step2-5.js
```diff
@@ -6,6 +6,12 @@ function multiplyAll(arr) {
 
   // Replace this comment and the next line with your code
   console.log(arr, product);
+  for (let i = 0; i < arr.length; i++) {
+    for (let j = 0; j < arr[i].length; j++) {
+      console.log(arr[i][j]);
```
remarcmij: We don't need all these `console.log` statements in a pull request. It's fine when you are debugging your code, but once your code is ready you should remove them before sending in the pull request. Note that a pull request is a request to another developer to review your code. The console.log's are of no interest to the reviewer; they are just unnecessary noise (same applies to commented-out code, but so far, I haven't seen that in your pull request).

Week3/homework/step2-6.js
```diff
@@ -6,11 +6,13 @@ const arr3d = [[[1, 2], [3, 4]], [[5, 6], [7, 8]]];
 function flattenArray2d(arr) {
   // Replace this comment and the next line with your code
   console.log(arr);
+  return arr.flat([2]);
```
remarcmij: This `.flat` array method is rather new and not universally supported by all browsers. Can you revise this code without using `.flat`?

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flat#Browser_compatibility

Week3/homework/step2-7.js
```diff
@@ -21,3 +21,7 @@ f2(y);
 console.log(y);
 
 // Add your explanation as a comment here
+// Pass by reference is when variables passed in to functions are given the direct memory address.
+// This allows the function to manipulate the object or primitive as it exists outside the scope of the function.
+// Pass by value means that when you pass that variable into a function its the equivalent of creating a new variable and making a copy of the passed in variable.
+// They’re not necessarily the same exact variable in the same exact memory address but rather copies. Modifying the copy does not affect the original.
```
remarcmij: Yes, that seems about right.

Week3/homework/step3-bonus.js
```diff
@@ -5,6 +5,7 @@ const values = ['a', 'b', 'c', 'd', 'a', 'e', 'f', 'c'];
 function makeUnique(arr) {
   // Replace this comment and the next line with your code
   console.log(arr);
+  return arr.filter((v, i) => arr.indexOf(v) === i);
```
remarcmij: Yes, that works. Again, the console.log can go.

Week3/homework/step3.js
```diff
@@ -3,6 +3,7 @@
 function createBase(base) {
   // Replace this comment and the next line with your code
   console.log(base);
+  return x => base + x;
 }
 
 const addSix = createBase(6);
```
remarcmij: 👍 (except for the console.log)

fedefu:
All clear @irfankaradeniz ? Thanks for the feedback Jim!

PR #366, opened 2019-06-04
Suheyb - Week3

suh3yb:


Week3/homework/step2-1.js
```diff
@@ -1,9 +1,7 @@
 'use strict';
 
 function foo(func) {
-  // What to do here?
-  // Replace this comment and the next line with your code
-  console.log(func);
+  func();
```
remarcmij: 👍

Week3/homework/step2-2.js
```diff
@@ -3,18 +3,28 @@
 function threeFive(startIndex, stopIndex, threeCallback, fiveCallback) {
   const numbers = [];
 
-  // Replace this comment and the next line with your code
-  console.log(startIndex, stopIndex, threeCallback, fiveCallback, numbers);
+  for (let i = 0; i <= stopIndex - startIndex; i++) {
+    numbers.push(startIndex + i);
+  }
```
remarcmij: While this works, it could be a bit simpler:

```js
for (let i = startIndex; i <= stopIndex; i++) {
  numbers.push(i);
}
```

Week3/homework/step2-2.js
```diff
@@ -3,18 +3,28 @@
...
+
+  for (const number of numbers) {
```
remarcmij: 👍 for the `for-of` loop.

Week3/homework/step2-2.js
```diff
@@ -3,18 +3,28 @@
...
+    if (!(number % 3)) {
```
remarcmij: I guess you could read this as "if the number divided by 3 doesn't have a remainder ...", which you could take to mean "if the number is divisible by 3 ...". To me, the expression `if (number % 3 === 0)` more clearly expresses what is intended here.

Week3/homework/step2-2.js
```diff
@@ -3,18 +3,28 @@
...
+      threeCallback(number);
+      if (!(number % 5)) {
+        fiveCallback(number);
+      }
+    } else if (!(number % 5)) {
+      fiveCallback(number);
+    }
+  }
 }
```
remarcmij: Instead of using an `if-then-else` statement, two simple, independent `if` statements are all that is needed here.

```js
for (const number of numbers) {
  if (number % 3 === 0) {
    threeCallback(number);
  }
  if (number % 5 === 0) {
    fiveCallback(number);
  }
}
```

Week3/homework/step2-2.js
```diff
@@ -3,18 +3,28 @@
...
 
 function sayThree(number) {
-  // Replace this comment and the next line with your code
-  console.log(number);
+  console.log(`Number: ${number} is divisible by 3!`);
```
remarcmij: Nice use of ES6 template literals to produce user-friendly output. Use exclamation marks sparingly though, no need to shout!

Week3/homework/step2-3.js
```diff
@@ -5,9 +5,12 @@ function repeatStringNumTimesWithFor(str, num) {
   // eslint-disable-next-line prefer-const
   let result = '';
 
-  // Replace this comment and the next line with your code
-  console.log(str, num, result);
-
+  if (num <= 0) {
+    return result;
+  }
```
remarcmij: This test is redundant and can be removed. The `for` loop body will not be executed if `num <= 0`. This is because the loop condition is tested at the start of the loop.

Week3/homework/step2-3.js
```diff
@@ -18,8 +21,15 @@ function repeatStringNumTimesWithWhile(str, num) {
   // eslint-disable-next-line prefer-const
   let result = '';
 
-  // Replace this comment and the next line with your code
-  console.log(str, num, result);
+  if (num <= 0) {
+    return result;
+  }
```
remarcmij: This test is redundant and can be removed. The `while` loop condition is tested at the start of the loop, therefore the loop body will not be executed if `num <= 0`.

Week3/homework/step2-3.js
```diff
@@ -31,8 +41,15 @@ function repeatStringNumTimesWithDoWhile(str, num) {
...
```
remarcmij: Here you are absolutely correct to test the value of `num` before entering the `do-while` loop as now the loop condition is tested at the end of the loop.

Week3/homework/step2-4.js
```diff
@@ -1,7 +1,9 @@
 'use strict';
 
 function Dog() {
-  // add your code here
+  this.name = 'Doggo';
+  this.color = 'brown';
+  this.numLegs = 5;
 }
 
 const hound = new Dog();
```
remarcmij: 👍

Week3/homework/step2-5.js
```diff
@@ -4,8 +4,11 @@ function multiplyAll(arr) {
   // eslint-disable-next-line
   let product = 1;
 
-  // Replace this comment and the next line with your code
-  console.log(arr, product);
+  for (let i = 0; i < arr.length; i++) {
+    for (let j = 0; j < arr[i].length; j++) {
+      product *= arr[i][j];
```
remarcmij: Nice use of the short-hand multiplication operator.

Week3/homework/step2-6.js
```diff
@@ -2,19 +2,44 @@
...
+const flattenNestedArray = (arr, newArray = []) => {
+  arr.forEach(element => {
+    if (Array.isArray(element)) {
+      flattenNestedArray(element, newArray);
+    } else {
+      newArray.push(element);
+    }
+  });
+  return newArray;
+};
```
remarcmij: Some pretty good stuff in here:
- recursive function
- default parameters
- forEach

And as you say, this works for any dimension: 💯

Week3/homework/step2-6.js
```diff
@@ -2,19 +2,44 @@
...
 function flattenArray2d(arr) {
-  // Replace this comment and the next line with your code
-  console.log(arr);
+  const newArray = [];
+  for (let i = 0; i < arr.length; i++) {
+    for (let j = 0; j < arr[i].length; j++) {
+      newArray.push(arr[i][j]);
+    }
+  }
+  return newArray;
```
remarcmij: This works too, but I wouldn't have minded if you just called your `flattenNestedArray` function here, just as you did in `flattenArray3d`.

Week3/homework/step2-7.js
```diff
@@ -21,3 +21,10 @@ f2(y);
 console.log(y);
 
 // Add your explanation as a comment here
+
+// console.log(f1(x)) === 10 but even after calling f1 function
+// value of x does not change. That's because only the value is passed
+// to the function. Thus x itself is not affected. But when an
+// object or array passed as argument to a function, its reference
+// is passed. So if function change the reference's value, array or
+// object is also affected from this change.
```
remarcmij: I'm completely satisfied with your explanation. 👍

Week3/homework/step3-bonus.js
```diff
@@ -3,8 +3,13 @@
 const values = ['a', 'b', 'c', 'd', 'a', 'e', 'f', 'c'];
 
 function makeUnique(arr) {
-  // Replace this comment and the next line with your code
-  console.log(arr);
+  const uniques = [];
+  arr.forEach(element => {
+    if (uniques.find(unique => unique === element) === undefined) {
```
remarcmij: This could be simplified to:

```js
if (!uniques.find(unique => unique === element)) {
```

Even simpler:

```js
if (!uniques.includes(element)) {
```

Week3/homework/step3.js
```diff
@@ -1,8 +1,9 @@
...
 
 function createBase(base) {
-  // Replace this comment and the next line with your code
-  console.log(base);
+  return function(num) {
+    return base + num;
+  };
 }
 
 const addSix = createBase(6);
```
remarcmij: 👍

remarcmij:
Thanks @suh3yb for the updates. But why did you need to force the push?

suh3yb:
Thank you very much @remarcmij for your very detailed feedback. 
Unfortunately I pushed some changes to this approved homework, sorry about that. I tried to undo but the homework status has already changed and I could not revert it. 😅

suh3yb:
Thanks again. :smile:

PR #369, opened 2019-06-05
abdullah Week3 homework

heartwhite:
js2 week3 homework js files. test week3 run,
 test3 didnt run


Week3/homework/step2-2.js
```diff
@@ -2,19 +2,29 @@
...
+    if (value % 3 === 0) {
+      threeCallback(value);
+      if (value % 5 === 0) {
+        fiveCallback(value);
+      }
+    }
+    if (value % 5 === 0 && !(value % 3 === 0)) {
+      fiveCallback(value);
+    }
+  }
```
remarcmij: This logic seems a bit convoluted. You need to call `threeCallBack` if the number is divisible by 3 and `fiveCallBack` if divisible by 5. You only need two `if` statements like this:

```js
for (const value of numbers) {
  if (value % 3 === 0) {
    threeCallback(value);
  }
  if (value % 5 === 0) {
    fiveCallback(value);
  }
}
```

Week3/homework/step2-2.js
```diff
@@ -2,19 +2,29 @@
...
   console.log(startIndex, stopIndex, threeCallback, fiveCallback, numbers);
```
remarcmij: This console.log was supposed to be replaced by your code (i.e. deleted).

Week3/homework/step2-3.js
```diff
@@ -4,9 +4,15 @@
 function repeatStringNumTimesWithFor(str, num) {
   // eslint-disable-next-line prefer-const
   let result = '';
-
-  // Replace this comment and the next line with your code
-  console.log(str, num, result);
+  const repeatArr = [];
+  if (num <= 0) {
```
remarcmij: This extra check is unneeded because the `for` loop body will also not be executed if `num <= 0` if you leave out this test.

Week3/homework/step2-3.js
```diff
@@ -17,8 +23,17 @@ console.log('for', repeatStringNumTimesWithFor('abc', 3));
 function repeatStringNumTimesWithWhile(str, num) {
   // eslint-disable-next-line prefer-const
   let result = '';
-
-  // Replace this comment and the next line with your code
+  const repeatArr = [];
+  if (num <= 0) {
```
remarcmij: The same applies here. Because the loop condition for the while loop is tested before the loop body is executed, there is no need for an extra check beforehand.

Week3/homework/step2-3.js
```diff
@@ -30,8 +45,17 @@ console.log('while', repeatStringNumTimesWithWhile('abc', 3));
 function repeatStringNumTimesWithDoWhile(str, num) {
   // eslint-disable-next-line prefer-const
   let result = '';
-
-  // Replace this comment and the next line with your code
+  const repeatArr = [];
+  if (num <= 0) {
```
remarcmij: The `do-while` loop, on the other hand, checks the loop condition only at the end of the loop. Therefore, without an extra check it will execute the loop body at least once. However, there is no need to re-initialize `result` with an empty string as it already has that value:

```js
let result = '';
if (num > 0) {
  const repeatArr = [];
  let i = 0;
  while (i < num) {
    repeatArr.push(str);
    i++;
  }
  result = repeatArr.join('');
}
```

Week3/homework/step2-3.js
```diff
@@ -17,8 +23,17 @@ console.log('for', repeatStringNumTimesWithFor('abc', 3));
...
+    result = '';
+  } else {
+    let i = 0;
+    while (i < num) {
+      repeatArr.push(str);
+      i++;
+    }
+    result = repeatArr.join('');
+  }
   console.log(str, num, result);
```
remarcmij: This console.log was supposed to be replaced/deleted.

Week3/homework/step2-3.js
```diff
@@ -30,8 +45,17 @@ console.log('while', repeatStringNumTimesWithWhile('abc', 3));
...
+    result = '';
+  } else {
+    let i = 0;
+    do {
+      repeatArr.push(str);
+      i++;
+    } while (i < num);
+    result = repeatArr.join('');
+  }
   console.log(str, num, result);
```
remarcmij: This console.log was supposed to be replaced/deleted.

Week3/homework/step2-4.js
```diff
@@ -1,10 +1,13 @@
...
-  // add your code here
+function Dog(name, color) {
+  this.name = name;
+  this.numLegs = 4;
+  this.enemies = ['cats'];
+  this.color = color;
 }
 
-const hound = new Dog();
+const hound = new Dog('Jimmy', 'black');
```
remarcmij: Okay, no offence taken 😄.

Week3/homework/step2-5.js
```diff
@@ -1,15 +1,18 @@
...
 function multiplyAll(arr) {
-  // eslint-disable-next-line
   let product = 1;
-
-  // Replace this comment and the next line with your code
-  console.log(arr, product);
-
+  for (let i = 0; i < arr.length; i++) {
+    for (let j = 0; j < arr[i].length; j++) {
+      product = product * arr[i][j];
```
remarcmij: You could use a short-hand operator here:

```js
product *= arr[i][j];
```

Week3/homework/step2-6.js
```diff
@@ -4,13 +4,26 @@ const arr2d = [[1, 2], [3, 4], [5, 6]];
...
+      for (let k = 0; k < arr[i][j].length; k++) {
+        flattenArr.push(arr[i][j][k]);
+      }
+    }
+  }
+  console.log(flattenArr);
+  return flattenArr;
 }
 
 console.log(flattenArray2d(arr2d)); // -> [1, 2, 3, 4, 5, 6]
```
remarcmij: These are straight-forward, fit-for-purpose implementations that work with the arrays provided. Have you thought about how to do this for arrays of any dimension?

Week3/homework/step2-7.js
```diff
@@ -21,3 +21,25 @@ f2(y);
...
+if the existing one is a primitive, than it copies the value to new one 
+and the existing one is completely independent from any affect on new one. 
+but if i assign a variable to a existing object, than it snot a copy, 
+it is a reference to existing variables value. it is a way to there. than changes 
+affects both variable. value is one thing, variables are like access to it or sth like that.
+i sow 3 page, but i understood these from;
+https://codeburst.io/javascript-passing-by-value-vs-reference-explained-in-plain-english-8d00fd06a47c
+
+thanks.
+*/
```
remarcmij: I'm happy with your explanation.

Week3/homework/step3-bonus.js
```diff
@@ -1,14 +1,44 @@
...
+  '11',
+  '54',
+];
 function makeUnique(arr) {
-  // Replace this comment and the next line with your code
-  console.log(arr);
+  for (let i = 0; i < arr.length; i++) {
+    for (let j = i + 1; j < arr.length; j++) {
+      if (arr[i] === arr[j]) {
+        arr.splice(j, 1);
```
remarcmij: While this works, I would recommend against deleting elements from an array while you are relying on its (changing) length. It is like trying to remove the leg of a chair you sit on. I would prefer a method that creates a new array with unique values without modifying (i.e., 'mutating') the existing array. Can you think of such a method?

Week3/homework/step3.js
```diff
@@ -1,8 +1,10 @@
 'use strict';
 
 function createBase(base) {
-  // Replace this comment and the next line with your code
-  console.log(base);
+  function addition(x) {
+    return base + x;
+  }
+  return addition;
```
remarcmij: Fine. But you can also return an anonymous function like this:

```js
function createBase(base) {
  return function(x) {
    return base + x;
  };
}
```

Or, if you are adventurous:

```js
const createBase = base => x => base + x;
```

Week3/homework/step2-4.js
heartwhite: :) it was one of names from Lyndas video lesson (or FCC challenge).

Week3/homework/step2-1.js
```diff
@@ -1,9 +1,8 @@
 'use strict';
 
 function foo(func) {
-  // What to do here?
-  // Replace this comment and the next line with your code
-  console.log(func);
+  console.log('and a voice heard from deep:    ');
```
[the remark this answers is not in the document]
heartwhite: I removed unneeded part.

Week3/homework/step2-1.js
```diff
@@ -1,9 +1,8 @@
...
+  func();
```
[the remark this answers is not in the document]
heartwhite: thanks

Week3/homework/step2-2.js
```diff
@@ -2,19 +2,29 @@
 
 function threeFive(startIndex, stopIndex, threeCallback, fiveCallback) {
   const numbers = [];
-
-  // Replace this comment and the next line with your code
+  for (let i = startIndex; i < stopIndex + 1; i++) {
+    numbers.push(i);
+  }
+  for (const value of numbers) {
```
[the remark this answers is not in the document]
heartwhite: yes. you re right. I need to get used to do logical naming.

Week3/homework/step2-2.js
heartwhite: I don't know why but somehow I thought only one of if statements code will work. but no, if statements run one after another.
I updated my code

Week3/homework/step2-2.js
heartwhite: removed.

Week3/homework/step2-5.js
heartwhite: yes, I used it now.!

Week3/homework/step2-6.js
heartwhite: I updated my code. I couldn't achieve the K dimension array but typed some code which shows my mind ...

Week3/homework/step2-7.js
heartwhite: thanks

Week3/homework/step3-bonus.js
heartwhite: I updated my code using a new array and without touching source array.

Week3/homework/step3.js
heartwhite: I am not so adventurous for now. but the anonymous function nice. I changed my code.

Week3/homework/step2-3.js
remarcmij: The change that you made boils down to letting the error happen and then fixing it afterwards. The correct way of handling this is to not let the error occur at all. This is still something I would like you to fix.

heartwhite: I just used '  let result;  '  now.

remarcmij: That is not quite what I meant. You should not execute the do-while loop at all if the loop condition `num > 0` does not hold true. Otherwise you might fall off the cliff before you know it, just like the rabbit here:

https://github.com/HackYourFuture/fundamentals/blob/master/fundamentals/loops.md#dowhile

heartwhite: oh..! yes understand. 
I used an if statement. I make sense.
is ti right?

Week3/homework/step2-3.js
```diff
@@ -29,10 +35,19 @@ console.log('while', repeatStringNumTimesWithWhile('abc', 3));
...
 function repeatStringNumTimesWithDoWhile(str, num) {
   // eslint-disable-next-line prefer-const
-  let result = '';
-
-  // Replace this comment and the next line with your code
-  console.log(str, num, result);
+  let result;
+  const repeatArr = [];
+  let i = 0;
+  if (i >= 0) {
```
remarcmij: The condition of this `if` statement will always be `true` (i.e. `i >= 0`) because you set `i` to zero just before this `if` statement. Therefore the do-loop will always be executed once. You need to test `num`, not `i` before entering the loop.

Week3/homework/step2-3.js
```diff
@@ -29,10 +35,19 @@ console.log('while', repeatStringNumTimesWithWhile('abc', 3));
...
+    do {
+      repeatArr.push(str);
+      i++;
+    } while (i < num);
+    result = repeatArr.join('');
+    if (num <= 0) {
+      result = '';
+    }
```
remarcmij: If you have correctly implemented the test **_before_** entering the loop you no longer need lines 47-49.

Week3/homework/step2-3.js
heartwhite: I made it only ' > '

Week3/homework/step2-3.js
heartwhite: I think I could not change my focus to this file. that's why every time I don't see a stupid point. 
I thought to type lines 47 and 48 but I couldn't think >= will work for 0 anytime.
I updated with if i > 0, and I removed the 47-49 lines

remarcmij: I can see that you are struggling with this. Making the test `i > 0` doesn't help. You need to test `num`. That is the parameter that is passed to your function and that might have a value <= 0. So the test in your `if` statement should be `num > 0`, not `i > 0`.

Does that now make sense?

heartwhite: oh why am I like that..
I am sorry for taking that much time of you. 
thanks for understanding. 
and yes. the action is about 'num' not about ' i '. just loop uses ' i ' for counting.

PR #377, opened 2019-07-02
add compeleted homework js2-week1 /Adil

musafir928:


musafir928:
index.html - basic html page layout
style.css - html pae style
app.js - mian function includes (createUlList function, displayBookInformation function and add picture function)

PR #378, opened 2019-07-03
hyf-js2-week1 homework from Shyar

ShyarK:
Please review my homework!

### DIFF
--- a/Week1/homework/app.js
+++ b/Week1/homework/app.js
@@ -37,7 +37,7 @@
     }
     
     console.log(ul);
-  };
+  }
 // This function will generate ul element
   const generateUl = () =>{
     const ul = document.createElement('ul');
--- a/Week1/homework/app.js
+++ b/Week1/homework/app.js
@@ -34,10 +34,9 @@
      const list = document.createElement('li');
       list.textContent = bookTitles[i]; 
       ul.appendChild(list);
-    }
-    
+    } 
     console.log(ul);
-  }
+  };
 // This function will generate ul element
   const generateUl = () =>{
     const ul = document.createElement('ul');
--- a/Week1/homework/app.js
+++ b/Week1/homework/app.js
@@ -1,11 +1,11 @@
 'use strict';
 
-// 1.1 Open the apps.js and start by declaring an array that contains 10 strings. 
+// 1-1 Open the apps.js and start by declaring an array that contains 10 strings 
 // These strings should be of book titles you have read (or made up) and be lowercase 
-// without spaces so that you can use these later as HTML id attributes. 
-// (Example: Harry Potter's - The Chamber of Secrets -> harry_potter_chamber_secrets). 
-// Add a console.log statement to output this array to console. 
-// (This is for debugging and making sure everything is in order.
+// without spaces so that you can use these later as HTML id attributes
+// (Example: Harry Potter's - The Chamber of Secrets -> harry_potter_chamber_secrets)
+// Add a console.log statement to output this array to console 
+// (This is for debugging and making sure everything is in order
 // Delete it later when you're done :))
 
 {
@@ -29,7 +29,7 @@
   // I set the array elements which are string as a text content for li element using textContent property.
   // I append all li elements to the ul by using append method to add a node to the given parent node parent.appendChild(child).
   // For loop will loop as long as the condition is true.
-  const createList = (ul) => {
+  const createList = (ul => {
     for (let i = 0; i < bookTitles.length; i++) {
      const list = document.createElement('li');
       list.textContent = bookTitles[i]; 
@@ -46,9 +46,6 @@
 
   generateUl();
 
-
-
-
   // I created books object which, includes 10 properties, and each property's value is 
   //another object nested inside books object, inside that object I have four keys and values:
   // Here I wanted to create a constructor function! But I didn't want to get out of question context.  
@@ -56,6 +53,7 @@
   // title of the book.
   // the name of the author.
   // book's language.
+  
   const books = {
     'the-da-vinci-code': {
       title: 'The Da Vinci Code',
--- a/Week1/homework/app.js
+++ b/Week1/homework/app.js
@@ -1,13 +1,5 @@
 'use strict';
 
-// 1-1 Open the apps.js and start by declaring an array that contains 10 strings
-// These strings should be of book titles you have read (or made up) and be lowercase
-// without spaces so that you can use these later as HTML id attributes
-// (Example: Harry Potter's - The Chamber of Secrets -> harry_potter_chamber_secrets)
-// Add a console.log statement to output this array to console
-// (This is for debugging and making sure everything is in order
-// Delete it later when you're done :))
-
 {
   const bookTitles = [
     // Replace with your own book titles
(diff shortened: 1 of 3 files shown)


Week1/homework/index.html
```diff
@@ -1 +1,23 @@
...
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <meta http-equiv="X-UA-Compatible" content="ie=edge">
+    <link rel="stylesheet" href="style.css">
+    <title>Document</title>
+</head>
+<body>
+    <header>
+        <nav>
+            <ul>
+                <li>About</li>
```
yash-kapila: Do we need this navigation list for About and Contact here?

Week1/homework/index.html
```diff
@@ -1 +1,23 @@
...
+<head>
+    <meta charset="UTF-8">
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <meta http-equiv="X-UA-Compatible" content="ie=edge">
+    <link rel="stylesheet" href="style.css">
+    <title>Document</title>
+</head>
+<body>
+    <header>
+        <nav>
```
yash-kapila: I like your choice of using semantic HTML tags i.e `header`, `nav` and so on. You can also look up `main` tag where you can insert the main content of your app which in this case is the books 🙂

Week1/homework/app.js
```diff
@@ -1,11 +1,169 @@
 'use strict';
 
+// 1-1 Open the apps.js and start by declaring an array that contains 10 strings
```
yash-kapila: We need not have this comment as it doesn't really add much value here. It would have been helpful to you during development but we need not have it in the final version, right? 😉

Week1/homework/app.js
```diff
@@ -1,11 +1,169 @@
...
 
-  // Replace with your own code
-  console.log(bookTitles);
+  // I have two functions, one to create a li items, and the second is to create ul element which is the parent of li
+  // I used for loop to loop through the array items, inside for loop I generate a list item
+  // I used document.createElement() method to create the HTML element,  which takes one parameter (tagName).
+  // I set the array elements which are string as a text content for li element using textContent property.
+  // I append all li elements to the ul by using append method to add a node to the given parent node parent.appendChild(child).
+  // For loop will loop as long as the condition is true.
+
```
yash-kapila: Hi Shyar - even though comments are helpful but in this case I think they are a bit too much. The code we write should more or less explain what the intention is of a function. In such cases, it is quite important that you name the functions correctly, have proper variable names and use more out of the box methods like Array.map, Array.filter or even the for loop which you are using. For example, your function names are quite understandable and explain what to expect out of them.

Week1/homework/app.js
```diff
@@ -1,11 +1,169 @@
...
+  const createList = ul => {
+    for (let i = 0; i < bookTitles.length; i++) {
+      const list = document.createElement('li');
+      list.textContent = bookTitles[i];
+      ul.appendChild(list);
+    }
+    console.log(ul);
```
yash-kapila: Ah - unnecessary console log statement here which can be removed.

Week1/homework/app.js
```diff
@@ -1,11 +1,169 @@
...
+    'harry-potter-and-the-order-of-the-phoenix': {
+      title: 'Harry Potter and the Order of the Phoenix',
+      cover:
+        'https://upload.wikimedia.org/wikipedia/en/7/70/Harry_Potter_and_the_Order_of_the_Phoenix.jpg',
+      language: 'English',
+      author: 'By J. K. Rowling',
+    },
+  };
+
+  // I created another function, which takes books object as a parameter  Object.keys() method returns an array of a given object's own property names
```
yash-kapila: I think the comments here again are a bit too much. You can avoid having such long comments in the future assignments. For example, if you name the parameter of your function as `books` instead of `obj`, you wouldn't need the comment `function which takes books object as a parameter`

Week1/homework/app.js
```diff
@@ -1,11 +1,169 @@
...
+  // if you use dot notation JS thinks that I'm looking for a property that its name is property.
+  // Used if statement to set alt attribute only when the elementName is img.
+  // For how to use bracket notation:
+  // https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Property_Accessors
+
+  const listOfBooks = obj =>
+    Object.keys(obj).forEach(item => {
+      const booksContainer = document.getElementById('books-container');
+
+      const bookDiv = document.createElement('div');
```
yash-kapila: Although it works, semantically speaking, having a `div` tag acting as a list container instead of `ul` isn't the best idea.

Week1/homework/app.js
```diff
@@ -1,11 +1,169 @@
...
+      bookDiv.setAttribute('class', 'book-info');
+
+      const createElementAndSetAttr = (elementName, name, value, altName, altValue) => {
+        const element = document.createElement(elementName);
+        element.setAttribute(name, value);
+        if (elementName === 'img') {
+          element.setAttribute(altName, altValue);
+        }
+        bookDiv.appendChild(element);
```
yash-kapila: If you note closely, the function name doesn't correctly represent the actions inside it. For example, you are creating an element and attaching an attribute to it but you are also attaching it to the container. I would suggest that you **return** the new element from this function and append it to `bookDiv` where it's being created. Something like:

```
bookDiv.appendChild(createElementAndSetAttr('img', 'src', obj[item].cover, 'alt', obj[item].title));
```

or

```
const bookImage = createElementAndSetAttr('img', 'src', obj[item].cover, 'alt', obj[item].title);
bookDiv.appendChild(bookImage);
```

Same goes for the `createElementAndAppend` function 🙂

Week1/homework/app.js
yash-kapila: Also, `bookDiv` isn't the best of variable names. How about `booksContainer` because that is what the element acts like, isn't it?

yash-kapila:
Another thing I would like to add is that you should create a new branch from `master` for each week's homework. I see for this homework, you did it all on the `master` branch 🙂

Week1/homework/index.html
ShyarK: @yash-kapila Thank you for having time to review my homework, much appreciated!
I'm not sure if I got you well! Do you mean using About and Contact inside li is not correct or we don't need to use it here at all? Could you please clarify what you mean?

Week1/homework/index.html
ShyarK: I changes div with html<main> element!

Week1/homework/app.js
ShyarK: I deleted unnecessarily comment!

Week1/homework/app.js
ShyarK: I respect your point but from my perspective, even though my function names are quite understandable, but comments will help me later on when coming back to check my codes and understand what I exactly have done!

Week1/homework/app.js
ShyarK: Unnecessary console log statement has been removed!

Week1/homework/app.js
ShyarK: I'm not sure if I got you well! I created another function and I append child to it! Please have a look if this what you've expected!

Week1/homework/app.js
ShyarK: I created a new list item and appended to the ul element!

Week1/homework/index.html
yash-kapila: I don't think we need them at all for this homework.

PR #379, opened 2019-07-03
javascript2 week 1 homework - Salih S -- Class22

salih18:
there are 2 functions in the codes. One of them creates elements for book properties, the other one creates elements for book cover.


Week1/homework/app.js
```diff
@@ -1,11 +1,158 @@
...
+  //   ul.setAttribute('id', 'books-id');
+
+  //   myFavoriteBooks.forEach(book => {
+  //     const li = document.createElement('li');
+  //     li.classList.add('book');
+  //     li.textContent = book;
+  //     ul.append(li);
+  //   });
+
+  //   document.body.prepend(ul);
```
yash-kapila: Ah nice - you learnt a new method.

Week1/homework/app.js
```diff
@@ -1,11 +1,158 @@
...
+  //   'war_and_peace',
+  //   'crime_and_punishment',
+  //   'the_great_gasby',
+  //   'the_red_and_the_black',
+  //   'the_lord_of_the_rings',
+  // ];
+
+  // 1.3
+
+  // const listBooks = () => {
```
yash-kapila: Just a suggestion - a better function name could be `generateListOfBooks`. If you look at the Naming Conventions document from JS1, function names should contain a verb which describes the action being performed by it which in this case, we are trying to generate a list of books.

Week1/homework/app.js
```diff
@@ -1,11 +1,158 @@
...
+    the_red_and_the_black: 'books_cover/red_and_black.jpg',
+  };
+
+  const listBooksInDetail = bookList => {
+    Object.entries(bookList).forEach(book => {
+      //
+      const bookID = book[0];
+      const bookDetails = book[1];
+      // create a div element for each book and write them to body
+      const div = document.createElement('div');
```
yash-kapila: Even though the code works just fine, in future be mindful of using semantic tags while creating HTML elements. For example, if we need to create a list, we should use `ul` and `li` tags instead of using `div` tags. Also, we should be using `h1, h2, ..., h6` tags for headings instead of resized `p` or `span`. Not only it gives a better understanding of the HTML but is also helpful for accessibility.

In this case, when I run your application, I see an unordered list with a list of items(an item being a container of book image on left and some details on the right side. However, when we inspect the HTML in browser, I see a list of `div` tags with two `ul` tags and the first one without any `li` item in it.

It's fine for JS1-Week1 but please take care of it in the future 🙂

Week1/homework/app.js
```diff
@@ -1,11 +1,158 @@
...
+      div.classList.add('book-list');
+      document.body.prepend(div);
+      // create a ul element
+      const ul = document.createElement('ul');
+      ul.classList.add('book');
+      div.prepend(ul);
+      // create a li element for each book properties
+      const liTitleEl = document.createElement('li');
+      liTitleEl.textContent = bookDetails.title;
+      liTitleEl.classList.add('list', 'title');
```
yash-kapila: I see a repetition of two lines of code throughout this function and the next one:

```
const liTitleEl = document.createElement('li');
liTitleEl.classList.add('list', 'title');
```

Maybe you can create a function out of it like `createElement` and move this repeated logic in it. Something like this perhaps:

```
const createElement = (tag, ...cssClass) => {
  const elem = document.createElement(tag);
  cssClass.forEach(className => {
    elem.classList.add(className);
  });  
  return elem;
}
```

And now you can use this function in other places. Give it a try if you like. 🙂

PR #380, opened 2019-07-04
Homework JavaScript2-Week1/ Farzanna

Farzanna:
Farzanna_Homework_JavaScript2_Week1

### DIFF
--- a/Week1/homework/app.js
+++ b/Week1/homework/app.js
@@ -107,97 +107,6 @@
     lust_life:
       'https://i.gr-assets.com/images/S/compressed.photo.goodreads.com/books/1348532272i/79834.jpg',
   };
-   function makeList() {
-    const body = window.document.getElementById('body');
-
-    const ul = window.document.createElement('ul');
-
-    for (let i = 0; i < bookTitles.length; i++) {
-      const li = window.document.createElement('li');
-
-      ul.appendChild(li);
-      li.innerHTML = li.innerHTML + bookTitles[i];
-    }
-    body.appendChild(ul);
-  }
-  window.onload = () => makeList();
-}
-{
-  const bookInfo = {
-    parallel_worlds: {
-      title: 'Parallel Worlds: The Science of Alternative Universes & Our Future in the Cosmos',
-      language: 'english',
-      author: 'Michio Kaku',
-    },
-    masnavi: {
-      title: 'masnavi',
-      language: 'persian',
-      author: ' Rumi',
-    },
-    black_holes_baby_universes: {
-      title: 'Black Holes and Baby Universes',
-      language: 'english',
-      author: 'Stephen Hawking',
-    },
-    blindness: {
-      title: 'Blindness',
-      language: 'english',
-      author: 'José Saramago',
-    },
-    one_Hundred_years_solitude: {
-      title: 'One Hundred Years of Solitude',
-      language: 'english',
-      author: ' Gabriel García Márquez',
-    },
-    girl_train: {
-      title: 'The Girl on the Train',
-      language: 'english',
-      author: 'Paula Hawkins',
-    },
-    habits_highly_effective_people: {
-      title: 'The 7 Habits of Highly Effective People: Powerful Lessons in Personal Change',
-      language: 'english',
-      author: 'Stephen R. Covey',
-    },
-    philosophy_fear: {
-      title: 'A Philosophy of Fear',
-      language: 'english',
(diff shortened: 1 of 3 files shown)


Week1/homework/app.js
```diff
@@ -1,11 +1,153 @@
...
+  function makeList() {
+    const body = window.document.getElementById('body');
+
+    const ul = window.document.createElement('ul');
+
+    for (let i = 0; i < bookTitles.length; i++) {
+      const li = window.document.createElement('li');
+
+      ul.appendChild(li);
+      li.innerHTML = li.innerHTML + bookTitles[i];
```
yash-kapila: Is there a reason why you did it like this than simply `li.innerHTML = bookTitles[i];`?

Week1/homework/app.js
```diff
@@ -1,11 +1,153 @@
...
+    }
+    body.appendChild(ul);
+  }
+  window.onload = () => makeList();
+}
+{
```
yash-kapila: Just wanted to confirm - is the reason of having two different block scopes because you wanted to have functions with the same name `makeList` or did you have a different reason for it?

Week1/homework/app.js
```diff
@@ -1,11 +1,153 @@
...
+
+        // const h4Language = window.document.createElement('h4');
+        // h4Language.innerHTML = "Language:" + bookInfo[item].language;
+        // li.appendChild(h4Language);
+      }
+    }
+    body.appendChild(ul);
+  }
+
+  window.onload = () => makeList();
```
yash-kapila: If you remember from our lesson of Week1, if we make use of event listeners like this i.e `window.onload` instead of using `addEventListeners`, it overrides any previous listeners. For example, you have an event listener here as well as on line 32. Since this one is parsed at the end by JavaScript, it will override the previous event listener. I hope you were aware of it when you did it like this 🙂

Week1/homework/app.js
```diff
@@ -1,11 +1,153 @@
...
+
+  function makeList() {
+    const body = window.document.getElementById('body');
+    const ul = window.document.createElement('ul');
+    // I think there is no need to add id to lists, because we can find it just by item name at first I used this codes but now I comment it:
+    // li.setAttribute('id', item);
+    // console.log(bookCover[item]);
+
+    for (const item in bookInfo) {
+      if (item) {
```
yash-kapila: Is there a reason of having this `if` check?

Week1/homework/app.js
```diff
@@ -1,11 +1,153 @@
...
+  };
+
+  function makeList() {
+    const body = window.document.getElementById('body');
+    const ul = window.document.createElement('ul');
+    // I think there is no need to add id to lists, because we can find it just by item name at first I used this codes but now I comment it:
+    // li.setAttribute('id', item);
+    // console.log(bookCover[item]);
+
+    for (const item in bookInfo) {
```
yash-kapila: Err!! You should avoid using `for-in` loops as it may have unexpected results. Remember we saw an example in classroom last week? Try and use `Object.keys`, `Object.values` or `Object.entries` instead. 🙂

Week1/homework/app.js
```diff
@@ -1,11 +1,153 @@
...
+        image.setAttribute('src', bookCover[item]);
+        li.appendChild(image);
+
+        ul.appendChild(li);
+        const h3Title = window.document.createElement('h3');
+        h3Title.innerHTML = bookInfo[item].title;
+        li.appendChild(h3Title);
+
+        const bookProperties = Object.keys(bookInfo[item]);
+        for (const itemChild in bookProperties) {
```
yash-kapila: Aahaan. `bookProperties` variable is an array rather than an `object`. We use `for-in` loops for iteration over an objects. You could have made use of `forEach` for a regular `for-of` loop here instead. 🙂

Week1/homework/app.js
```diff
@@ -1,11 +1,153 @@
...
+          if (itemChild !== 0) {
+            const bookProperty = bookProperties[itemChild];
+            const h4Author = window.document.createElement('h4');
+            h4Author.innerHTML = bookProperty + ':' + bookInfo[item][bookProperty];
+            li.appendChild(h4Author);
+          }
+        }
+        // I made another loop inside each object to get out each property inside nested object insted of calling them by their names like this command codes:
```
yash-kapila: I see. Curious to know if there is a reason you selected to use a loop instead of doing as you show in the comments here?

PR #382, opened 2019-07-04
Add homework files

gulyener:


Week1/homework/app.js
```diff
@@ -1,11 +1,303 @@
...
+};
+
+window.addEventListener('load', getBookInfo);
+window.addEventListener('load', getBookCovers);
 
-  // Replace with your own code
-  console.log(bookTitles);
-}
+// I used element.attribute instead of setAttribute() throughout this homework
+// it feels easier to do it that way, but is it bad practice?
```
nudge: Using `element.attribute` is perfectly fine in practice!

Week1/homework/app.js
```diff
@@ -1,11 +1,303 @@
...
+  'Crime and Punishment': './images/crime_and_punishment.jpeg',
+  'Moby Dick': './images/moby_dick.jpg',
+  Hamlet: './images/hamlet.jpg',
+  'Madame Bovary': './images/madame_bovary.jpg',
+  'Jane Eyre': './images/jane_eyre.jpg',
+  'Pride and Prejudice': './images/pride_and_prejudice.jpg',
+  'Anna Karenina': './images/anna_karenina.jpeg',
+};
+
+// 1.8 I could not do this with Object.keys(objectName) I hope I can learn it from someone
```
nudge: A way to do this is to set the `id` attribute on the `li` element you create in `getBookInfo()`, then in `getBookCovers()` you can iterate over the `li` elements, grab the `id`, and match it with `myBooks`.

Week1/homework/.vscode/launch.json
```diff
@@ -0,0 +1,15 @@
...
+  "configurations": [
+    {
+      "type": "chrome",
+      "request": "launch",
+      "name": "Launch Chrome against localhost",
+      "url": "http://localhost:8080",
+      "webRoot": "${workspaceFolder}"
+    }
+  ]
+}
```
nudge: You don't need to check this file in, you can remove it from the pull request

Week1/homework/app.js
```diff
@@ -1,11 +1,303 @@
...
+//     li.innerHTML = bookTitles[i];
+//     ul.appendChild(li);
+//   }
+// };
+// document.getElementById('renderList').appendChild(ul);
+
+// 1.4
+
+const myBooks = {
+  'War and Peace': {
```
nudge: Usually keys are used as identifiers. I've noticed you're using them here for display as titles of the books. A more flexible approach is to keep the key as an identifier, e.g. `war_and_peace`, then add a key within the object like `title: 'War and Peace'`.

Week1/homework/app.js
gulyener: Okay that's good to know thank you for the information

PR #386, opened 2019-07-04
I hope I had done my homework week1 as needed.

hnnhack:
JS2 W1 homework

### DIFF
--- a/Week1/homework/app.js
+++ b/Week1/homework/app.js
@@ -1,11 +1,135 @@
 'use strict';
 
 {
+  /*
   const bookTitles = [
-    // Replace with your own book titles
-    'harry_potter_chamber_secrets',
+    'Java_Script',
+    'Eloquent_JavaScript',
+    'The_JavaScript',
+    'High_Performance_JavaScript',
+    'JavaScript_Patterns',
+    'Secrets_of_the_JavaScript_Ninja',
+    'Effective_JavaScript',
+    'Professional_JavaScript_for_Web_Developers',
+    'Learning_Javascript_Design_Patterns',
+    'Closure',
   ];
 
-  // Replace with your own code
-  console.log(bookTitles);
+  const newUl = document.createElement('ul');
+  bookTitles.forEach(book => {
+    const listItem = document.createElement('li');
+    const itemText = document.createTextNode(book);
+    listItem.appendChild(itemText);
+    newUl.appendChild(listItem);
+  });
+
+  document.body.appendChild(newUl);
+  */
+
+  const bookObject = {
+    'Java Script': {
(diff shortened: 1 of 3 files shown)


Week1/homework/app.js
```diff
@@ -1,11 +1,135 @@
...
+      source:
+        'https://i.gr-assets.com/images/S/compressed.photo.goodreads.com/books/1328761606i/8300295.jpg',
+    },
+  };
+
+  const createBookCard = () => {
+    /* Creating <ul> </ul> in the index.html */
+    const newUlElement = document.createElement('ul');
+    /* setting attribute to the <ul> </ul> */
+    newUlElement.setAttribute('id', 'grid-flex');
```
yash-kapila: `ID` attributes usually should be chosen in such a way that they can uniquely identify an element in a page. I think in this case, `grid-flex` seems to be a very generic selector which would be better suited as a class name rather than `id` selector. 🙂

Week1/homework/app.js
```diff
@@ -1,11 +1,135 @@
...
+    const bookNames = Object.keys(bookObject);
+    for (let i = 0; i < bookNames.length; i++) {
+      const newLiElement = document.createElement('li');
+      newLiElement.setAttribute('class', 'grid-item');
+      newUlElement.appendChild(newLiElement);
+
+      const newHeading = document.createElement('h3');
+      newLiElement.appendChild(newHeading);
+      const book = bookNames[i];
+      newHeading.innerHTML += `${book}`;
```
yash-kapila: Why do we need to do it like this rather than `newHeading.innerHTML = book;`? 🤔

Week1/homework/app.js
```diff
@@ -1,11 +1,135 @@
...
+
+      const newImage = document.createElement('img');
+      //  newImg.setAttribute('src', '/images/' + `${book}` + '.jpg');
```
yash-kapila: Unnecessary comment here. We can remove it. 🙂

Week1/homework/app.js
```diff
@@ -1,11 +1,135 @@
...
+      newImage.setAttribute('src', bookObject[`${book}`].source);
+      newImage.setAttribute('alt', bookObject[`${book}`].title);
+      newLiElement.appendChild(newImage);
+
+      const newParagraph = document.createElement('p');
+      newLiElement.appendChild(newParagraph);
+      newParagraph.innerHTML += 'Author: ' + bookObject[`${book}`].author;
```
yash-kapila: Do we need the `+=` here?

hnnhack:
Hi Yash!
It's ok, it's not late. Thank you for reviewing. I will fix your commented
lines.

PR #403, opened 2019-07-08
Adham week02 homework

Adham-Elias-Botours:
The file maartjes-work.js does not pass the test, and it will never pass it. And I will never change it to be passed. My steps to solve the situation are very different from the person who write the question.
I believe that I have no misunderstanding about: filter() Method, map() Method and reduce() Method. 
I wrote a function to collect the data, then I used  used Reduse to get one number instead of array. Then I calculated the earnings. 
so I do not Need map, Filter, or function like the expected one.

### DIFF
--- a/Week2/homework/maartjes-work.js
+++ b/Week2/homework/maartjes-work.js
@@ -46,7 +46,7 @@ const maartjesTasks = monday.concat(tuesday);
 const maartjesHourlyRate = 20;
 
 // Adham's code
-const maartjesDurationsHours = []; /* an array to collect our data */
+const maartjesDurationsHours = []; /* Collect our data in one array */
 maartjesTasks.forEach(element => {
   const durationMinutes = element.duration;
   const durationHours = durationMinutes / 60;
@@ -58,18 +58,21 @@ maartjesTasks.forEach(element => {
 const reducingProcess = (totalToLeft, right) => totalToLeft + right;
 const maartjesTotalDuration = maartjesDurationsHours.reduce(reducingProcess);
 // console.log(maartjesTotalDuration); /* One Number: the total of durations in hours */
+
 const maartjesEarnings = maartjesTotalDuration * maartjesHourlyRate;
-// console.log(maartjesEarnings); /* One Number: The exactly earnings with lots of decimals */
-const computeEarnings = maartjesEarnings.toFixed(2);
-// console.log(computeEarnings); /* Number with 2 decimals as string */
-const ComputeEarningsNumber = parseFloat(computeEarnings);
-// console.log(ComputeEarningsNumber);
-console.log(`Maartje has earned € ${computeEarnings}`);
-console.log(`Maartje has earned € ${ComputeEarningsNumber}`);
+// console.log(earnings); /* One Number: The exactly earnings with lots of decimals */
+// add code to convert `earnings` to a string rounded to two decimals (euro cents)
+const earnings2decimals = maartjesEarnings.toFixed(2);
+// console.log(Earnings2decimals); /* Number with 2 decimals as string */
+const earningsNumber = parseFloat(earnings2decimals);
+// console.log(earningsNumber);
+
+console.log(`Maartje has earned € ${maartjesEarnings}`);
+console.log(`Maartje has earned € ${earningsNumber}`);
 
 // Do not change or remove anything below this line
 module.exports = {
   maartjesTasks,
   maartjesHourlyRate,
-  computeEarnings,
+  maartjesEarnings,
 };
--- a/Week2/homework/maartjes-work.js
+++ b/Week2/homework/maartjes-work.js
@@ -45,17 +45,27 @@ const tuesday = [
 const maartjesTasks = monday.concat(tuesday);
 const maartjesHourlyRate = 20;
 
-function computeEarnings(tasks, hourlyRate) {
-  // Replace this comment and the next line with your code
-  console.log(tasks, hourlyRate);
-}
-
-// eslint-disable-next-line no-unused-vars
-const earnings = computeEarnings(maartjesTasks, maartjesHourlyRate);
-
-// add code to convert `earnings` to a string rounded to two decimals (euro cents)
-
-console.log(`Maartje has earned €${'replace this string with the earnings rounded to euro cents'}`);
+// Adham's code
+const maartjesDurationsHours = []; /* an array to collect our data */
+maartjesTasks.forEach(element => {
+  const durationMinutes = element.duration;
+  const durationHours = durationMinutes / 60;
+  if (durationHours >= 2) {
+    maartjesDurationsHours.push(durationHours);
+  }
+});
+// console.log(maartjesDurationsHours); /* Array of durations in hours, just if the duration is more than 2 hours */
+const reducingProcess = (totalToLeft, right) => totalToLeft + right;
+const maartjesTotalDuration = maartjesDurationsHours.reduce(reducingProcess);
+// console.log(maartjesTotalDuration); /* One Number: the total of durations in hours */
+const maartjesEarnings = maartjesTotalDuration * maartjesHourlyRate;
+// console.log(maartjesEarnings); /* One Number: The exactly earnings with lots of decimals */
+const computeEarnings = maartjesEarnings.toFixed(2);
+// console.log(computeEarnings); /* Number with 2 decimals as string */
+const ComputeEarningsNumber = parseFloat(computeEarnings);
+// console.log(ComputeEarningsNumber);
+console.log(`Maartje has earned € ${computeEarnings}`);
+console.log(`Maartje has earned € ${ComputeEarningsNumber}`);
 
 // Do not change or remove anything below this line
 module.exports = {
--- a/Week2/homework/maartjes-work.js
+++ b/Week2/homework/maartjes-work.js
@@ -45,7 +45,7 @@ const tuesday = [
 const maartjesTasks = monday.concat(tuesday);
 const maartjesHourlyRate = 20;
 
-// Adham's code
+// Start: Adham's code
 const maartjesDurationsHours = []; /* Collect our data in one array */
 maartjesTasks.forEach(element => {
   const durationMinutes = element.duration;
@@ -67,7 +67,7 @@ const earningsNumber = parseFloat(earnings2decimals);
 // console.log(earningsNumber);
 console.log(`Maartje has earned € ${maartjesEarnings}`);
 console.log(`Maartje has earned € ${earningsNumber}`);
-
+// Finish: Adham's code
 // Do not change or remove anything below this line
 module.exports = {
   maartjesTasks,
--- a/Week2/homework/maartjes-work.js
+++ b/Week2/homework/maartjes-work.js
@@ -58,7 +58,6 @@ maartjesTasks.forEach(element => {
 const reducingProcess = (totalToLeft, right) => totalToLeft + right;
 const maartjesTotalDuration = maartjesDurationsHours.reduce(reducingProcess);
 // console.log(maartjesTotalDuration); /* One Number: the total of durations in hours */
-
 const maartjesEarnings = maartjesTotalDuration * maartjesHourlyRate;
 // console.log(earnings); /* One Number: The exactly earnings with lots of decimals */
 // add code to convert `earnings` to a string rounded to two decimals (euro cents)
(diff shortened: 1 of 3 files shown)


Adham-Elias-Botours:
Thanks,
that is exactly what I was trying to do, to give the test what it wants and
to do what I want. But after a while I got confused.

your hint is very helpful.
I will adjust my work tomorrow and resend the pull request.

best regards

Adham-Elias-Botours:
It seems to me that it is not so good to work together in this way. 
Next time I expect to get a list of the variables names and types that must be in your project and also in my work. And then I can code free on my way with that variables names and types. 
This time, I changed the code in the test file , so my work will pass the test. 
I changed it without experience, but it is changed and it test what must be tested.
Best regards.

remarcmij:
I'm sorry that the assignment has confused you. But you can not pass the test by changing the test itself. These 'unit tests' are used in professional projects with multiple team members to ensure that individual pieces of code meet overall, agreed requirements for the project, and continue to do as the project evolves over time. It is not something we invented for just this homework. I suggest that for week 3 you try and pass the tests without altering the test themselves. It can be done, as many students before you have already demonstrated.

Adham-Elias-Botours:
Thanks for the explanation. 
But really I will change nothing any more. 
Let us say : I failed with this homework.  
oke, I accept that and I will keep going on with what I have learned.

remarcmij:
Hi Adham, I haven't look in detail how you computed Maartje's earnings. That is for whoever will review your homework (I'm not that person). I'm sure you will have computed the right answer. My points were about the confusion of the role of the unit tests.

Adham-Elias-Botours:
oke,, so  when we work as a team we need to know what everyone does. now I got your point. 
so , I do not care about that test anymore, and will wait the review from a person. 
thanks

Week2/homework/maartjes-work.js
```diff
@@ -45,21 +45,32 @@ const tuesday = [
...
-function computeEarnings(tasks, hourlyRate) {
-  // Replace this comment and the next line with your code
-  console.log(tasks, hourlyRate);
-}
-
-// eslint-disable-next-line no-unused-vars
-const earnings = computeEarnings(maartjesTasks, maartjesHourlyRate);
-
+// Start: Adham's code
+const maartjesDurationsHours = []; /* Collect our data in one array */
```
yash-kapila: Hi Adham, after going through your discussion with @remarcmij about tests for this homework and reviewing your `maartjes-work.js`, I think I know what's happening and what possibly was the cause of all the pain you had to go through last week.

Firstly, the reason why your homework didn't pass the tests was because you removed the original function `computeEarnings` in which you were supposed to actually add your code to. The unit tests expect `computeEarnings` to be a valid function in the JS file and for it to return a numerical value. So when the test saw you removed the function entirely, that upset it. I think it was its way of revenge from you. 🤷‍♂ 

Secondly, the reason why we have it like this in HYF's homework is because it closely follows to what developers actually do in real life. This is also known as **Test-Driven Development**. That means, before a developer starts writing code to solve a problem, he/she is required to first write unit tests for the code which will be eventually written.

So if you had put all your logic inside the `computeEarnings` function and simply returned `maartjesEarnings` from it, you wouldn't have had that problem. I am not sure why you decided to remove that function though. 🙂

Week2/homework/maartjes-work.js
```diff
@@ -45,21 +45,32 @@ const tuesday = [
...
+maartjesTasks.forEach(element => {
```
yash-kapila: Ah - I wonder why you opted for this approach rather than using `map` and `filter` functions here. There was a hint in the homework itself when it was stated - 

```
- Map the tasks to durations in hours.
- Filter out everything that took less than two hours (i.e., remove from the collection)
```

Week2/homework/maartjes-work.js
```diff
@@ -45,21 +45,32 @@ const tuesday = [
...
+  const durationMinutes = element.duration;
+  const durationHours = durationMinutes / 60;
+  if (durationHours >= 2) {
+    maartjesDurationsHours.push(durationHours);
+  }
+});
+// console.log(maartjesDurationsHours); /* Array of durations in hours, just if the duration is more than 2 hours */
+const reducingProcess = (totalToLeft, right) => totalToLeft + right;
```
yash-kapila: Hi Adham - similar to my comment in the doubling odd numbers exercise, I think we can combine these two operations in a single statement. Something like:

```
const maartjesTotalDuration = maartjesDurationHours.reduce((total, hours) => total + hours);
```

It has got similar advantages as I have mentioned in that comment 🙂

Week2/homework/maartjes-work.js
```diff
@@ -45,21 +45,32 @@ const tuesday = [
...
+// console.log(earningsNumber);
+console.log(`Maartje has earned € ${maartjesEarnings}`);
+console.log(`Maartje has earned € ${earningsNumber}`);
+// Finish: Adham's code
 // Do not change or remove anything below this line
 module.exports = {
   maartjesTasks,
   maartjesHourlyRate,
-  computeEarnings,
+  maartjesEarnings,
```
yash-kapila: This change can be reverted once you work on my above comment 🙂

Week2/test/maartjes-work.test.js
```diff
@@ -1,9 +1,12 @@
...
+  maartjesHourlyRate,
+  maartjesEarnings,
+} = require(`../homework/maartjes-work`);
 
 describe('maartjes_work', () => {
-  test('earnings rounded to euro cents', () => {
-    const earnings = computeEarnings(maartjesTasks, maartjesHourlyRate);
-    const result = earnings.toFixed(2);
-    expect(result).toBe('373.33');
+  test('result -> 373.33', () => {
```
yash-kapila: Ah - this is a 🚫 . We shouldn't be modifying test cases to make our code work. Instead, we should be writing our code in such a way that it passes all the tests. I would suggest that you read about **Test Driven Development** concepts and you might have some answers to the questions you are having around unit tests. 🙂

Week2/homework/map-filter.js
```diff
@@ -1,13 +1,18 @@
...
 
 function doubleOddNumbers(numbers) {
-  // Replace this comment and the next line with your code
-  console.log(numbers);
+  const filterCondition = number => number % 2 !== 0;
+  const filterResult = numbers.filter(filterCondition);
+  console.log(`The Filter Result is: [${filterResult}]`); /* array of odd numbers */
+  const numberDuplicate = number => number * 2;
+  const numbersDuplicate = filterResult.map(numberDuplicate);
+  return numbersDuplicate;
```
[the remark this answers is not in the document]
Adham-Elias-Botours: Thanks Yash,
I tried to divide my steps so I can understand exactly what I am doing.  After a while I should make my code as short as possible with no unwanted variables.

Week2/homework/maartjes-work.js
Adham-Elias-Botours: Thanks Yash,
I believe that  the target of unit tests is to test the code for errors and helps the developers to notice problems if they got  unexpected results such as negative value or too big value. But it is not to test if the writer has followed exactly the steps that some one else wants . If that one knows the steps and the result then no need to let me think about solving it.  In that case just give me a list with all the variables, functions and methods that I need to use to follow the wanted steps, and also make the steps very clear , so no need for me to think but just converting the steps to code. (I mean: something like user instruction in user manual, then I just need to press the correct buttons
I like the question very much it is not like all the others nonsense questions.  But the way how it is corrected makes it much worse than others questions,(I mean the way how the test code it written).
I hoop from HYF in the future : if they want me to use a specific method then do not give such a kind real situation practice, but just  an input like aan object (X) and then ask to use Map() and tell me that the output must be Y. so if I am not using map or not getting that output then sure I have misunderstanding about Map(). And no need for such a test file.
I deleted the Function because the test did not accept it . I tried to give the test what it wants, but always the test was returning to  me with nonsense  output ( the x function is not a function) something like that. 
so I decided to delete everything and keep my won code. 
I do not know when I will go back to that question, I have to do that but it will not be soon. 
Best regards.

Week2/homework/maartjes-work.js
Adham-Elias-Botours: This is my way in life and pc. First collect data, give it the desired case then analyse it. And as what you said it was a hint, and not a conditional step.
 A condition while I am collecting  the data so I pickup just what I need. Filter wil not help in this way.
Unfortunately , almost all the steps sound conditional steps. That is what I got form the Unit Test.

Week2/homework/maartjes-work.js
Adham-Elias-Botours: Yes , I have to consider that seriously . 
Thanks

Week2/test/maartjes-work.test.js
Adham-Elias-Botours: The same longest answer above. 
And Yes I will be back again to modify my code and pass the test , but it is not going to be soon.

Adham-Elias-Botours:
Thanks Yous for your attention while reading my code. 
Best regards.

Week2/homework/maartjes-work.js
yash-kapila: Hi Adham,

I agree that unit tests help developers in catching unexpected results before the code is submitted/shared with others. However, I think you are mistaken in saying that it forces you to follow exactly the steps that someone else wants. Firstly, in real life scenario, it is not someone else who writes unit tests for you. It is the individual developer's responsibility to write tests for their own code. In HackYourFuture, the mentors make it easier for students by writing unit tests for them.

By definition, unit tests are meant to test small units of an application which in a real world can be considered as a function. For example, if I have a function like:

```
const sum = (a,b) => a + b;
```

One will write unit tests for this function such as:

```
describe('testing sum function', () => {
  it('should return 4', () => {
    expect(sum(1,3)).toEqual(4);
  });

  it('should return 10', () => {
    expect(sum(5,5)).toEqual(10);
  });
});
```

No way do these unit tests prevent a developer for implementing their own sum function. All it tests is if I provide my sum function some arguments, do they return a value which is expected or something completely unexpected? I can easily have written the function as:

```
const sum = (num1,num2) => {
  const total = parseInt(num1, 10) + parseInt(num2, 10);
  return total;
};
```

Coming back to the homework, the function being unit tested is `computeEarnings` function.  Test written for this function are in such a way that it expects result to be `373.33` if provided it with `maartjesTasks` and `maartjesHourlyRate`. Again, no way does this test forces you to write code in a specific manner. It will be happy as long as your code returns 373.33 when provided with `maartjesTasks` and `maartjesHourlyRate`.

If I were to copy your code and paste it inside the `computeEarnings` function, it would work just fine and that is what the test expected from you. 🙂 

```
function computeEarnings(maartjesTasks, maartjesHourlyRate) {
  const maartjesDurationsHours = []; /* Collect our data in one array */
  maartjesTasks.forEach(element => {
    const durationMinutes = element.duration;
    const durationHours = durationMinutes / 60;
    if (durationHours >= 2) {
      maartjesDurationsHours.push(durationHours);
    }
  });
  // console.log(maartjesDurationsHours); /* Array of durations in hours, just if the duration is more than 2 hours */
  const reducingProcess = (totalToLeft, right) => totalToLeft + right;
  const maartjesTotalDuration = maartjesDurationsHours.reduce(reducingProcess);
  // console.log(maartjesTotalDuration); /* One Number: the total of durations in hours */
  const maartjesEarnings = maartjesTotalDuration * maartjesHourlyRate;

  return maartjesEarnings;
}

```

I can also rewrite the function like below and tests would still pass

```
  const totalEarnings = tasks
    .map(task => ({ name: task.name, duration: task.duration / 60 })) // transforms minutes to hours
    .filter(time => time.duration >= 2) // filter array task duration equal and greater than 2
    .reduce((acc, task) => acc + task.duration * hourlyRate, 0); // add duration * hourly rate values to the acc

  return totalEarnings;
}
```

As you can see, one can write code in any way one wants and the unit test do not restrict me with the approach to be taken.

As far as revisiting homework is concerned, I would suggest you to speak with Wouter, Fede or Noer about it as they would be able to consult you better. But for now, I won't be able to approve or close this PR. See you tomorrow. 🙂

Adham-Elias-Botours:
Good Saturday, @remarcmij and @yash-kapila ,
Now my homework is 100% Done. And passed the test.  Here is my explanation: 
Something like:
function functionName(parameter1, parameter2) {return somthingElse;}

something lik that will cause ESLint Errors:
 "parameter1 is assigned a value but never used."
 "parameter2 is assigned a value but never used."

So if the Test Unit going to check that Function then the test will not be able to run. Even if the function returns the value that the test wants, the test will not run "fail in running the test, not in the result of the test

That kind of errors make the "Test Unit" not able to run. But on the other hand the Travis build will pass, because the format  has no bugs and the code follows the build rules. 

As a conclusion, I did not changed my own code but , I added few thing to avoid that errors. 
I believe that we  miss important instructions in the section "run test unit" and  "submitting the homework".
I am someone how like to follow the instructions, But if the instruction are not complete and the missed part is not found. Then I throw all the instruction that I have a way, they are useless ".

Best regards.

PR #406, opened 2019-07-09
Homework week2

ShyarK:


Week2/homework/map-filter.js
```diff
@@ -2,7 +2,7 @@
 
 function doubleOddNumbers(numbers) {
   // Replace this comment and the next line with your code
-  console.log(numbers);
+  return numbers.filter(num => num % 2 !== 0).map(el => el * 2);
```
yash-kapila: Hi Shyar - this is fine. One small thing I could suggest is to use proper name like `num` again instead of `el` inside of `map` function.

Week2/homework/maartjes-work.js
```diff
@@ -42,20 +42,33 @@ const tuesday = [
...
+// Choose variable and parameters names that most accurately describe their contents or purpose. When naming an array, use a plural form, e.g. durations. For a single item, use a singular form, e.g. duration. For details, see Naming Conventions.
+// Don't forget to use =>.
+
 const maartjesTasks = monday.concat(tuesday);
 const maartjesHourlyRate = 20;
 
 function computeEarnings(tasks, hourlyRate) {
   // Replace this comment and the next line with your code
-  console.log(tasks, hourlyRate);
+  const twoHoursTasks = tasks.map(el => el.duration / 60).filter(duration => duration >= 2);
```
yash-kapila: Maybe use `task` instead of `el` here

Week2/homework/map-filter.js
ShyarK: @yash-kapila The problem is I created the new branch earlier at this level ``desktop/js2w1/JavaScript2 `` this was my mistake! I should've gone three levels deeper at this point ``desktop/js2w1/JavaScript2/week2/homework ``  then created a new branch.  Will be fine if I remove ``homework-week2`` branch? Then create a new one and change what you have suggested then push everything again?

PR #408, opened 2019-07-10
Homework js2 week2 submission

hnnhack:


Week2/homework/map-filter.js
```diff
@@ -1,8 +1,12 @@
 'use strict';
 
 function doubleOddNumbers(numbers) {
-  // Replace this comment and the next line with your code
+  const oddNumbers = numbers.filter(elem => elem % 2);
+  const doubledNumbers = oddNumbers.map((elem, index, arr) =>
```
yash-kapila: Ah - I don't think we need to write our `map` function like this. Since we have already filtered out odd numbers on line 4 using the `filter` method, when we run the map function on `oddNumbers` array, we are already sure that all the numbers inside it are odd. Thus, we can skip the check `elem % 2 !== 0 ? arr[index] * 2 : elem,` and simply write it as `oddNumbers.map(elem => elem * 2)` 🙂

Week2/homework/maartjes-work.js
```diff
@@ -46,16 +46,21 @@ const maartjesTasks = monday.concat(tuesday);
 const maartjesHourlyRate = 20;
 
 function computeEarnings(tasks, hourlyRate) {
-  // Replace this comment and the next line with your code
-  console.log(tasks, hourlyRate);
+  const makeDurationHours = tasks.map(elem => elem.duration / 60);
+  const TwoHoursOrBigger = makeDurationHours.filter(elem => elem >= 2);
```
yash-kapila: One small thing here - if you remember the variables naming conventions document here - https://github.com/HackYourFuture/fundamentals/blob/master/fundamentals/naming_conventions.md, the variable names should be camelCasing. Thus, `twoHoursOrBigger` instead of `TwoHoursOrBigger`. 🙂

Week2/homework/maartjes-work.js
```diff
@@ -46,16 +46,21 @@ const maartjesTasks = monday.concat(tuesday);
...
+  let sum = 0;
```
yash-kapila: I see you played it safe and didn't go for the bonus round of summing everything using `reduce`. 😉 

I would definitely recommend though to try to use reduce here instead of this big for loop.

PR #411, opened 2019-07-10
add compeleted homework/Adil-js2-week1

musafir928:
add basic html 
add page layout css
add js file


musafir928:
I deleted the old one(No.377) accidentally.

Week1/homework/style.css
```diff
@@ -1 +1,56 @@
...
+  padding-bottom: 40px;
+}
+.book-information {
+  width: calc(50% - 80px);
+  margin: 40px;
+  opacity: 1;
+  overflow: hidden;
+  box-shadow: black 3px 3px 4px;
+}
+.book-information h1 {
```
nickmcmillan: Your `h1` elements already have a class assigned to them (`book-title`) - I'd suggest using that class name here instead.

Week1/homework/style.css
```diff
@@ -1 +1,56 @@
...
+  font-size: 1.5rem;
+  background-color: #976094;
+}
+.book-information p {
```
nickmcmillan: Same goes - `book-language-author`

Week1/homework/app.js
```diff
@@ -4,8 +4,136 @@
...
+      author: 'William Shakespeare',
+    },
+  };
+  // 1.5
+  function DisplayBookInformation(element) {
+    const ul = document.createElement('ul');
+    ul.setAttribute('id', 'bookInformationList');
+    ul.setAttribute('class', 'bookInformationList');
+    // eslint-disable-next-line guard-for-in
+    for (const book in element) {
```
nickmcmillan: Yash mentioned in class that we should avoid using `for in` loops as it may have unexpected results. Try and use Object.keys, Object.values or Object.entries instead.

Week1/homework/app.js
```diff
@@ -4,8 +4,136 @@
...
+      const li = document.createElement('li');
+      li.setAttribute('id', element[book].title);
+      ul.appendChild(li);
+      li.setAttribute('class', 'book-information');
+      const h1 = document.createElement('h1');
+      h1.setAttribute('class', 'book-title');
+      const p = document.createElement('p');
+      p.setAttribute('class', 'book-language-author');
+      h1.innerHTML = element[book].title.toUpperCase();
```
nickmcmillan: If you're not inserting HTML (like in this case here), I'd suggest using `innerText` instead.

Week1/homework/app.js
```diff
@@ -4,8 +4,136 @@
...
+    footprint: './img/footprint.jpg',
+    revitalized_land: './img/revitalized_land.jpg',
+    the_man_who_has_two_graves: './img/the_man_who_has_two_graves.jpg',
+    home_land: './img/home_land.jpg',
+    the_last_theorem_of_fermat: './img/the_last_theorem_of_fermat.jpg',
+    the_old_man_and_the_sea: './img/the_old_man_and_the_sea.jpg',
+    romeo_and_juliet: './img/romeo_and_juliet.jpg',
+  };
+  const bookTitlesArray = Object.keys(bookImages);
+  const addPicture = element => {
```
nickmcmillan: Great that you separated this out into its own function 👍

Week1/homework/app.js
```diff
@@ -4,8 +4,136 @@
...
+      img.setAttribute('src', './img/' + element[index] + '.jpg');
+      img.setAttribute('alt', element[index]);
+      img.setAttribute('class', 'bookImage');
+      document.getElementById(element[index]).appendChild(img);
+    }
+  };
+
+  const main = () => {
+    createUlList(bookTitles);
+    DisplayBookInformation(bookInformationObject);
```
nickmcmillan: An uppercase first letter in your variable name indicates that this variable is a constructor function (created with the `new` keyword). It's a JS naming convention. In this case `DisplayBookInformation` is an ordinary function - so avoid using an uppercase first letter here.

Week1/homework/app.js
```diff
@@ -4,8 +4,136 @@
...
+      liList.innerText = element[i];
+      liList.setAttribute('class', 'book');
+      ulList.appendChild(liList);
+    }
+    document.body.appendChild(ulList);
+  }
+  //  1.4
+  const bookInformationObject = {
+    harry_potter_chamber_secrets: {
+      title: bookTitles[0],
```
nickmcmillan: In the instructions (1.4) it said that the `title` here be the string you entered in the `bookTitles` array. Your solution is certainly a clever way of solving the homework but I'm not sure it's what Yash intended!

Week1/homework/style.css
musafir928: Thank you for your time and nice comment. css line-45 solved;

Week1/homework/app.js
musafir928: I will try it and push the changes.

PR #413, opened 2019-07-10
Js2-week2-homework-Adil-c22

musafir928:
add 3 js files


musafir928:
Ok, thank you for your time and comment. I will fix it in the next homework. In the beginning, Github was joking with me :). I tried to change the branch name, but after changing, everything has become strange. I am addicted to solve that problem and forgot master branch.

PR #415, opened 2019-07-10
homework for week2

Remzi1:


yash-kapila:
Hi @Remzi1 - it seems that you missed submitting the other part of the homework as I currently see only the `doubleOddNumbers` function. Could you please have a look?

PR #416, opened 2019-07-10
Js2w2 mehmet bozdemir week2

mehmet-bozdemir:


Week2/homework/maartjes-work.js
```diff
@@ -46,16 +48,20 @@ const maartjesTasks = monday.concat(tuesday);
 const maartjesHourlyRate = 20;
 
 function computeEarnings(tasks, hourlyRate) {
-  // Replace this comment and the next line with your code
-  console.log(tasks, hourlyRate);
+  const xyz = maartjesTasks.map(time => time.duration / 60).filter(time => time > 2);
+  const qqq = xyz.map(time => time * 20);
+  const reducer = (accumulator, currentValue) => accumulator + currentValue;
+  const sum = qqq.reduce(reducer);
+  return sum;
```
yash-kapila: Hi Mehmet - even though this logic is correct, I am not particularly satisfied with your choice of variable names. If you remember Naming Conventions topic from JS1(https://github.com/HackYourFuture/fundamentals/blob/master/fundamentals/naming_conventions.md), variable names should be nouns describing the value they contain. However, in this case, it is hard to identify what `xyz`, `qqq`, `reducer` could possibly be. Could you please be mindful of choosing variable names next time onwards otherwise you might not get your homework approved. 😉

PR #420, opened 2019-07-12
Adham week03 homework

Adham-Elias-Botours:
I am sending this homework before its time, because I am going to be busy next week.
Best regards

### DIFF
--- a/Week3/homework/step2-2.js
+++ b/Week3/homework/step2-2.js
@@ -4,17 +4,43 @@ function threeFive(startIndex, stopIndex, threeCallback, fiveCallback) {
   const numbers = [];
 
   // Replace this comment and the next line with your code
-  console.log(startIndex, stopIndex, threeCallback, fiveCallback, numbers);
+  let i;
+  for (i = startIndex; i < stopIndex + 1; i++) {
+    numbers.push(i);
+  }
+  let ii;
+  for (ii = 0; ii < numbers.length; ii++) {
+    if (numbers[ii] % 5 === 0 && numbers[ii] % 3 === 0) {
+      console.log(`Hi: I am ${numbers[ii]},\n my index is ${numbers.indexOf(numbers[ii])}`);
+      threeCallback(numbers[ii]);
+      fiveCallback(numbers[ii]);
+      console.log(`Bye.`);
+    } else if (numbers[ii] % 3 === 0) {
+      console.log(`Hi: I am ${numbers[ii]},\n my index is ${numbers.indexOf(numbers[ii])}`);
+      threeCallback(numbers[ii]);
+      console.log(`Bye.`);
+    } else if (numbers[ii] % 5 === 0) {
+      console.log(`Hi: I am ${numbers[ii]},\n my index is ${numbers.indexOf(numbers[ii])}`);
+      fiveCallback(numbers[ii]);
+      console.log(`Bye.`);
+    } else {
+      console.log(
+        `Hi: I am ${numbers[ii]}\n my index is ${numbers.indexOf(
+          numbers[ii],
+        )} . \n I can not be divided on 3 neither on 5 \n Bye.`,
+      );
+    }
+  }
 }
 
 function sayThree(number) {
   // Replace this comment and the next line with your code
-  console.log(number);
+  console.log(`I can be divided on 3 `);
 }
 
 function sayFive(number) {
   // Replace this comment and the next line with your code
-  console.log(number);
+  console.log(`I can be divided on 5 `);
 }
 
 threeFive(10, 15, sayThree, sayFive);
--- a/Week3/homework/step2-2.js
+++ b/Week3/homework/step2-2.js
@@ -35,12 +35,12 @@ function threeFive(startIndex, stopIndex, threeCallback, fiveCallback) {
 
 function sayThree(number) {
   // Replace this comment and the next line with your code
-  console.log(`I can be divided on 3 `);
+  console.log(`Because I am ${number}I can be divided on 3 `);
 }
 
 function sayFive(number) {
   // Replace this comment and the next line with your code
-  console.log(`I can be divided on 5 `);
+  console.log(`Because I am ${number} I can be divided on 5 `);
 }
 
 threeFive(10, 15, sayThree, sayFive);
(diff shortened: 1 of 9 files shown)


Week3/homework/step2-2.js
```diff
@@ -1,20 +1,47 @@
...
   const numbers = [];
 
   // Replace this comment and the next line with your code
-  console.log(startIndex, stopIndex, threeCallback, fiveCallback, numbers);
+  let i;
+  for (i = startIndex; i < stopIndex + 1; i++) {
+    numbers.push(i);
+  }
+  let ii;
+  for (ii = 0; ii < numbers.length; ii++) {
```
yash-kapila: 3 things here:
1. It's better to declare `i` while you are writing the `for` loop rather than declaring it separately as you do right now. This is because we intend to use the variable only inside the loop for iterating over an array and not anywhere outside of it. Thus, it's better to write it as `for(let i = 0; i < arr.length; i++)`.
2. You need not write separate for loops here for creating the `numbers` array and then iterating it for callback logic. You can combine both these actions in a single for loop like
```
  numbers.push(i);
    // and so on
  }
}
```
3. Mathematically speaking, a number which is divisible by 3 and 5 will be divisible by 15 as well. Thus, you can also write this logic as simply `if (numbers[ii] % 15 === 0)` instead of having the `&&` condition.

Week3/homework/step2-3.js
```diff
@@ -1,12 +1,19 @@
...
+// Test PASS
 // Use a 'for' loop
 function repeatStringNumTimesWithFor(str, num) {
   // eslint-disable-next-line prefer-const
   let result = '';
 
-  // Replace this comment and the next line with your code
-  console.log(str, num, result);
+  if (num < 0) {
+    result = '';
```
yash-kapila: 👍 for checking the `num` value in if condition. I would write the same comment from above about the `for` loop though. We need not separate the declaration of `i` outside the for loop. 🙂

Week3/homework/step2-3.js
```diff
@@ -18,8 +25,14 @@ function repeatStringNumTimesWithWhile(str, num) {
...
+  } else {
+    while (num > 0) {
+      result += str;
```
yash-kapila: 👍

Week3/homework/step2-3.js
```diff
@@ -31,8 +44,14 @@ function repeatStringNumTimesWithDoWhile(str, num) {
   // eslint-disable-next-line prefer-const
   let result = '';
 
-  // Replace this comment and the next line with your code
-  console.log(str, num, result);
+  if (num <= 0) {
+    result = '';
+  } else {
+    do {
+      result += str;
```
yash-kapila: 👍

Week3/homework/step2-4.js
```diff
@@ -1,10 +1,13 @@
 'use strict';
 
-function Dog() {
-  // add your code here
+// Test PASS
+function Dog(name, color, numLegs) {
+  this.name = name;
+  this.color = color;
```
yash-kapila: 👍

Week3/homework/step2-5.js
```diff
@@ -1,11 +1,16 @@
...
 function multiplyAll(arr) {
   // eslint-disable-next-line
-  let product = 1;
+    let product = 1;
 
-  // Replace this comment and the next line with your code
-  console.log(arr, product);
+  function oneTime(element) {
+    product = product * element;
+    return product;
```
yash-kapila: 👍. However, returning `product` here is redundant as `forEach` doesn't really return anything.

Week3/homework/step2-6.js
```diff
@@ -1,16 +1,18 @@
...
-  console.log(arr);
+  return arr.reduce((firstLeft, nextRight) => firstLeft.concat(nextRight));
 }
 
 function flattenArray3d(arr) {
-  // Replace this comment and the next line with your code
-  console.log(arr);
+  const firstLevel = arr.reduce((firstLeft, nextRight) => firstLeft.concat(nextRight));
+  const secondLevel = firstLevel.reduce((firstLeft, nextRight) => firstLeft.concat(nextRight));
+  return secondLevel;
```
yash-kapila: Very well. 👍 

If you would like to think about it further, try and search on google how you could implement it using `recursion`.

Week3/homework/step2-7.js
```diff
@@ -21,3 +24,10 @@ f2(y);
 console.log(y);
 
 // Add your explanation as a comment here
+/** ******
+ * The answer in this link:
+https://www.w3schools.com/js/js_function_parameters.asp
+Exactly in these 2 titles:
```
yash-kapila: I am not too convinced with the explanation you've provided here. Firstly, I would have liked if you would have tried to explain it in your words rather than simply stating what is written in `w3schools`. Secondly, IMO, the statement `arguments are passed by value, objects are passed by reference` is not the most reasonable explanation for this because even objects can be function arguments. 🙂

Week3/homework/step3-bonus.js
```diff
@@ -1,10 +1,12 @@
 'use strict';
 
+// Test PASS
 const values = ['a', 'b', 'c', 'd', 'a', 'e', 'f', 'c'];
 
 function makeUnique(arr) {
   // Replace this comment and the next line with your code
-  console.log(arr);
+  const uniques = [...new Set(arr)];
+  return uniques;
```
yash-kapila: 👍

Week3/homework/step3.js
```diff
@@ -1,14 +1,18 @@
 'use strict';
 
+// Test PASS
 function createBase(base) {
-  // Replace this comment and the next line with your code
-  console.log(base);
+  return function(add) {
+    return base + add;
+  };
```

Week3/homework/step2-7.js
Adham-Elias-Botours: recently I am watching this:
https://youtu.be/Bv_5Zv5c-Ts
And I got the idea, The relationship between JavaScript and the memory.

Week3/homework/step2-5.js
Adham-Elias-Botours: yes, your point is important. I have to read more about nested ForEach methods.

Adham-Elias-Botours:
Thanks @yash-kapila
I like your way in explaining the information. For me it is very understandable. 
Thanks alot for your effort in JS2.
Best regards

PR #430, opened 2019-07-15
Week3

mayqf:


yash-kapila:
Hi @mayqf - it seems that there are some unit tests failing for your homework. Could you please have a look and correct the same? Also, did you checkout the `master` branch first before creating a new branch for this week's homework? If yes, you don't need to delete the files for week1 and week2's homework. If not, I would request you to please resubmit your homework but after creating a new branch from `master`. 🙂

mayqf:
Hey Jash. Many thanks for your note. I had test problem at step2-2. I could not still solve it. It works and results are correct. If I manage to do, I will do PR again. Yes, I think I forgot to checkout master. There is not any problem at Week1&Week2 . Probably, I forgot it.  I did not select week3  branch

Week3/homework/step2-1.js
```diff
@@ -1,15 +1,13 @@
...
-  // Replace this comment and the next line with your code
-  console.log(func);
-}
-
 function bar() {
   console.log('Hello, I am bar!');
 }
 
+function foo(func) {
+  return func();
```
nickmcmillan: Nice. No need to `return` though - seeing as you're not using the return value of calling `foo`

Week3/homework/step2-2.js
```diff
@@ -2,19 +2,29 @@
 
 function threeFive(startIndex, stopIndex, threeCallback, fiveCallback) {
   const numbers = [];
-
-  // Replace this comment and the next line with your code
-  console.log(startIndex, stopIndex, threeCallback, fiveCallback, numbers);
+  for (let i = startIndex; i <= stopIndex; i++) {
```
nickmcmillan: Great that you did this all in one loop. Not sure why you're creating the `numbers` array and pushing into it though? It doesn't look like you're using this array anywhere?

Week3/homework/step2-2.js
```diff
@@ -2,19 +2,29 @@
...
+    numbers.push(i);
+    const number = i;
+    const isNumberDivisibleBy3 = number % 3 === 0;
+    const isNumberDivisibleBy5 = number % 5 === 0;
+    const isNumberDivisibleBy15 = number % 15 === 0;
+    if (isNumberDivisibleBy3) {
+      console.log(threeCallback(number));
```
nickmcmillan: It'd be better to just call the callback function - not to console log it.
Instead you'd put the console log within the function.

Week3/homework/step2-3.js
```diff
@@ -4,10 +4,9 @@
 function repeatStringNumTimesWithFor(str, num) {
   // eslint-disable-next-line prefer-const
   let result = '';
-
-  // Replace this comment and the next line with your code
-  console.log(str, num, result);
-
+  for (let i = 0; i < num; i++) {
+    result += str;
```
nickmcmillan: 👍 nice and tidy

Week3/homework/step2-3.js
```diff
@@ -17,22 +16,22 @@ console.log('for', repeatStringNumTimesWithFor('abc', 3));
...
 function repeatStringNumTimesWithDoWhile(str, num) {
   // eslint-disable-next-line prefer-const
   let result = '';
-
-  // Replace this comment and the next line with your code
-  console.log(str, num, result);
+  do {
+    result += str;
+    num--;
+  } while (num > 0);
```
nickmcmillan: Your `while` and `do while` loops here are great.

Week3/homework/step2-4.js
```diff
@@ -1,10 +1,13 @@
 'use strict';
 
 function Dog() {
-  // add your code here
+  this.name = 'Golden';
+  this.color = 'gold';
+  this.numLegs = 4;
```
nickmcmillan: This is a bit limiting. You've hard-coded Golden, gold, and 4, into the Dog constructor. That means any hound you create will always have these values.

const hound = new Dog('Golden', 'gold', 4);

Week3/homework/step2-5.js
```diff
@@ -3,10 +3,11 @@
 function multiplyAll(arr) {
   // eslint-disable-next-line
   let product = 1;
-
-  // Replace this comment and the next line with your code
-  console.log(arr, product);
-
+  arr.forEach(element => {
+    element.forEach(item => {
+      product *= item;
```
nickmcmillan: Good! I like that you've used `forEach` here. Maybe the param names could be more descriptive - as `element` and `item` are a bit vague to me. Maybe `innerArray`  and `innerArrayItem`.

Week3/homework/step2-6.js
```diff
@@ -4,13 +4,25 @@ const arr2d = [[1, 2], [3, 4], [5, 6]];
...
 }
 
 function flattenArray3d(arr) {
-  // Replace this comment and the next line with your code
-  console.log(arr);
+  const newArray2 = [];
+  arr.forEach(element => {
+    element.forEach(child =>
+      child.forEach(item => {
+        newArray2.push(item);
```
nickmcmillan: Nice work! Though I'm usually a little wary when I see more than 2 levels of nested loops - the fewer loops the better. If you'd like you could re-think this using recursion?

Week3/homework/step2-7.js
```diff
@@ -20,4 +24,7 @@ f2(y);
 
 console.log(y);
 
-// Add your explanation as a comment here
+// As Jash explained in this and earlier lessons,
+// 1 st snippet writes 9 and 2n d writes { x: 10,}. JS calls by value, but in object and other complex data types with reference.
+// we used 9 and made val value 10, but it did not change x variable 's value. Complex data types such as object take place in memory.
+// So we modified the value of const y which has place in the memory,reference. So we are manipulating the memory place.
```
nickmcmillan: Yep, good explanation! Sounds like you've got a good grasp on primitive vs complex types.

Week3/homework/step3.js
```diff
@@ -1,10 +1,10 @@
 'use strict';
 
 function createBase(base) {
-  // Replace this comment and the next line with your code
-  console.log(base);
+  return function(y) {
+    return base + y;
```
nickmcmillan: Nice! You could even try using an arrow function here, which would reduce the amount of times you have to write `return` - and even reduce it onto one line.

Week3/homework/step2-2.js
mayqf: yes. we do not need it

Week3/homework/step2-1.js
mayqf: Ok

mayqf: we do not return value, only console.log()

Week3/homework/step2-2.js
mayqf: ok

Week3/homework/step2-4.js
mayqf: this.name=name
this.color=color
this.numLegs=Legs
const hound = new Dog('Golden', 'gold', 4);

Week3/homework/step2-5.js
mayqf: innerArrayItem is great name

Week3/homework/step2-6.js
mayqf: Yes, I will keep on seperate, not nested

Week3/homework/step3.js
mayqf: y=>base+y

PR #432, opened 2019-07-15
Week3 homework/Shyar

ShyarK:


Week3/homework/step2-1.js
```diff
@@ -2,8 +2,7 @@
 
 function foo(func) {
   // What to do here?
-  // Replace this comment and the next line with your code
-  console.log(func);
+  return func();
```

Week3/homework/step2-2.js
```diff
@@ -2,19 +2,27 @@
...
+  }
+  numbers.forEach(num => {
+    if (num % 15 === 0) {
+      threeCallback(num);
+      fiveCallback(num);
+    } else if (num % 3 === 0) {
+      threeCallback(num);
+    } else if (num % 5 === 0) {
+      fiveCallback(num);
+    }
```
yash-kapila: I think you need not write separate for loops here for creating the numbers array and then iterating it for callback logic. You can combine both these actions in a single for loop like
```
  numbers.push(i);
    // and so on
  }
}
```

Week3/homework/step2-3.js
```diff
@@ -4,10 +4,9 @@
 function repeatStringNumTimesWithFor(str, num) {
   // eslint-disable-next-line prefer-const
   let result = '';
-
-  // Replace this comment and the next line with your code
-  console.log(str, num, result);
-
+  for (let i = 0; i < num; i++) {
```
yash-kapila: Correct. However, it would have been nice if you had an `if` condition to check for values of `num` less than or equal to 0. 🙂

Week3/homework/step2-3.js
```diff
@@ -17,10 +16,10 @@ console.log('for', repeatStringNumTimesWithFor('abc', 3));
...
   // eslint-disable-next-line prefer-const
   let result = '';
-
-  // Replace this comment and the next line with your code
-  console.log(str, num, result);
-
+  while (num > 0) {
+    num--;
+    result += str;
+  }
```
yash-kapila: The same comment as above `for` loop 🙂

Week3/homework/step2-3.js
```diff
@@ -30,16 +29,24 @@ console.log('while', repeatStringNumTimesWithWhile('abc', 3));
...
+      result += str;
+    } else if (num === 2) {
+      result += str;
+    } else if (num === 1) {
+      result += str;
+    } else {
+      result = '';
+    }
+    num--;
+  } while (num > 0);
```
yash-kapila: Ah. I think this is not completely right. You are presuming that the value of `num` would always be 3 with this logic. But it is not in your control, is it? It can be of any value. Could you revisit this please?

Week3/homework/step2-4.js
```diff
@@ -1,7 +1,9 @@
 'use strict';
 
 function Dog() {
-  // add your code here
+  this.name = 'Fino';
+  this.color = 'white';
+  this.numLegs = 4;
```
yash-kapila: Ah. This is not completely right. You are correct in the way of creating a constructor function however, the values that you have provided to it i.e `Fino`, `white` and `4` are all hardcoded. This means, I can't use this constructor function to create a new dog object. The right way of doing it will be using function parameters:
```
  this.name = name;
  this.color = color;
  this.numLegs = legs;
}
```

```
// and so on
```

Week3/homework/step2-5.js
```diff
@@ -3,10 +3,7 @@
 function multiplyAll(arr) {
   // eslint-disable-next-line
   let product = 1;
-
-  // Replace this comment and the next line with your code
-  console.log(arr, product);
-
+  product = [].concat(...arr).reduce((acc, curr) => acc * curr);
```

Week3/homework/step2-6.js
```diff
@@ -4,13 +4,11 @@ const arr2d = [[1, 2], [3, 4], [5, 6]];
...
 function flattenArray2d(arr) {
-  // Replace this comment and the next line with your code
-  console.log(arr);
+  return arr.reduce((acc, curr) => [...acc, ...curr]);
 }
 
 function flattenArray3d(arr) {
-  // Replace this comment and the next line with your code
-  console.log(arr);
+  return [].concat(...arr).reduce((acc, curr) => [...acc, ...curr]);
```
yash-kapila: 👍 But, if you like to take a step further, try to search on google on how you can implement it using `recursion`.

Week3/homework/step2-7.js
```diff
@@ -21,3 +21,26 @@ f2(y);
...
+
+// 1- variable x is a primitive value because it is a type of number, stored on the top of the stack somewhere in the memory
+// and when we pass it as an argument to val parameter, we actually copy its value and kind of past it as a value for val.
+// So we have two different variable val and x, and when we increase val by one, we only change the value of val variable and not x variable.
+// val is 10 and x is 9
+
+// 2-We created an object and we labeled y. Y is a pointer stored on the stack and points to the actual object which is a reference type stored on the heap.
+// And if we pass variable y as an argument to the parameter val, we don't actually copy its value
+// but we create a new pointer on the stack and this pointer points to the same object in the heap
+// Therefore if we increase val by one in the f2 function we change the same object value and the value of y.
```

Week3/homework/step3-bonus.js
```diff
@@ -1,12 +1,10 @@
 'use strict';
 
 const values = ['a', 'b', 'c', 'd', 'a', 'e', 'f', 'c'];
-
 function makeUnique(arr) {
-  // Replace this comment and the next line with your code
-  console.log(arr);
+  const uniqValArr = arr.filter((el, i, array) => array.indexOf(el) === i);
```
yash-kapila: Just a suggestion here - try and read about a new concept in ES6 which is called Sets and see if you can use it to make this code shorter. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set

Week3/homework/step3.js
```diff
@@ -1,8 +1,9 @@
 'use strict';
 
 function createBase(base) {
-  // Replace this comment and the next line with your code
-  console.log(base);
+  return function(num) {
+    return base + num;
```

PR #434, opened 2019-07-16
js2 homework for week3 submission

hnnhack:
This is from niyas.

### DIFF
--- a/Week3/homework/step2-1.js
+++ b/Week3/homework/step2-1.js
@@ -3,7 +3,7 @@
 function foo(func) {
   // What to do here?
   // Replace this comment and the next line with your code
-  console.log(func);
+  return func();
 }
 
 function bar() {
--- a/Week3/homework/step2-2.js
+++ b/Week3/homework/step2-2.js
@@ -2,19 +2,29 @@
 
 function threeFive(startIndex, stopIndex, threeCallback, fiveCallback) {
   const numbers = [];
-
   // Replace this comment and the next line with your code
+  for (let i = startIndex; i <= stopIndex; i++) {
+    numbers.push(i);
+    if (i % 15 === 0) {
+      threeCallback(i);
+      fiveCallback(i);
+    } else if (i % 5 === 0) {
+      fiveCallback(i);
+    } else if (i % 3 === 0) {
+      threeCallback(i);
+    }
+  }
   console.log(startIndex, stopIndex, threeCallback, fiveCallback, numbers);
 }
 
 function sayThree(number) {
   // Replace this comment and the next line with your code
-  console.log(number);
+  console.log(number + ' is divisible by 3');
 }
 
 function sayFive(number) {
   // Replace this comment and the next line with your code
-  console.log(number);
+  console.log(number + ' is divisible by 5');
 }
 
 threeFive(10, 15, sayThree, sayFive);
--- a/Week3/homework/step2-3.js
+++ b/Week3/homework/step2-3.js
@@ -6,9 +6,13 @@ function repeatStringNumTimesWithFor(str, num) {
   let result = '';
 
   // Replace this comment and the next line with your code
-  console.log(str, num, result);
-
-  return result;
+  if (num > 0) {
+    for (let i = 0; i < num; i++) {
+      result += str;
+    }
+    return result;
+  }
+  return '';
 }
 
 console.log('for', repeatStringNumTimesWithFor('abc', 3));
@@ -19,8 +23,10 @@ function repeatStringNumTimesWithWhile(str, num) {
   let result = '';
 
   // Replace this comment and the next line with your code
-  console.log(str, num, result);
-
+  while (num > 0) {
+    result += str;
+    num--;
+  }
   return result;
 }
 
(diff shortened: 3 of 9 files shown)


Week3/homework/step2-1.js
```diff
@@ -3,7 +3,7 @@
 function foo(func) {
   // What to do here?
   // Replace this comment and the next line with your code
-  console.log(func);
+  return func();
```

Week3/homework/step2-2.js
```diff
@@ -2,19 +2,29 @@
...
+    numbers.push(i);
+    if (i % 15 === 0) {
+      threeCallback(i);
+      fiveCallback(i);
+    } else if (i % 5 === 0) {
+      fiveCallback(i);
+    } else if (i % 3 === 0) {
+      threeCallback(i);
+    }
+  }
```
yash-kapila: 👍

Week3/homework/step2-3.js
```diff
@@ -6,9 +6,13 @@ function repeatStringNumTimesWithFor(str, num) {
...
-
-  return result;
+  if (num > 0) {
+    for (let i = 0; i < num; i++) {
+      result += str;
+    }
+    return result;
+  }
+  return '';
 }
```
yash-kapila: 👍 but you can also simply return `result` at the end of the function and need not return anything then inside `if` block. It's just one less line of code 😄 

```
  let result = '';
  if (num > 0) {
    for (let i = 0; i < num; i++) {
      result += str;
    }
  }
  return result;
```

Week3/homework/step2-3.js
```diff
@@ -19,8 +23,10 @@ function repeatStringNumTimesWithWhile(str, num) {
   let result = '';
 
   // Replace this comment and the next line with your code
-  console.log(str, num, result);
-
+  while (num > 0) {
+    result += str;
+    num--;
+  }
```
yash-kapila: Ah. You missed adding the logic of checking `num` like you did in `for` loop here.

Week3/homework/step2-3.js
```diff
@@ -32,8 +38,10 @@ function repeatStringNumTimesWithDoWhile(str, num) {
   let result = '';
 
   // Replace this comment and the next line with your code
-  console.log(str, num, result);
-
+  do {
+    result += str;
+    num--;
+  } while (num > 0);
```
yash-kapila: Missing logic here as well 🙂

Week3/homework/step2-4.js
```diff
@@ -1,7 +1,9 @@
 'use strict';
 
 function Dog() {
-  // add your code here
+  this.name = 'Johnny';
+  this.color = 'whiite';
+  this.numLegs = 4;
 }
```
yash-kapila: Ah. This is not completely right. You are correct in the way of creating a constructor function however, the values that you have provided to it i.e `Johnny`, `white` and `4` are all hardcoded. This means, I can't use this constructor function to create a new dog object. The right way of doing it will be using function parameters:
```
  this.name = name;
  this.color = color;
  this.numLegs = legs;
}
```

```
// and so on
```

Week3/homework/step2-5.js
```diff
@@ -5,8 +5,11 @@ function multiplyAll(arr) {
   let product = 1;
 
   // Replace this comment and the next line with your code
-  console.log(arr, product);
-
+  for (let i = 0; i < arr.length; i++) {
+    for (let j = 0; j < arr[i].length; j++) {
+      product *= arr[i][j];
+    }
+  }
```

Week3/homework/step2-6.js
```diff
@@ -5,12 +5,26 @@ const arr3d = [[[1, 2], [3, 4]], [[5, 6], [7, 8]]];
...
-  console.log(arr);
+  const flattenedArr = [];
+  for (let i = 0; i < arr.length; i++) {
+    for (let j = 0; j < arr[i].length; j++) {
+      for (let k = 0; k < arr[i][j].length; k++) {
+        flattenedArr.push(arr[i][j][k]);
+      }
+    }
+  }
+  return flattenedArr;
```
yash-kapila: If you like to take a step further, try to search on google on how you can implement it using `recursion`. 🙂

Week3/homework/step2-7.js
```diff
@@ -21,3 +21,5 @@ f2(y);
 console.log(y);
 
 // Add your explanation as a comment here
+// JavaScript is call by value.
+// However, in case if object, the value being passed is the reference.
```
yash-kapila: Haha. I think you just copied what I mentioned during the class here. 😄

Week3/homework/step3-bonus.js
```diff
@@ -4,7 +4,8 @@ const values = ['a', 'b', 'c', 'd', 'a', 'e', 'f', 'c'];
 
 function makeUnique(arr) {
   // Replace this comment and the next line with your code
-  console.log(arr);
+  const uniqueArray = arr.filter((item, pos) => arr.indexOf(item) === pos);
+  return uniqueArray;
```
yash-kapila: Very well. This logic works fine but it seems a bit too much. Try and read about a new concept in ES6 which is called Sets and see if you can use it to make this code shorter. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set

Week3/homework/step3.js
```diff
@@ -2,7 +2,9 @@
 
 function createBase(base) {
   // Replace this comment and the next line with your code
-  console.log(base);
+  return function(number) {
+    return number + base;
```

PR #436, opened 2019-07-16
Homework/week2

hnnhack:
homework week 2 from niyas. After review and corrected notes and comments

### DIFF
--- a/Week2/homework/maartjes-work.js
+++ b/Week2/homework/maartjes-work.js
@@ -47,7 +47,11 @@ const maartjesHourlyRate = 20;
 
 function computeEarnings(tasks, hourlyRate) {
   // Replace this comment and the next line with your code
-  console.log(tasks, hourlyRate);
+  // console.log(tasks, hourlyRate);
+  const makeDurationHours = tasks.map(elem => elem.duration / 60);
+  const twoHoursOrBigger = makeDurationHours.filter(elem => elem >= 2);
+  const sum = twoHoursOrBigger.reduce((acc, elem) => acc + elem * hourlyRate, 0);
+  return sum;
 }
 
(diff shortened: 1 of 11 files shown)


Week2/homework/map-filter.js
```diff
@@ -3,6 +3,7 @@
 function doubleOddNumbers(numbers) {
   // Replace this comment and the next line with your code
   console.log(numbers);
+  return numbers.filter(elem => elem % 2).map(elem => elem * 2);
```

Week2/homework/maartjes-work.js
```diff
@@ -47,15 +47,21 @@ const maartjesHourlyRate = 20;
 
 function computeEarnings(tasks, hourlyRate) {
   // Replace this comment and the next line with your code
-  console.log(tasks, hourlyRate);
+  // console.log(tasks, hourlyRate);
+  const makeDurationHours = tasks.map(elem => elem.duration / 60);
+  const twoHoursOrBigger = makeDurationHours.filter(elem => elem >= 2);
+  const sum = twoHoursOrBigger.reduce((acc, elem) => acc + elem * hourlyRate, 0);
+  return sum;
```
yash-kapila: Very well. Later as you become more experienced with programming, you would like to make use of a concept called `chaining` to further reduce this function. It would be written something like:
```
  return tasks
    .map(elem => elem.duration / 60)
    .filter(elem => elem >= 2)
    .reduce((acc, elem) => acc + elem * hourlyRate, 0);
}
```

Since this is a pure function, anyone who wishes to make use of the function then need not really worry about the underlying logic of it. All we care is that if we supply this function a list of tasks with an hourly rate, I would get the total earnings. 🙂

PR #437, opened 2019-07-17
JS2-Week3 Homework/ Farzanna

Farzanna:


Week3/homework/step2-7.js
```diff
@@ -21,3 +26,7 @@ f2(y);
 console.log(y);
 
 // Add your explanation as a comment here
+//  first x is an const variable that doesnt changed during the function
+//  second X is inside an object y and it can change
+// eslint-disable-next-line prettier/prettier
+//  my explanation isn't finished, I will explain it more later....
```
nickmcmillan: This is a good start. Have a read through some of the answers in this link and then try come back and explain a little more!
https://stackoverflow.com/questions/6605640/javascript-by-reference-vs-by-value

PR #439, opened 2019-07-17
add js-2 week3 homework files/Adil

musafir928:


Week3/homework/step2-2.js
```diff
@@ -2,19 +2,27 @@
...
+  numbers.forEach(number => {
+    if (number % 15 === 0) {
+      threeCallback(number);
+      fiveCallback(number);
+    } else if (number % 3 === 0) {
+      threeCallback(number);
+    } else if (number % 5 === 0) {
+      fiveCallback(number);
+    }
+  });
```
nickmcmillan: Nailed it 👍 Good work.

Week3/homework/step2-2.js
```diff
@@ -2,19 +2,27 @@
...
 }
 
 function sayThree(number) {
-  // Replace this comment and the next line with your code
-  console.log(number);
+  console.log(number + ': divisible by 3');
```
nickmcmillan: You might be interested to know that inside console.log you can separate parameters by commas. So you could write it like this, 
`console.log(number, ': divisible by 3');`
I often find using commas makes it easier - so you can log multiple things on a single line.

Week3/homework/step2-3.js
```diff
@@ -4,10 +4,11 @@
 function repeatStringNumTimesWithFor(str, num) {
   // eslint-disable-next-line prefer-const
   let result = '';
-
-  // Replace this comment and the next line with your code
-  console.log(str, num, result);
-
+  if (num >= 1) {
```
nickmcmillan: Good idea to check the value of the number first 👍

Week3/homework/step2-3.js
```diff
@@ -30,10 +33,12 @@ console.log('while', repeatStringNumTimesWithWhile('abc', 3));
...
-
-  // Replace this comment and the next line with your code
-  console.log(str, num, result);
-
+  if (num > 0) {
+    do {
+      result += str;
+      num--;
+    } while (num);
+  }
```
nickmcmillan: Nice. Both your `while` and `do while` loops are spot on.

Week3/homework/step2-4.js
```diff
@@ -2,6 +2,10 @@
 
 function Dog() {
   // add your code here
+  this.name = 'kuchuk';
+  this.color = 'black and white';
+  this.food = 'meat';
+  this.numLegs = 4;
```
nickmcmillan: This is a bit limiting. You've hard-coded 'kuchuk', 'black and white', 'meat', and 4, into the Dog constructor. That means any hound you create will always have these values.

The idea here is that Dog should take name, color, and numLegs as parameters (and also food if you'd like to add that too) - which you'd pass in when you create it. Like this;

const hound = new Dog('kuchuk', 'black and white', 'meat', 4);

Week3/homework/step2-5.js
```diff
@@ -3,10 +3,11 @@
 function multiplyAll(arr) {
   // eslint-disable-next-line
   let product = 1;
-
-  // Replace this comment and the next line with your code
-  console.log(arr, product);
-
+  for (let x = 0; x < arr.length; x++) {
+    for (let y = 0; y < arr[x].length; y++) {
+      product *= arr[x][y];
```
nickmcmillan: Cool, good solution.

Week3/homework/step2-6.js
```diff
@@ -4,13 +4,17 @@ const arr2d = [[1, 2], [3, 4], [5, 6]];
...
+    .map(element => +element);
 }
 
 function flattenArray3d(arr) {
-  // Replace this comment and the next line with your code
-  console.log(arr);
+  return arr
+    .toString()
+    .split(',')
+    .map(element => +element);
```
nickmcmillan: Interesting! I wouldn't have ever thought to use `toString` to flatten the array. I guess this would work with any level of nested arrays too, so it's a very creative solution.

This does have limitations though - for instance it wouldn't work if one of the items in the array happened to be a function, and it might have performance penalties for larger arrays.

@yash-kapila curious on your thoughts here too?

Week3/homework/step2-7.js
```diff
@@ -21,3 +21,11 @@ f2(y);
 console.log(y);
 
 // Add your explanation as a comment here
+/* My explain:
+  when a primitive type assigned to a variable, it means: variable as a container of that value
+ but, in a non-primitive type: the variable point to the location, not actually as a container of that value.
+ so, in this two function above, when we call the function f1(x), we just copy x's value to the val, so after running function, x not changed.
+ but, in f2(y), when the function get y as a parameter val also gets the location(address) of what y gets. Both y and val pointed to same location.
+ so, changing one of them means changing the another. 
+ This is a main difference between primitive type and none-primitive type in javascript   data types.
```
nickmcmillan: Yep, you nailed it. Nice work 👍

Week3/homework/step3-bonus.js
```diff
@@ -3,8 +3,7 @@
 const values = ['a', 'b', 'c', 'd', 'a', 'e', 'f', 'c'];
 
 function makeUnique(arr) {
-  // Replace this comment and the next line with your code
-  console.log(arr);
+  return [...new Set(arr)];
```
nickmcmillan: Very good use of `Set`!

Week3/homework/step3.js
```diff
@@ -1,8 +1,9 @@
 'use strict';
 
 function createBase(base) {
-  // Replace this comment and the next line with your code
-  console.log(base);
+  return function(num) {
+    return num + base;
```
nickmcmillan: Nice! You could even try using an arrow function here, which would reduce the amount of times you have to write return - and even reduce it onto one line.

Week3/homework/step2-6.js
musafir928: Thank you for your time and nice comment. I will try to fix it with other solutions.

Week3/homework/step2-4.js
musafir928: I've added parameters to the function.

Week3/homework/step2-4.js
```diff
@@ -1,14 +1,14 @@
...
-  this.food = 'meat';
-  this.numLegs = 4;
+  this.name = name;
+  this.color = color;
+  this.food = food;
+  this.numLegs = numLegs;
 }
 
-const hound = new Dog();
+const hound = new Dog('kuchuk', 'black and white', 'meat', 4);
```
nickmcmillan: Nice work!

PR #440, opened 2019-07-17
Js2 Week3 Hw -- Mehmet Ergorgec

Mehmet-Ergorgec:


Week3/homework/step2-1.js
```diff
@@ -3,6 +3,7 @@
 function foo(func) {
   // What to do here?
   // Replace this comment and the next line with your code
+  func();
```
nickmcmillan: Perfect 👍

Week3/homework/step2-2.js
```diff
@@ -2,21 +2,32 @@
...
+  numbers.forEach(number => {
+    if (number % 15 === 0) {
+      threeCallback(number);
+      fiveCallback(number);
+    } else if (number % 3 === 0) {
+      threeCallback(number);
+    } else if (number % 5 === 0) {
+      fiveCallback(number);
+    }
+  });
```
nickmcmillan: Nice and tidy 👍

Week3/homework/step2-3.js
```diff
@@ -7,7 +7,9 @@ function repeatStringNumTimesWithFor(str, num) {
 
   // Replace this comment and the next line with your code
   console.log(str, num, result);
-
+  for (let i = 0; i < num; i++) {
+    result += `${str}`;
```
nickmcmillan: No need to use a template literal here, this would work just the same because `str` is already a string;
`result += str;`

Week3/homework/step2-3.js
```diff
@@ -20,7 +22,10 @@ function repeatStringNumTimesWithWhile(str, num) {
 
   // Replace this comment and the next line with your code
   console.log(str, num, result);
-
+  while (num > 0) {
```
nickmcmillan: Nice, but don't forget to remove the `console.log`s as per the instructions

Week3/homework/step2-3.js
```diff
@@ -33,7 +38,14 @@ function repeatStringNumTimesWithDoWhile(str, num) {
...
   console.log(str, num, result);
-
+  do {
+    result += `${str}`;
+    num--;
+    if (num < 0) {
+      result = '';
+    }
+  } while (num > 0);
+  console.log(str, num, result);
```
nickmcmillan: Good work on the `while` and `do while` loops 👍

Week3/homework/step2-4.js
```diff
@@ -2,6 +2,9 @@
 
 function Dog() {
   // add your code here
+  this.name = 'Max';
+  this.color = 'Grey';
+  this.numLegs = 4;
```
nickmcmillan: You've hard-coded Max, Grey, and 4, into the Dog constructor. That means any hound you create will always have these values.

const hound = new Dog('Max', 'Grey', 4);

Week3/homework/step2-5.js
```diff
@@ -5,8 +5,12 @@ function multiplyAll(arr) {
   let product = 1;
 
   // Replace this comment and the next line with your code
+  for (let i = 0; i < arr.length; i++) {
+    for (let j = 0; j < arr[i].length; j++) {
+      product = product * arr[i][j];
```
nickmcmillan: Good.

Week3/homework/step2-6.js
```diff
@@ -6,11 +6,15 @@ const arr3d = [[[1, 2], [3, 4]], [[5, 6], [7, 8]]];
...
   console.log(arr);
+  return arr.reduce((acc, element) => acc.concat(element), []);
 }
 
 function flattenArray3d(arr) {
   // Replace this comment and the next line with your code
   console.log(arr);
+  return arr
+    .reduce((acc, element) => acc.concat(element), [])
+    .reduce((acc, element) => acc.concat(element), []);
```
nickmcmillan: Really good usage of `reduce` in this file, happy to see you use it, this is a good solution.

Week3/homework/step2-7.js
```diff
@@ -21,3 +21,29 @@ f2(y);
 console.log(y);
 
 // Add your explanation as a comment here
+// Golden rule is: Primitive data types in Javascript like Number, String, Boolean always pass by value
+// whereas Objects are pass by reference.
```
nickmcmillan: Don't forget Arrays!

Week3/homework/step2-7.js
```diff
@@ -21,3 +21,29 @@ f2(y);
...
+// This example shows us "Pass by value in Js". Because;
+// console.log(Object.prototype.toString.call(9) === '[object Number]'); returns true
+// typeof 9 is Number which is primitive.
+
+// const x = 9
+// In this code we put x value somewhere in the memory heap.
+// When We call f1(x), this function uses that variable which is in the memory heap.Actual x will not change.
+// So we are just passing value of x inside of function.
+// functions have their own scope and execution context So x will be 10 inside of function.
+// At the end we are logging x. not f1(x). They are both different scope and execution context.
```
nickmcmillan: Yep, well explained

Week3/homework/step2-7.js
```diff
@@ -21,3 +21,29 @@ f2(y);
...
+// Second Example
+
+// This example shows us "Pass by reference in Js". Because;
+// console.log(Object.prototype.toString.call(y) === '[object Object]'); returns true
+// means typeof y is Object
+
+// When we want to update/do sth. inside of function with objects, actual object will change.
+// Because parameter of f2(y) which is y and const y are both referencing the same object.
+// If we change reference object both of them will be affected from that code.
+// console.log(y) returns {x: 10}
```
nickmcmillan: I think you've done a good job of explaining this, and demonstrated a good understanding of primitives vs complex types 👍

Week3/homework/step3-bonus.js
```diff
@@ -5,6 +5,7 @@ const values = ['a', 'b', 'c', 'd', 'a', 'e', 'f', 'c'];
 function makeUnique(arr) {
   // Replace this comment and the next line with your code
   console.log(arr);
+  return arr.filter((value, index) => arr.indexOf(value) === index);
```
nickmcmillan: Nice solution! If you're interested I'd encourage you to check out `Set` which is another way to solve this one https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set

Week3/homework/step3.js
```diff
@@ -3,6 +3,10 @@
 function createBase(base) {
   // Replace this comment and the next line with your code
   console.log(base);
+  return function addSix(number) {
```
nickmcmillan: Good. You don't need to name this function `addSix` though - you could leave it as an anonymous function.

PR #441, opened 2019-07-17
Javascript2 homework 3

elicarcar:


Week3/homework/step2-1.js
```diff
@@ -1,9 +1,7 @@
 'use strict';
 
 function foo(func) {
-  // What to do here?
-  // Replace this comment and the next line with your code
-  console.log(func);
+  func();
```
nickmcmillan: Perfect 🙌

Week3/homework/step2-2.js
```diff
@@ -2,19 +2,25 @@
...
-  // Replace this comment and the next line with your code
-  console.log(startIndex, stopIndex, threeCallback, fiveCallback, numbers);
+  numbers.forEach(function(number) {
+    if (number % 3 === 0) {
+      threeCallback(number);
+    }
+    if (number % 5 === 0) {
+      fiveCallback(number);
+    }
+  });
```
nickmcmillan: Don't forget to call both callback functions if a number is divisible by 15.

Week3/homework/step2-2.js
```diff
@@ -2,19 +2,25 @@
...
 function sayThree(number) {
-  // Replace this comment and the next line with your code
-  console.log(number);
+  console.log(`The number ${number} is divisible by 3!`);
 }
 
 function sayFive(number) {
-  // Replace this comment and the next line with your code
-  console.log(number);
+  console.log(`The number ${number} is divisible by 5!`);
```
nickmcmillan: Good use of template literals 👍

Week3/homework/step2-3.js
```diff
@@ -4,41 +4,42 @@
 function repeatStringNumTimesWithFor(str, num) {
   // eslint-disable-next-line prefer-const
   let result = '';
-
-  // Replace this comment and the next line with your code
-  console.log(str, num, result);
+  for (let i = 0; i < num; i++) {
+    result = result.concat(str);
```
nickmcmillan: `concat` is a good solution, but this could be simpler. Try just using regular string concatenation, ie; `result += str`

Week3/homework/step2-3.js
```diff
@@ -4,41 +4,42 @@
...
-  // Replace this comment and the next line with your code
-  console.log(str, num, result);
+  let i = 0;
+  if (num <= 0) {
+    return '';
+  }
+  do {
+    i++;
+    result += str;
+  } while (i < num);
```
nickmcmillan: Nice work on the `while` and `do while` loops

Week3/homework/step2-4.js
```diff
@@ -1,7 +1,9 @@
 'use strict';
 
 function Dog() {
-  // add your code here
+  this.name = 'Rupert';
+  this.color = 'brown';
+  this.numLegs = 4;
```
nickmcmillan: You've hard-coded Rupert, brown, and 4, into the Dog constructor. That means any hound you create will always have these values.

const hound = new Dog('Rupert', 'brown', 4);

Week3/homework/step2-5.js
```diff
@@ -3,9 +3,11 @@
 function multiplyAll(arr) {
   // eslint-disable-next-line
   let product = 1;
-
-  // Replace this comment and the next line with your code
-  console.log(arr, product);
+  for (let i = 0; i < arr.length; i++) {
+    for (let j = 0; j < arr[i].length; j++) {
+      product *= arr[i][j];
```
nickmcmillan: Nice 👍

Week3/homework/step2-6.js
```diff
@@ -2,15 +2,19 @@
 
 const arr2d = [[1, 2], [3, 4], [5, 6]];
 const arr3d = [[[1, 2], [3, 4]], [[5, 6], [7, 8]]];
-
 function flattenArray2d(arr) {
-  // Replace this comment and the next line with your code
-  console.log(arr);
+  const arr2 = arr.reduce(function(newArr, arrItem) {
```
nickmcmillan: Great use of `reduce`. You could even `return` directly on line 6 - instead of assigning it to a new variable which you return later (`arr2`)

Week3/homework/step2-6.js
```diff
@@ -2,15 +2,19 @@
...
+  }, []);
+  return arr2;
 }
 
 function flattenArray3d(arr) {
-  // Replace this comment and the next line with your code
-  console.log(arr);
+  return arr.reduce(
+    (acc, arrItem) =>
+      Array.isArray(arrItem) ? acc.concat(flattenArray3d(arrItem)) : acc.concat(arrItem),
```
nickmcmillan: Very smart to be checking if it's an Array first, and by using a ternary operator. Great work.

Week3/homework/step2-7.js
```diff
@@ -1,15 +1,18 @@
 'use strict';
 
+// In this first example, we see that this function passes directly the variable value inside the function.
+// Changing the argument inside the function doesn't affect the variable passed outside of the function.
+// So as long as we don't console log the variable inside the function (f1) our variables 'true' value will remain as 9.
```
nickmcmillan: While this is true, it's only part of the story. 
It's also important to understand the difference between changing the value of a variable (like changing x from 9 to 10), and changing the a property of a variable (like changing val.x from 9 to 10). Have a read through some of the discussions on this page https://stackoverflow.com/questions/6605640/javascript-by-reference-vs-by-value

Week3/homework/step3-bonus.js
```diff
@@ -3,8 +3,16 @@
...
+  arr.sort(function(letter1, letter2) {
+    return letter1 > letter2;
+  });
+  const unique = [];
+  for (let i = 0; i < arr.length; i++) {
+    if (arr[i] !== arr[i + 1]) {
+      unique.push(arr[i]);
+    }
+  }
+  return unique;
```
nickmcmillan: Not quite, if I run this code `uniqueValues` has the same exact letters and order as the original `values`. Perhaps re-think this. `filter` seems like it might be a suitable method for the job? Or perhaps `Set`?

Week3/homework/step3.js
```diff
@@ -1,12 +1,13 @@
 'use strict';
 
 function createBase(base) {
-  // Replace this comment and the next line with your code
-  console.log(base);
+  function addNumToBase(num) {
```
nickmcmillan: Nice work. You could even `return` an anonymous function here, instead of naming it and returning it later.

PR #443, opened 2019-07-18
mehmet bozdemir Js2w3 homweork week-3

mehmet-bozdemir:


Week3/homework/step2-1.js
```diff
@@ -2,8 +2,7 @@
 
 function foo(func) {
   // What to do here?
-  // Replace this comment and the next line with your code
-  console.log(func);
+  return func();
```
nickmcmillan: Nice. But no need to use `return` here, seeing as you don't do anything with the returned value when you call `foo`

Week3/homework/step2-2.js
```diff
@@ -2,19 +2,28 @@
 
 function threeFive(startIndex, stopIndex, threeCallback, fiveCallback) {
   const numbers = [];
-
-  // Replace this comment and the next line with your code
-  console.log(startIndex, stopIndex, threeCallback, fiveCallback, numbers);
+  for (let i = startIndex; i <= stopIndex; i++) {
+    numbers.push(i);
+  }
+  console.log(numbers);
```
nickmcmillan: Good work here, don't forget to remove your extra console logs though!

Week3/homework/step2-2.js
```diff
@@ -2,19 +2,28 @@
...
+      fiveCallback();
+    }
+  }
 }
 
-function sayThree(number) {
-  // Replace this comment and the next line with your code
-  console.log(number);
+function sayThree() {
+  console.log(3);
```
nickmcmillan: You could go a bit further here, what if this function accepted the number as a parameter? You could then use that in the console log so that it's a bit more descriptive.

Week3/homework/step2-3.js
```diff
@@ -2,39 +2,44 @@
...
-  let result = '';
-
-  // Replace this comment and the next line with your code
-  console.log(str, num, result);
-
-  return result;
+  // repeat after me
+  let repeatString = '';
+
+  for (; num > 0; ) {
```
nickmcmillan: This works, but I'm not sure where you learnt it from! I'd prefer to see properly written for loops, with the variable declaration and the incrementer

Week3/homework/step2-3.js
```diff
@@ -2,39 +2,44 @@
...
+  // repeat after me
+  let repeatString = '';
+  do {
+    repeatString += str;
+    num--;
+  } while (num > 0);
+  if (num < 0) {
+    return '';
+  }
+  return repeatString;
```
nickmcmillan: Good work here

Week3/homework/step2-4.js
```diff
@@ -1,10 +1,11 @@
 'use strict';
 
+// Define a Constructor Function
 function Dog() {
-  // add your code here
+  this.name = 'Rupert';
+  this.color = 'brown';
+  this.numLegs = 4;
```
nickmcmillan: You've hard-coded Rupert, brown, and 4, into the Dog constructor. That means any hound you create will always have these values.

const hound = new Dog('Rupert', 'brown', 4);

Week3/homework/step2-5.js
```diff
@@ -1,17 +1,20 @@
...
 function multiplyAll(arr) {
-  // eslint-disable-next-line
   let product = 1;
 
-  // Replace this comment and the next line with your code
-  console.log(arr, product);
+  for (let i = 0; i < arr.length; i++) {
+    for (let j = 0; j < arr[i].length; j++) {
+      console.log([j]);
+      product = product * arr[i][j];
```
nickmcmillan: 👍

Week3/homework/step2-6.js
```diff
@@ -1,22 +1,21 @@
...
+  }
 }
 
-function flattenArray3d(arr) {
-  // Replace this comment and the next line with your code
-  console.log(arr);
+for (let i = 0; i < flattenArray3d.length; i++) {
+  for (let z = 0; z < flattenArray3d[i].length; z++) {
+    for (let j = 0; j < flattenArray3d[i].length; j++) {
+      console.log(flattenArray3d[i][z][j]);
```
nickmcmillan: Great that you found this solution. You could also investigate using recursion which will help you avoid nesting 3 loops - generally speaking the fewer loops the better!

Week3/homework/step2-7.js
```diff
@@ -21,3 +21,9 @@ f2(y);
 console.log(y);
 
 // Add your explanation as a comment here
+/*
+---pass by value 
+the parameter of "val" has nothing to do with argument "x". Const x is beyond the function scope.
```
nickmcmillan: While this is true, it's only part of the story.
It's also important to understand the difference between changing the value of a variable (like changing x from 9 to 10), and changing the a property of a variable (like changing val.x from 9 to 10). Have a read through some of the discussions on this page https://stackoverflow.com/questions/6605640/javascript-by-reference-vs-by-value

Week3/homework/step2-7.js
```diff
@@ -21,3 +21,9 @@ f2(y);
...
+---pass by reference
+here the code "val.x" is directly addressing the object by targeting object value.
```
nickmcmillan: Yep, spot on!

Week3/homework/step3-bonus.js
```diff
@@ -2,9 +2,9 @@
 
 const values = ['a', 'b', 'c', 'd', 'a', 'e', 'f', 'c'];
 
-function makeUnique(arr) {
-  // Replace this comment and the next line with your code
-  console.log(arr);
+function makeUnique() {
+  const unique = values.filter((v, i, a) => a.indexOf(v) === i);
```
nickmcmillan: Brilliant. You could `return` here instead of creating a variable.

Week3/homework/step3.js
```diff
@@ -1,8 +1,9 @@
 'use strict';
 
-function createBase(base) {
-  // Replace this comment and the next line with your code
-  console.log(base);
+function createBase(a) {
+  return function(y) {
+    console.log(y + a);
```
nickmcmillan: The function should be responsible for returning a number, not console logging. You should return y + a here instead of console logging it.

Week3/homework/step2-1.js
mehmet-bozdemir: Dear Nick, thank you for your great feedback. I am gonna implement your feedback ASAP. Good luck.

PR #445, opened 2019-07-18
Homework week3

gulyener:


Week3/homework/step2-1.js
```diff
@@ -1,9 +1,7 @@
 'use strict';
 
 function foo(func) {
-  // What to do here?
-  // Replace this comment and the next line with your code
-  console.log(func);
+  return func();
```

Week3/homework/step2-2.js
```diff
@@ -2,19 +2,27 @@
...
+  numbers.forEach(number => {
+    const isNumberDivisibleBy5 = number % 5 === 0;
+    const isNumberDivisibleBy3 = number % 3 === 0;
+    if (isNumberDivisibleBy3) {
+      threeCallback(number);
+    }
+    if (isNumberDivisibleBy5) {
+      fiveCallback(number);
+    }
+  });
```

You need not write separate for loops here for creating the numbers array and then iterating it for callback logic. You can combine both these actions in a single for loop like
```
  numbers.push(i);
    // and so on
  }
}
```

2. Also, I see that you haven't implemented the third requirement which requires to call both callbacks if a number is divisible by both 3 and 5. Could you please add that as well?
Hint: mathematically speaking, a number which is divisible by 3 and 5 will be divisible by 15 as well. Thus, you can also write this logic as simply if (numbers[ii] % 15 === 0) instead of having the && condition. 🙂

Week3/homework/step2-3.js
```diff
@@ -4,40 +4,57 @@
 function repeatStringNumTimesWithFor(str, num) {
   // eslint-disable-next-line prefer-const
   let result = '';
-
-  // Replace this comment and the next line with your code
-  console.log(str, num, result);
-
+  for (let i = 0; i < num; i++) {
+    if (num < 0) {
```
yash-kapila: I think this `if` condition belongs outside the `for` loop rather than inside. Could you please check?

Week3/homework/step2-3.js
```diff
@@ -4,40 +4,57 @@
...
 function repeatStringNumTimesWithWhile(str, num) {
   // eslint-disable-next-line prefer-const
   let result = '';
-
-  // Replace this comment and the next line with your code
-  console.log(str, num, result);
-
+  let i = 0;
+  while (i < num) {
+    if (num < 0) {
```
yash-kapila: Same as `for` loop. I think the `if` condition needs to be outside the `loop`. 🙂

Week3/homework/step2-3.js
```diff
@@ -4,40 +4,57 @@
...
 function repeatStringNumTimesWithDoWhile(str, num) {
   // eslint-disable-next-line prefer-const
   let result = '';
-
-  // Replace this comment and the next line with your code
-  console.log(str, num, result);
-
+  let i = 0;
+  do {
+    if (num <= 0) {
```
yash-kapila: The same comment here from `for` and `while` loop 😄

Week3/homework/step2-4.js
```diff
@@ -1,7 +1,9 @@
 'use strict';
 
 function Dog() {
-  // add your code here
+  this.name = 'Karabas';
+  this.color = 'Black and white';
+  this.numLegs = 4;
```
yash-kapila: Ah. This is not completely right. You are correct in the way of creating a constructor function however, the values that you have provided to it i.e `Karabas `, `Black and white` and `4` are all hardcoded. This means, I can't use this constructor function to create a new dog object. The right way of doing it will be using function parameters:
```
  this.name = name;
  this.color = color;
  this.numLegs = legs;
}
```

```
// and so on
```

Week3/homework/step2-5.js
```diff
@@ -3,10 +3,11 @@
 function multiplyAll(arr) {
   // eslint-disable-next-line
   let product = 1;
-
-  // Replace this comment and the next line with your code
-  console.log(arr, product);
-
+  for (let i = 0; i < arr.length; i++) {
+    for (let j = 0; j < arr[i].length; j++) {
+      product = product * arr[i][j];
```

Week3/homework/step2-6.js
```diff
@@ -4,18 +4,39 @@ const arr2d = [[1, 2], [3, 4], [5, 6]];
...
-  console.log(arr);
+  const newArr = [];
+  arr.forEach(element => {
+    element.forEach(item => {
+      item.forEach(elm => {
+        newArr.push(elm);
+      });
+    });
+  });
+  return newArr;
```
yash-kapila: Very well. Would you like to try it with `reduce` as well? I believe your code would be much smaller and easier to understand if using `reduce` rather than `forEach`. 🙂 

Also, if you like to take a step further, try to search on google on how you can implement it using recursion.

Week3/homework/step2-7.js
```diff
@@ -21,3 +27,28 @@ f2(y);
...
+the 'val' parameter in f2 will get the argument of 'y' then val.x will actually be y.x
+which will choose the x property of y: {x:9} Then val.x = val.x + 1; adds 1 to the value of x. 
+Later when we log y, we can see that it has changed to {x: 10}. This is pass by reference. 
+
+This is an explanation that I like:
+
+Javascript is always pass by value, but when a variable refers to an object (including arrays), 
+the "value" is a reference to the object. 
+
+Changing the value of a variable never changes the underlying primitive or object,
```

Week3/homework/step3-bonus.js
```diff
@@ -3,8 +3,8 @@
 const values = ['a', 'b', 'c', 'd', 'a', 'e', 'f', 'c'];
 
 function makeUnique(arr) {
-  // Replace this comment and the next line with your code
-  console.log(arr);
+  arr = arr.filter((item, index, a) => a.indexOf(item) === index);
+  return arr;
```
yash-kapila: Very well. This logic works fine but it seems a bit too much. Try and read about a new concept in ES6 which is called Sets and see if you can use it to make this code shorter. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set

Week3/homework/step3.js
```diff
@@ -1,8 +1,9 @@
 'use strict';
 
 function createBase(base) {
-  // Replace this comment and the next line with your code
-  console.log(base);
+  return function(num) {
+    return base + num;
```

Week3/homework/step2-2.js
gulyener: Thank you i applied the changes

Week3/homework/step2-3.js
gulyener: Okay I did it thanks!

Week3/homework/step2-4.js
gulyener: I think I did this on freecodecamp with my solution and it worked, so I didn't put much thought into it. But you are right, the whole point of creating a constructor function is to be able to pass whichever argument we want. Thank you!

Week3/homework/step2-5.js
gulyener: Thanks!

Week3/homework/step2-6.js
gulyener: Okay I added two functions using reduce and recursion.

Week3/homework/step3-bonus.js
gulyener: That's very practical thank you!

gulyener:
I apologise for messing up with the commits. I thought I was being careful but it seems I wasn't.. Thanks for still reviewing my homework. I hope never to make mistakes like this in the future.

PR #447, opened 2019-07-20
finished js02-week03 homework

Laurelinduilwen:


Week3/homework/step2-1.js
```diff
@@ -1,9 +1,7 @@
 'use strict';
 
 function foo(func) {
-  // What to do here?
-  // Replace this comment and the next line with your code
-  console.log(func);
+  return func();
```

Week3/homework/step2-2.js
```diff
@@ -3,18 +3,28 @@
 function threeFive(startIndex, stopIndex, threeCallback, fiveCallback) {
   const numbers = [];
 
-  // Replace this comment and the next line with your code
-  console.log(startIndex, stopIndex, threeCallback, fiveCallback, numbers);
+  for (let number = startIndex; number <= stopIndex; number++) {
```

1. You need not write separate for loops here for creating the numbers array and then iterating it for callback logic. You can combine both these actions in a single for loop like
```
  numbers.push(i);
    // and so on
  }
}
```
2. Also, I see that you haven't implemented the third requirement which requires to call both callbacks if a number is divisible by both 3 and 5. Could you please add that as well?
Hint: mathematically speaking, a number which is divisible by 3 and 5 will be divisible by 15 as well. Thus, you can also write this logic as simply if (numbers[ii] % 15 === 0) instead of having the && condition. 🙂

Week3/homework/step2-3.js
```diff
@@ -30,7 +37,15 @@ console.log('while', repeatStringNumTimesWithWhile('abc', 3));
 function repeatStringNumTimesWithDoWhile(str, num) {
   // eslint-disable-next-line prefer-const
   let result = '';
-
+  let i = 0;
+  if (num <= 0) {
+    return '';
+  }
```
yash-kapila: Why not have this `if` condition check for `for` and `while` loops as well? 🙂

Week3/homework/step2-4.js
```diff
@@ -1,7 +1,9 @@
 'use strict';
 
 function Dog() {
-  // add your code here
+  this.name = 'Caramel';
+  this.color = 'Beige';
+  this.numLegs = 4; // add your code here
```
yash-kapila: Ah. This is not completely right. You are correct in the way of creating a constructor function however, the values that you have provided to it i.e `Caramel`, `Beige` and `4` are all hardcoded. This means, I can't use this constructor function to create a new dog object. The right way of doing it will be using function parameters:
```
  this.name = name;
  this.color = color;
  this.numLegs = legs;
}
```

```
// and so on
```

Week3/homework/step2-5.js
```diff
@@ -3,10 +3,11 @@
...
   let product = 1;
-
-  // Replace this comment and the next line with your code
-  console.log(arr, product);
-
+  for (let i = 0; i < arr.length; i++) {
+    for (let j = 0; j < arr[i].length; j++) {
+      product = product * arr[i][j];
+    }
+  }
```

Week3/homework/step2-6.js
```diff
@@ -4,13 +4,18 @@ const arr2d = [[1, 2], [3, 4], [5, 6]];
...
 }
 
 function flattenArray3d(arr) {
-  // Replace this comment and the next line with your code
-  console.log(arr);
+  const new3d = arr.reduce(
+    (acc, curr) => (Array.isArray(curr) ? acc.concat(flattenArray3d(curr)) : acc.concat(curr)),
+    [],
+  );
+  return new3d;
```
If you would like to take a step further, I would suggest that you have a look in Google to see how you could do the same thing using a concept called `recursion` 🙂

Week3/homework/step2-7.js
```diff
@@ -20,4 +20,18 @@ f2(y);
 
 console.log(y);
 
-// Add your explanation as a comment here
+f2(y);
+console.log(y);
+
+/* when assigning variables javascript passes objects as reference (pretty much everything is an object in JS) while assigning primitives as values
+when we declare x with a number in the first part JS assigns the value of 9 in the memory and places it there and it is not changeable
```
yash-kapila: 👏

Week3/homework/step3-bonus.js
```diff
@@ -3,8 +3,16 @@
...
-  console.log(arr);
+  arr.sort(function(letter1, letter2) {
+    return letter1 > letter2;
+  });
+  const unique = [];
+  for (let i = 0; i < arr.length; i++) {
+    if (arr[i] !== arr[i + 1]) {
+      unique.push(arr[i]);
+    }
+  }
```
yash-kapila: Aahaan. This logic works fine but it seems a bit too much. Try and read about a new concept in ES6 which is called Sets and see if you can use it to make this code shorter. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set

Week3/homework/step3.js
```diff
@@ -1,8 +1,7 @@
 'use strict';
 
 function createBase(base) {
-  // Replace this comment and the next line with your code
-  console.log(base);
+  return number => number + base;
```

Laurelinduilwen:
thank you for the feedback Yash :D 

I am terribly sorry that it took this long to implement them but JS3 kinda hit me like a truck when I was already falling behind 😅

Week3/homework/step2-2.js
Laurelinduilwen: aah I completely forgot about the third one while doing this 🗡 I hope it is fixed now ^^

Week3/homework/step2-3.js
Laurelinduilwen: done ^_^

Week3/homework/step2-4.js
Laurelinduilwen: I think I also messed up this one and forgot the real purpose of it  even though I have done similar constructor functions in the past -_- Fixed :D

Week3/homework/step2-6.js
Laurelinduilwen: I tried it but got lost in weird places while exploring. This will be on my list, recursion is scary :*(

Week3/homework/step3-bonus.js
Laurelinduilwen: wow, this was really easy with using sets :D thank you!

Week3/homework/step3-bonus.js
```diff
@@ -3,8 +3,10 @@
 const values = ['a', 'b', 'c', 'd', 'a', 'e', 'f', 'c'];
 
 function makeUnique(arr) {
-  // Replace this comment and the next line with your code
-  console.log(arr);
+  const set1 = new Set(arr);
+  const setArray = [];
+  set1.forEach(i => setArray.push(i));
+  return setArray;
```
yash-kapila: How about this:
```
const set = new Set(arr);
return Array.from(set);
```
😊

PR #449, opened 2019-08-18
Homework Week-1 Step-1 Done

msnmz:


wilgert:
This looks really good. I have no real comments on the JavaScript part.
The only thing you can really improve is try to make the css in such a way that your title does not end up sticking out of the box on narrow screens. See screenshot below:
Screenshot 2019-08-19 at 21 25 53

msnmz:
Thank you very much for your feedback :)
I added a width attribute to the CSS rule and now it is better.

PR #451, opened 2019-08-20
JS2-Week1-Homework

malnaqeeb:


Week1/homework/app.js
```diff
@@ -1,11 +1,142 @@
...
+
+    // add language
+    language.innerText = 'language: ' + info.language;
+    li.appendChild(language);
+    language.className = 'language';
+  }
 }
+// the next step is to design a function to create a div => add it to the body => and call the bookLibrary(createTheMainList) 
+function page() {
+  // create div contains  ul & h1
```
wilgert: Maybe my explanation was not clear. But the idea is that a comment should tell **why** the code does something a certain way. Your comments tell the reader **what** the code does. But the code is already clear enough to tell that by itself.

Week1/homework/app.js
```diff
@@ -1,11 +1,142 @@
...
+  Tom_Jones:'https://i.postimg.cc/43LtD7LV/1281362.jpg',
+  Emma:'https://i.postimg.cc/vTMQXXrz/Emma.jpg',
+  Frankenstein:'https://i.postimg.cc/brFBdwwt/Frankenstein.jpg',
+  Nightmare_Abbey:'https://i.postimg.cc/mDdsNRrb/Nightmare-Abbey.jpg',
+  Sybil:'https://i.postimg.cc/TP0MfGgr/Sybil.jpg',
+  Jane_Eyre:'https://i.postimg.cc/prY1YJCK/Jane-Eyre.jpg',
+};
+
+// Create a function to build a list of books including the required  information.
+function bookLibrary(createTheMainList) {
```
wilgert: The name of the parameter `createTheMainList` suggests that it is a function. But it is actually an html element. Please rename the parameter to something more appropriate.

Week1/homework/app.js
```diff
@@ -1,11 +1,142 @@
...
+  Jane_Eyre:'https://i.postimg.cc/prY1YJCK/Jane-Eyre.jpg',
+};
+
+// Create a function to build a list of books including the required  information.
+function bookLibrary(createTheMainList) { 
+  for (let i = 0; i < bookTitles.length; i++) {
+    // declare the variables to loop through the objects.(bookTitles, myBooks, bookCover)
+    const theTitle = bookTitles[i]; // loop through the titles   (bookTitles object)
+    const info = myBooks[theTitle]; // loop through the book information (myBooks object)
+    const photoOfCover = bookCover[theTitle]; // loop through the covers (bookCover object)
```
wilgert: Please read the instructions 1.7 and 1.8 carefully again.
Doing everything in one loop and putting the img in the html in one function is not a bad solution. But it is not what was asked for.

Week1/homework/style.css
```diff
@@ -1 +1,81 @@
...
+  
+  h2 {
+    text-align: center;
+    font-family: Impact, Charcoal, sans-serif;
+    font-size: 1.7em;
+    font-weight: normal;
+  }
+  
+  p {
+    font-family: "Comic Sans MS", cursive, sans-serif;
```
wilgert: As a side note, and maybe you already know this. But Comic Sans MS is normally not chosen as a font-family on websites. Furthermore I would not use 3 different font-families on a webpage.

Week1/homework/style.css
```diff
@@ -1 +1,81 @@
...
+  
+.item {
+    float: left;
+    width: 23%;
+    height: 20em;
+    display: block;
+    background-color: rgb(255, 255, 255);
+    border-radius: 8px;
+    }
+  li.item{
```
wilgert: All `.items` are `li.items`. So you can remove the duplicate code here.
Interesting question: what would the values for display, margin and padding be in the browser with the css that you wrote?

Week1/homework/style.css
```diff
@@ -1 +1,81 @@
...
+      padding: 0;
+    
+    }
+  
+  img {
+    max-height: 50%;
+    margin: auto;
+    display: block;
+    }
+  @media (min-width: 768px) and (max-width: 1024px) {
```
wilgert: Good that you thought about responsive design!

Week1/homework/style.css
```diff
@@ -1 +1,81 @@
...
+    display: block;
+    }
+  @media (min-width: 768px) and (max-width: 1024px) {      
+    .item{
+      width: 45%;
+    }
+    h1 {
+        font-size: 3em;
+        font-weight: bold;
+        color: antiquewhite;
```
wilgert: Why do you specify a slightly different `color` when the screen is at this size?

Week1/homework/app.js
```diff
@@ -1,11 +1,142 @@
...
+    img.setAttribute('src', photoOfCover);
+    img.setAttribute('alt', 'a cover for ' + info.name);
+
+    // add the Author
+    author.innerText = 'Author: ' + info.author;
+    li.appendChild(author);
+    author.className = 'author';
+
+    // add language
+    language.innerText = 'language: ' + info.language;
```
wilgert: Try to be consistent with capitals. _language_ should be _Language_.

PR #454, opened 2019-08-21
Week 1 Homework

ahmthsrvsmz:


Week1/homework/app.js
```diff
@@ -1,11 +1,141 @@
...
+let button = document.getElementById('clickButton');
+
+button.addEventListener('click', function() {
+  const divElement = document.createElement('div');
+
+  document.getElementsByClassName('flexbox1')[0].appendChild(divElement);
+  let list = document.createElement('ul');
+  divElement.appendChild(list);
+  let line = document.createElement('li');
+  line.innerText = 'harry_potter_chamber_secrets';
```
wilgert: Please read the instructions again carefully. The idea is that you put all the data about books in the `bookDetails` object and that you loop over the `bookTitles` array to get the data from this object.
You have now put the data (like the title) in function instead of taking it from the object.

Week1/homework/app.js
```diff
@@ -1,11 +1,141 @@
...
+};
+
+//1.5
+const buttonTry = document.getElementById('tryMeButton');
+
+buttonTry.addEventListener('click', function getBookDetails() {
+  const bookInfo = Object.entries(bookDetails);
+  console.log(bookInfo);
+
+  for (let i = 0; i < bookInfo.length; i++) {
```
wilgert: here you should loop over the `bookTitles` and use each value to access the data about each book in the `bookDetails` object.

ahmthsrvsmz: I changed this code and use bookInfo.
const bookInfo = Object.entries(bookDetails);

//1.8

Thank you for your feedback.

Week1/homework/app.js
ahmthsrvsmz: I changed this code. Created new for loop inside of the function. But there is not any bookDetails object yet. I picked bookTitles's child. 

//1.3

Thank you for your feedback

Tjebbee:
@ahmthsrvsmz , are you done implementing the feedback from @wilgert ? If so, ask him to check it :)

PR #458, opened 2019-08-24
js2 week1 pushing files

ersarue:


Week1/homework/app.js
```diff
@@ -1,11 +1,178 @@
...
+
+// *** 1.3
+function list() {
+  for (let i = 0; i < bookTitles.length; i++) {
+    const listItem = document.createElement('li');
+    listItem.innerHTML = bookTitles[i];
+    createList.appendChild(listItem);
+  }
+}
+const createDiv = document.createElement('div');
```
J05J4: `createDic` is a name I would expect for a function, because of the verb in the name. `document.creteElement('div')` creates a `<div>` but it would make sense to to give this a name by which you could recognise it.

Week1/homework/app.js
```diff
@@ -1,11 +1,178 @@
...
+createDiv.id = 'bookList';
+document.getElementsByTagName('body')[0].appendChild(createDiv);
```
```suggestion
document.body.appendChild(createDiv);
```

Week1/homework/app.js
```diff
@@ -1,11 +1,178 @@
...
+
+const createList = document.createElement('ul');
+document.getElementById('bookList').appendChild(createList);
```
J05J4: This works, but you could also do `createDiv.appendChild(createList);`

Week1/homework/app.js
```diff
@@ -1,11 +1,178 @@
...
+  1984: 'https://i.gr-assets.com/images/S/compressed.photo.goodreads.com/books/1532714506l/40961427._SX318_.jpg',
+  the_little_prince:
+    'https://i.gr-assets.com/images/S/compressed.photo.goodreads.com/books/1367545443l/157993.jpg',
+  frankenstein:
+    'https://i.gr-assets.com/images/S/compressed.photo.goodreads.com/books/1498841231l/35031085.jpg',
+  crime_and_punishment:
+    'https://i.gr-assets.com/images/S/compressed.photo.goodreads.com/books/1382846449l/7144.jpg',
+};
+
+const source = Object.values(bookCovers);
```
J05J4: In your JavaScript object you are using the name of the book as key. You have the same items in your array: you should go over the elements in your array and use every item in the array to lookup a value in both objects.

books/app.js
```diff
@@ -0,0 +1,178 @@
+'use strict';
```
J05J4: THe files in `books/` are similar; you should remove one of them.

ersarue: removed

Week1/homework/app.js
ersarue: code changed

Week1/homework/app.js
ersarue: code changed

Week1/homework/app.js
ersarue: I got what you meant and changed the code.

Week1/homework/app.js
ersarue: I'll go over Object.key issue and come up with right way

PR #459, opened 2019-08-24
JS2 Week1 Homework

FarukTercan:


Week1/homework/app.js
```diff
@@ -1,11 +1,129 @@
...
+
+    for (let i = 0; i < myBooks.length; i++) {
+      const liTag = document.createElement('li');
+      liTag.innerHTML = myBooks[i];
+      ulTag.appendChild(liTag);
+    }
+  }
+
+  // makeList(); 
+  //I canceled function calling to avoid showing on page
```
J05J4: You should update your function to work with the details of your book (now you have the same functionality twice in this file), or maybe eve n better: call the `makeList()` function from the body of your `infoBookList()` function.

Week1/homework/app.js
```diff
@@ -1,11 +1,129 @@
...
+    essays_montaigne: { title: 'Essays', release: '(1580)', author: 'Montaigne' },
+    blindness_saramago: { title: 'Blindness', release: '(1995)', author: 'Saramago' },
+    a_confession: { title: 'A Confession', release: '(1884)', author: 'Tolstoy' },
+    sherlock_holmes: { title: 'Sherlock Holmes', release: '(1892)', author: 'AC Doyle' },
+  };
+
+
+  function makeBookList() {
+    const divBookList = document.createElement('div');
+    document.getElementsByTagName('body')[0].appendChild(divBookList);
```
```suggestion
    document.body.appendChild(divBookList);
```

Week1/homework/app.js
```diff
@@ -1,11 +1,129 @@
...
+
+    const ulBookList = document.createElement('ul');
```
J05J4: Try to find names that are clear: it's good that you have "bookList" for this element, but it is not needed to put this `ul` in front. The problem in general with putting the type in the name, is that when you later want to change the type (use an ordered list for example) you would have to change the name too.

Week1/homework/app.js
```diff
@@ -1,11 +1,129 @@
...
+    divBookList.appendChild(ulBookList);
+
+    const values = Object.values(infoBookList);
```
J05J4: Why do you use `Object.values()`? 
You should have a look ad the message from Wilgert in Slack (https://hackyourfuture.slack.com/archives/CL795TNFQ/p1566576723000500) try to improve you functionality based on this message and if something is not clear, ask a question.

Week1/homework/app.js
```diff
@@ -1,11 +1,129 @@
...
+      animal_farm: 'https://s.s-bol.com/imgbase0/imagebase3/large/FC/6/2/8/8/1001004005988826.jpg',
+      the_little_prince: 'https://s.s-bol.com/imgbase0/imagebase3/large/FC/7/5/0/8/1001004000638057.jpg',
+      the_alchemist: 'https://s.s-bol.com/imgbase0/imagebase3/large/FC/9/3/2/3/9200000041223239.jpg',
+      essays_montaigne: 'https://s.s-bol.com/imgbase0/imagebase3/large/FC/1/0/6/6/1001004000636601.jpg',
+      blindness_saramago: 'https://s.s-bol.com/imgbase0/imagebase3/large/FC/0/6/8/0/9200000001310860.jpg',
+      a_confession: 'https://images-na.ssl-images-amazon.com/images/I/71i8kz3FrZL.jpg',
+      sherlock_holmes: 'https://images-na.ssl-images-amazon.com/images/I/91743TsDedL.jpg',
+    };
+
+    const coverValues = Object.values(coverBookList);
```
J05J4: Same here: with `Object.values()` you are changing the values of the object into an array: if this is the only way you are using this object, you cold have just create it as an object, which wasn't explicitly the task here. 
You are now counting on the same order of your object as your array while the homework also states "Remember that objects are not ordered, so you cannot guarantee that the first key is the first li element."
You should use the items in your array as keys to find the correct information in your objects. Improve your code to do so, and if anything is unclear please ask a question in Slack!

Week1/homework/app.js
```diff
@@ -1,11 +1,129 @@
...
+    'essays_montaigne',
+    'blindness',
+    'a_confession',
+    'sherlock_holmes',
   ];
 
-  // Replace with your own code
-  console.log(bookTitles);
+  // 1.2
+  console.log(myBooks);
```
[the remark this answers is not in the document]
FarukTercan: 👍  I removed it

Week1/homework/app.js
FarukTercan: I changed my function system and canceled this part

Week1/homework/app.js
FarukTercan: Thanks for the tip, I replaced it with the unproblematic name.

Week1/homework/app.js
FarukTercan: You are totally right, I did not pay attention to ordering issue. I improved my function and connected the elements in the myBooks array (as a key)  to the infoBookList object and coverBookList object. Thus, ordering problem is solved.

Week1/homework/app.js
FarukTercan: thanks it is more practical

PR #460, opened 2019-08-24
JavaScript2 Week1 HomeWork

mehmetekinci:


Week1/homework/app.js
```diff
@@ -1,11 +1,153 @@
...
+    divElementForInfo.className = 'secondDiv';
+    const title = document.createElement('h2');
+    divElementForInfo.appendChild(title);
+    const writer = document.createElement('h3');
+    divElementForInfo.appendChild(writer);
+    const lang = document.createElement('p');
+    divElementForInfo.appendChild(lang);
+  }
+  // creates a method which takes information from bookDetails object.
+  const objectValue = Object.values(bookDetails);
```
J05J4: Why are you using `Object.values()`?
Because of this you are using your object as an array. Other than that you are expecting the order of your object to be the same as the order of your array, this order is not guaranteed, so you can not count on it. 
Have a look at the message in Slack from Wilgert (https://hackyourfuture.slack.com/archives/CL795TNFQ/p1566576723000500) you should use the items in your array as keys to look for in your objects. Try to change your code to this, and if something is unclear, ask a question in Slack.

Week1/homework/app.js
```diff
@@ -1,11 +1,153 @@
...
+
+// creates a function which shows book details.
+function newBookInfo() {
+  // creates elements
+  for (let i = 0; i < bookTitles.length; i++) {
+    const listItem = document.createElement('li');
+    listElement.appendChild(listItem);
+    const divElement = document.createElement('div');
+    listItem.appendChild(divElement);
+    divElement.className = 'mainDiv';
```
J05J4: You don't really need those `div` elements: you could set the class to the `li` element or child element.

mehmetekinci:
I am sorry for late respond because was busy all weekend. Now I am working on it.

> Your code looks well structured. You are using `Object.values()` which you should not. I left a comment with a link to the message from Wilgert on this, try to improve your code and if something is unclear ask a question in Slack!

Week1/homework/app.js
mehmetekinci: for the first div you are right but I tried to use grid for. In this case how can I use flex without div?

mehmetekinci: hi @J05J4 .Finally I completed my changes. I hope it works.

PR #461, opened 2019-08-24
added hw layout

abdullahkanatsiz7:
my js2 week1 homework


Week1/homework/app.js
```diff
@@ -1,11 +1,143 @@
...
+  //the_great_gatsby
+  //war_and_peace
+  //the_odyssey
+  //crime_and_punishment
+  //the_stranger
+  //the_sun_also_rises
+  //candide
+
+  // 1.1
+  // const bookList = [
```
J05J4: You should keep using this array in the other functions.

Week1/homework/app.js
```diff
@@ -1,11 +1,143 @@
...
+    the_great_gatsby: 'images/the_great_gatsby.jpg',
+    war_and_peace: 'images/war_and_peace.jpg',
+    the_odyssey: 'images/the_odyssey.jpg',
+    crime_and_punishment: 'images/crime_and_punishment.jpg',
+    the_stranger: 'images/the_stranger.jpg',
+    the_sun_also_rises: 'images/the_sun_also_rises.jpg',
+    candide: 'images/candide.jpg',
+  };
+
+  const section = document.createElement('section');
```
J05J4: Nice semantic element!

Week1/homework/app.js
```diff
@@ -1,11 +1,143 @@
...
+  document.body.prepend(section);
```
J05J4: You should use `appendChild()`
```suggestion
  document.body.appendChild(section);
```

Week1/homework/app.js
```diff
@@ -1,11 +1,143 @@
...
+  const ul = document.createElement('ul');
+  section.appendChild(ul);
 
-  // Replace with your own code
-  console.log(bookTitles);
+  Object.keys(bookList).forEach(bookID => {
```
J05J4: This is fine too, you could have used the elements in your array `bookList` here.

Week1/homework/app.js
```diff
@@ -1,11 +1,143 @@
...
+    const li = document.createElement('li');
+    li.setAttribute('id', bookID);
+    ul.appendChild(li);
+  // I'm going to do create title element
+    const title = document.createElement('h1');
+    title.textContent = bookList[bookID].title;
+    li.appendChild(title);
+  // I'm going to do create author element
+    const author = document.createElement('p');
+    author.setAttribute('class', 'aut');
```
```suggestion
    author.classList.add('aut');
```

Week1/homework/app.js
```diff
@@ -1,11 +1,143 @@
...
+  const section = document.createElement('section');
+  document.body.prepend(section);
+  const ul = document.createElement('ul');
+  section.appendChild(ul);
 
-  // Replace with your own code
-  console.log(bookTitles);
+  Object.keys(bookList).forEach(bookID => {
+    const li = document.createElement('li');
+    li.setAttribute('id', bookID);
```
```suggestion
    li.id =bookID;
```

Week1/homework/app.js
abdullahkanatsiz7: thank you for this comment

Week1/homework/app.js
abdullahkanatsiz7: Yes you right I could be used appendChild but If you have a much JavaScript, it can visibly slow the loading of your page because it loads all of the JavaScript before it loads any of the HTML.
When you place your JavaScript links at the bottom of your HTML body, it gives the HTML time to load before any of the JavaScript loads, which can prevent errors, and speed up website response time.

Week1/homework/app.js
abdullahkanatsiz7: I have used this array in this function, do you want to mean something else?
//1.3
  // function general() {
  //   const newList = document.createElement('ul');

  //   for (let i = 0; i < bookList.length; i++) {
  //     const book = bookList[i];
  //     newList.innerHTML += `<li>${book}</li>`;
  //   }

  //   return newList;
  // }

  // document.body.append(general());

Week1/homework/app.js
abdullahkanatsiz7: thanks, I will be use it

Week1/homework/app.js
abdullahkanatsiz7: I didn't know that thanks again this is shorter I will use.

PR #462, opened 2019-08-24
JS2_Week_1_Homework

haha1906:


Week1/homework/app.js
```diff
@@ -2,10 +2,98 @@
...
+    },
+  };
+  function writeBookNamesInObject() {
+    for (const title of bookTitles) {
+      const container = document.createElement('div');
+      document.body.appendChild(container);
+      const coverImage = document.createElement('img');
+      coverImage.src = 'img/' + title + '.jpg';
+      container.appendChild(coverImage);
+      const ulElement = document.createElement('ul');
```
J05J4: You are creating a container and a unordered list for every item? You could use a nested list. 
Maybe interesting to take a look at the descriptive list: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dl

Week1/homework/app.js
```diff
@@ -2,10 +2,98 @@
...
+      container.appendChild(ulElement);
+      for (const property of Object.keys(allBooks[title])) {
+        console.log(property);
```
J05J4: You should remove this `console.log`s

PR #464, opened 2019-09-01
Js2 week2

malnaqeeb:


Week2/homework/maartjes-work.js
```diff
@@ -1,65 +1,80 @@
...
+    name: 'Look at application assignments new students',
+    duration: 40,
+  },
+];
+
+const maartjesTasks = monday.concat(tuesday);
+const maartjesHourlyRate = 20;
+
+function computeEarnings(tasks, hourlyRate) {
+  let result = tasks.map(toHours => toHours.duration / 60);
```
wilgert: The names you chose for this parameter is a bit strange. I would expect `toHours` to be the name of a function or method instead of a parameter. I would call this parameter just `task` because it is one item from the `tasks` array.

Week2/homework/maartjes-work.js
```diff
@@ -1,65 +1,80 @@
...
+  console.log(result); // [3, 2, 0.33, 3.35, 4, 3, 0.16, 3.35, 0.66]
+  result = result.filter(twoHoursOrMore => twoHoursOrMore >= 2);
```
wilgert: Same goes for this parameter. An array of durations that are `twoHoursOrMore` is the result you get from this filter.
I would call the item that get's passed in this filter `durationInHours`.

Week2/homework/maartjes-work.js
```diff
@@ -1,65 +1,80 @@
...
+  console.log(result); // [3, 2, 3.35, 4, 3, 3.35]
+  result = result.reduce((sum, twoHoursOrMore) => sum + twoHoursOrMore * hourlyRate, 0);
```
wilgert: I would also rename `twoHoursOrMore` to `durationInHours`.

Week2/homework/maartjes-work.js
malnaqeeb: done

Week2/homework/maartjes-work.js
malnaqeeb: done

Week2/homework/maartjes-work.js
malnaqeeb: done

malnaqeeb:
Thanks Welgert for your feedback ;) 
I've just implemented your comments.

Tjebbee:
@wilgert Thanks for reviewing, @malnaqeeb Well done!!

PR #466, opened 2019-09-02
Week2

rabrad:
For review

### DIFF
--- a/Week2/homework/maartjes-work.js
+++ b/Week2/homework/maartjes-work.js
@@ -47,7 +47,11 @@ const maartjesHourlyRate = 20;
 
 function computeEarnings(tasks, hourlyRate) {
   // Replace this comment and the next line with your code
-  console.log(tasks, hourlyRate);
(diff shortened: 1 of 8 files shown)


Week2/homework/maartjes-work.js
```diff
@@ -47,15 +47,20 @@ const maartjesHourlyRate = 20;
 
 function computeEarnings(tasks, hourlyRate) {
   // Replace this comment and the next line with your code
-  console.log(tasks, hourlyRate);
+  // console.log(tasks, hourlyRate);
+  return tasks
+    .map(task => task.duration / 60)
+    .filter(taskHour => taskHour >= 2)
+    .reduce((total, perHour) => total + perHour * hourlyRate, 0);
```
wilgert: The `perHour` variable is actually still a duration in hours. So I would name it something like `durationInHours`.

rabrad: Thank you for the suggestion. I've renamed the variable accordingly.

PR #470, opened 2019-09-03
JS2Week2 Homework

ihk2018:


Week2/homework/maartjes-work.js
```diff
@@ -46,16 +46,17 @@ const maartjesTasks = monday.concat(tuesday);
 const maartjesHourlyRate = 20;
 
 function computeEarnings(tasks, hourlyRate) {
-  // Replace this comment and the next line with your code
-  console.log(tasks, hourlyRate);
+  return tasks
+    .map(taskToDurations => taskToDurations.duration / 60)
```
wilgert: The names you chose for this parameter is a bit strange. I would expect `taskToDurations` to be the name of a function or method instead of a parameter. I would call this parameter just `task` because it is one item from the `tasks` array.

Week2/homework/maartjes-work.js
```diff
@@ -46,16 +46,17 @@ const maartjesTasks = monday.concat(tuesday);
...
+    .filter(lessThanTwoHours => lessThanTwoHours >= 2)
```
wilgert: Same goes for this parameter. An array of durations that are `lessThanTowHours` is the result you get from this filter.
I would call the item that get's passed in this filter `durationInHours`.

Week2/homework/maartjes-work.js
```diff
@@ -46,16 +46,17 @@ const maartjesTasks = monday.concat(tuesday);
...
+    .map(durationPerHour => durationPerHour * hourlyRate)
```
wilgert: I would also call this one `durationInHours`.

Week2/homework/maartjes-work.js
ihk2018: Dear Wilgert , thank you for the feedback. I must be more careful when I select the names. I made the changes according to your feedback.

PR #471, opened 2019-09-03
JS2 Week2 Homework

FarukTercan:


Week2/homework/maartjes-work.js
```diff
@@ -46,16 +46,18 @@ const maartjesTasks = monday.concat(tuesday);
 const maartjesHourlyRate = 20;
 
 function computeEarnings(tasks, hourlyRate) {
-  // Replace this comment and the next line with your code
-  console.log(tasks, hourlyRate);
+  return tasks
+    .map(task => task.duration)
+    .filter(minTwoHours => minTwoHours >= 120)
```
wilgert: The variable name `minTwoHours` suggests that it was already filtered. But you just get the duration in minutes in this filter function. I would name this variable `durationInMinutes`.

FarukTercan: Thanks for the suggestion, I've changed it. 
I gave that name because filter function creates an array with min two hours duration values.  But as you said the results are in minutes.

wilgert: Yes the result are in minutes. But even more important that variable is the value that gets filtered. You should give it a name that describes what that variable is, not what the filter returns. You could store the filter in a variable and give it a name that tells you what it does. Doing that can make your code more readable, because you can see immediately what the filter does. Like this: 

```
const fruits = [{name: 'apple', color: 'red'}, {name: 'banana', color: 'yellow'}, {name: 'strawberry', color: 'red'}]
const filterRedColor = fruit => fruit.color === 'red';

const onlyRedFruits = fruits.filter(filterRedColor)
```

PR #476, opened 2019-09-03
Js2 week2 home work

husenjanobul:


Week2/homework/maartjes-work.js
```diff
@@ -47,15 +47,24 @@ const maartjesHourlyRate = 20;
...
+  const totalHour = tasks.map(task => task.duration / 60);
+  const workHour = totalHour.filter(task => task >= 2);
+  const durationSum = workHour.reduce((durations, task) => durations + task, 0);
+  return durationSum * hourlyRate;
+  // return (
+  //   tasks
+  //     .map(task => task.duration / 60)
+  //     .filter(task => task >= 2)
+  //     .reduce((durations, task) => durations + task, 0) * hourlyRate
+  // );
```
yash-kapila: 👍 

I see that you tried to show both approaches of completing this homework. Quite nice. Both ways are just fine. Some developer's prefer approach 1 which is more explicit and succinct while some developers prefer approach 2 which is more short and crisp. When working in a team, everyone decides collectively which approach to take.

Week2/homework/map-filter.js
```diff
@@ -2,7 +2,9 @@
 
 function doubleOddNumbers(numbers) {
   // Replace this comment and the next line with your code
-  console.log(numbers);
+  const oddNumbers = numbers.filter(number => number % 2 !== 0);
+  return oddNumbers.map(number => number * 2);
```

PR #478, opened 2019-09-04
Js2 week2

veyselindi:
.


Week2/homework/maartjes-work.js
```diff
@@ -46,16 +46,17 @@ const maartjesTasks = monday.concat(tuesday);
 const maartjesHourlyRate = 20;
 
 function computeEarnings(tasks, hourlyRate) {
-  // Replace this comment and the next line with your code
-  console.log(tasks, hourlyRate);
+  const hourTask = tasks.map(task => task.duration / 60);
+  const moreThan2Hour = hourTask.filter(hour => hour >= 2);
+  return moreThan2Hour.reduce((acc, elem) => acc + elem, 0) * hourlyRate;
```
yash-kapila: 👍 but a small thing here - you could use better names for the callback arguments instead of `acc` and `elem`. Like you have meaningful names in `filter` and `map` like `hour` and `task` respectively.

Week2/homework/map-filter.js
```diff
@@ -1,8 +1,7 @@
 'use strict';
 
 function doubleOddNumbers(numbers) {
-  // Replace this comment and the next line with your code
-  console.log(numbers);
+  return numbers.filter(number => number % 2 !== 0).map(number => number * 2);
```

PR #479, opened 2019-09-04
NesibeInce JS2_HW2

NesibeInce:


Week2/homework/maartjes-work.js
```diff
@@ -47,15 +47,19 @@ const maartjesHourlyRate = 20;
 
 function computeEarnings(tasks, hourlyRate) {
   // Replace this comment and the next line with your code
-  console.log(tasks, hourlyRate);
+  const hourlyDuration = tasks.map(elem => elem.duration / 60);
+  const moreTwoHours = hourlyDuration.filter(elem => elem >= 2);
+  const sumHours = moreTwoHours.reduce((pv, cv) => pv + cv);
+  const total = sumHours * hourlyRate;
+  return total;
```
yash-kapila: 👍 for the logic. However, the name of arguments inside the `map`, `filter` and `reduce` functions aren't the best. Maybe you could use `task` or `hours` or `earnings` instead. What do you think?

Week2/homework/map-filter.js
```diff
@@ -2,7 +2,9 @@
 
 function doubleOddNumbers(numbers) {
   // Replace this comment and the next line with your code
-  console.log(numbers);
+  const oddNum = numbers.filter(num => num % 2 === 1);
+  const doubleOddNum = oddNum.map(num => num * 2);
+  return doubleOddNum;
```
yash-kapila: 👍 Interesting that you used `num % 2 === 1` instead of `num % 2 !== 0`. Most of the developers are used to second approach but I think your approach also works just fine. 🙂

Week2/homework/maartjes-work.js
NesibeInce: Yes, I also think that words you suggest look more suitable.
 Thank you in advance for your approval and concern.

Week2/homework/map-filter.js
NesibeInce: I think for a beginner, what I have written sounds more logical for me or easier to understand. However, as you mentioned, it think I should also use the most popular approach as a future pro. :)

PR #481, opened 2019-09-11
JS2-week3-homework

husenjanobul:


Week3/homework/step2-1.js
```diff
@@ -3,7 +3,7 @@
 function foo(func) {
   // What to do here?
   // Replace this comment and the next line with your code
-  console.log(func);
+  return func();
```

Week3/homework/step2-2.js
```diff
@@ -4,17 +4,30 @@ function threeFive(startIndex, stopIndex, threeCallback, fiveCallback) {
...
+    numbers.push(i);
+    if (i % 3 === 0 || i % 5 === 0) {
+      if (i % 3 === 0 && i % 5 === 0) {
+        threeCallback(i);
+        fiveCallback(i);
+      } else if (i % 3 === 0) {
+        threeCallback(i);
+      } else {
+        fiveCallback(i);
+      }
```
yash-kapila: Ah. I think you can improve the logic here a bit more. For example, if you think mathematically, a number which is divisible by both 3 and 5 will always be divisible by number 15, isn't it? Also, you don't really need a nested-if block here. You could rewrite the logic as:

```
if (i % 15 === 0) {
  // call both three and five callbacks
} else if(i % 3 === 0) {
  // call three callbacks
} else if(i % 5 === 0) {
  // call five callbacks
}
```

Week3/homework/step2-3.js
```diff
@@ -32,6 +39,12 @@ function repeatStringNumTimesWithDoWhile(str, num) {
   let result = '';
 
   // Replace this comment and the next line with your code
+  if (num > 0) {
```
yash-kapila: Any reason why you added this `if` check here in the `do-while` loop and not `for` or `while` loop?

Week3/homework/step2-4.js
```diff
@@ -2,6 +2,9 @@
 
 function Dog() {
   // add your code here
+  this.name = 'Xxx';
+  this.color = 'White';
+  this.numLegs = 3;
```

```
  this.name = name;
  this.color = color;
  this.numLegs = legs;
}

// and so on
```

Week3/homework/step2-5.js
```diff
@@ -5,6 +5,11 @@ function multiplyAll(arr) {
   let product = 1;
 
   // Replace this comment and the next line with your code
+  for (let i = 0; i < arr.length; i++) {
+    for (let a = 0; a < arr[i].length; a++) {
+      product *= arr[i][a];
+    }
+  }
```

Week3/homework/step2-6.js
```diff
@@ -5,12 +5,16 @@ const arr3d = [[[1, 2], [3, 4]], [[5, 6], [7, 8]]];
 
 function flattenArray2d(arr) {
   // Replace this comment and the next line with your code
-  console.log(arr);
+  const newArr = [];
+  arr.map(element1 => element1.map(element2 => newArr.push(element2)));
```
yash-kapila: How about trying out the same with `reduce`?

Week3/homework/step2-6.js
```diff
@@ -5,12 +5,16 @@ const arr3d = [[[1, 2], [3, 4]], [[5, 6], [7, 8]]];
...
+  return newArr;
 }
 
 function flattenArray3d(arr) {
   // Replace this comment and the next line with your code
-  console.log(arr);
+  const newArray = [];
+  arr.map(item1 => item1.map(item2 => item2.map(item3 => newArray.push(item3))));
```
yash-kapila: Logic works fine but imagine if you have to do this for a 4-D or a 5-D array. What happens in that case? Suggestion: try reading about a programming concept called recursion and see if you can apply the same here. 🙂

Week3/homework/step2-7.js
```diff
@@ -21,3 +21,9 @@ f2(y);
 console.log(y);
 
 // Add your explanation as a comment here
+
+// in f1(x) ==> x not a object, it's only passing the value (9).
+// the value of default variable x is not change by function, it's still 9.
+
+// in f2(y) ==> JavaScript is call by value, in case of objects the value being passed is the reference.
+// x manipulated by function and it's be 10.
```
yash-kapila: 👏

Week3/homework/step3-bonus.js
```diff
@@ -4,7 +4,14 @@ const values = ['a', 'b', 'c', 'd', 'a', 'e', 'f', 'c'];
...
   // Replace this comment and the next line with your code
-  console.log(arr);
+  for (let a = 0; a < arr.length; a++) {
+    for (let i = 0; i < arr.length; i++) {
+      if (a !== i && arr[a] === arr[i]) {
+        arr.splice(i, 1);
+      }
+    }
+  }
+  return arr;
```
yash-kapila: Aahaan. How about exploring `Sets`(https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set) in JavaScript. I am sure your code would reduce significantly afterwards 😄

Week3/homework/step3.js
```diff
@@ -2,7 +2,12 @@
 
 function createBase(base) {
   // Replace this comment and the next line with your code
-  console.log(base);
+  // console.log(base);
+  const num1 = base;
+  // return function name(x) {
+  //   return x + num1;
+  // };
+  return x => x + num1;
```
yash-kapila: You need not have created a new variable `num1` here. I think you can use `base` instead without a problem.

Week3/homework/step2-3.js
husenjanobul: Thanks for your feedback .when i have checked without `if ` ( npm run test-week3 ) , had an error that -

> step2-3 with do-while-loop › num = -2
so i added this `if` check.

Week3/homework/step2-4.js
husenjanobul: got it.

Week3/homework/step3.js
husenjanobul: yes, i don't need `num1`.

PR #483, opened 2019-09-11
Week3

rabrad:
For review

### DIFF
--- a/Week3/homework/step2-1.js
+++ b/Week3/homework/step2-1.js
@@ -1,9 +1,7 @@
 'use strict';
 
 function foo(func) {
-  // What to do here?
-  // Replace this comment and the next line with your code
-  console.log(func);
+  return func();
 }
 
 function bar() {
--- a/Week3/homework/step2-2.js
+++ b/Week3/homework/step2-2.js
@@ -5,10 +5,13 @@ function threeFive(startIndex, stopIndex, threeCallback, fiveCallback) {
 
   for (let number = startIndex; number <= stopIndex; number++) {
     numbers.push(number); // This create an array [10,11,12,13,14,15]
-  }
-  // Set the conditions of the callback for threeCallback, fiveCallback.
-  for (let i = 0; i < numbers.length; i++) {
-    const number = numbers[i];
+
+    if (number % 5 === 0 && number % 3 === 0) {
+      threeCallback(number);
+      fiveCallback(number);
+      return; // to stop repeating the log later on.
+    }
+
     if (number % 3 === 0) {
       threeCallback(number);
     }
--- a/Week3/homework/step2-2.js
+++ b/Week3/homework/step2-2.js
@@ -3,18 +3,27 @@
 function threeFive(startIndex, stopIndex, threeCallback, fiveCallback) {
   const numbers = [];
 
-  // Replace this comment and the next line with your code
-  console.log(startIndex, stopIndex, threeCallback, fiveCallback, numbers);
+  for (let number = startIndex; number <= stopIndex; number++) {
+    numbers.push(number); // This create an array [10,11,12,13,14,15]
+  }
+  // Set the conditions of the callback for threeCallback, fiveCallback.
+  for (let i = 0; i < numbers.length; i++) {
+    const number = numbers[i];
+    if (number % 3 === 0) {
+      threeCallback(number);
+    }
+    if (number % 5 === 0) {
+      fiveCallback(number);
+    }
+  }
 }
 
 function sayThree(number) {
-  // Replace this comment and the next line with your code
-  console.log(number);
+  console.log(number + ' is divisible by 3');
 }
 
 function sayFive(number) {
-  // Replace this comment and the next line with your code
-  console.log(number);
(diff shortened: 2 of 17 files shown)


Week3/homework/step2-1.js
```diff
@@ -1,9 +1,7 @@
 'use strict';
 
 function foo(func) {
-  // What to do here?
-  // Replace this comment and the next line with your code
-  console.log(func);
+  return func();
```

Week3/homework/step2-2.js
```diff
@@ -3,18 +3,27 @@
...
+  // Set the conditions of the callback for threeCallback, fiveCallback.
+  for (let i = 0; i < numbers.length; i++) {
+    const number = numbers[i];
+    if (number % 3 === 0) {
+      threeCallback(number);
+    }
+    if (number % 5 === 0) {
+      fiveCallback(number);
+    }
+  }
```
yash-kapila: Ah. Couple things here:
1. You don't need to have two separate loops for adding numbers to the array and then running the `if-else` condition. You could do both operations in a single loop.
2. You missed the requirement of printing when a number is divisible by both 3 and 5. Could you please add the same? **Hint**: a number divisible by both 3 and 5 will be divisible by 15 too. 😉

Week3/homework/step2-3.js
```diff
@@ -30,10 +31,15 @@ console.log('while', repeatStringNumTimesWithWhile('abc', 3));
...
-
+  let i = 0;
+  if (num <= 0) {
+    result += '';
+  } else {
+    do {
+      i++;
+      result += str;
+    } while (i < num);
+  }
```
yash-kapila: Any reason why you didn't add the `num <= 0` check to the `for` and `while` loops? 🙂

Week3/homework/step2-4.js
```diff
@@ -1,7 +1,9 @@
 'use strict';
 
 function Dog() {
-  // add your code here
+  this.name = 'DJ';
+  this.color = 'White';
+  this.numLegs = 4;
 }
```

```
  this.name = name;
  this.color = color;
  this.numLegs = legs;
}

// and so on
```

Week3/homework/step2-5.js
```diff
@@ -4,8 +4,11 @@ function multiplyAll(arr) {
   // eslint-disable-next-line
   let product = 1;
 
-  // Replace this comment and the next line with your code
-  console.log(arr, product);
+  for (let i = 0; i < arr.length; i++) {
+    for (let j = 0; j < arr[i].length; j++) {
+      product *= arr[i][j];
+    }
+  }
```

Week3/homework/step2-6.js
```diff
@@ -4,15 +4,37 @@ const arr2d = [[1, 2], [3, 4], [5, 6]];
...
+/* this can be achieved also by using flat() method.
+  function flattenArray2d(){
+  return arr2d.flat(); // flattens down 1 dimension
+  }
+
+  function flattenArray3d(){
+  return arr3d.flat(2); // flattens down 2 dimension
+  }
+*/
+
```
yash-kapila: Quite nice. I wasn't aware of `flat` method before. A good idea to share this with your fellow students and mentors like me who weren't aware of it in the Slack channel. 😜

Week3/homework/step2-7.js
```diff
@@ -21,3 +21,17 @@ f2(y);
...
+A distinct copy of the value is passed to the function; changes to it have no effect outside the function. 
+therefor console.log(x); // will log 9 
+ 
+
+Second snippet: (Passing an objects).
+A reference to the object value is passed to the function. 
+If the function modifies the values through the passed reference, the modification 
+is visible outside the function. Therefor console.log(y); // will log 10
+
+Ref. 1: https://stackoverflow.com/questions/518000/is-javascript-a-pass-by-reference-or-pass-by-value-language#3638034
```

Week3/homework/step3-bonus.js
```diff
@@ -3,12 +3,15 @@
...
-  console.log(arr);
+  arr = arr.filter((item, index, array) => array.indexOf(item) === index);
+  return arr;
 }
 
 const uniqueValues = makeUnique(values);
 console.log(uniqueValues);
+/* This can be du=one using Set to store unique values as:
+let uniqueValues = [...new Set(values)];
+*/
```
yash-kapila: 👏 - I think using Sets here is a better idea than a `filter` and `indexOf` operation 🙂

Week3/homework/step3.js
```diff
@@ -1,8 +1,9 @@
 'use strict';
 
 function createBase(base) {
-  // Replace this comment and the next line with your code
-  console.log(base);
+  return function(num) {
+    return num + base;
```

Week3/homework/step2-4.js
rabrad: Thanks for the explanation. I updated my file accordingly.

Week3/homework/step2-3.js
rabrad: I forgot to set the condition for negative values in both for and while. 
The file has been updated.

Week3/homework/step2-2.js
rabrad: Thanks for the hints. :)

PR #484, opened 2019-09-11
Week3

ahmthsrvsmz:


Week3/homework/step2-2.js
```diff
@@ -3,21 +3,30 @@
...
-  console.log(startIndex, stopIndex, threeCallback, fiveCallback, numbers);
+  for (let i = 0; i <= stopIndex - startIndex; i++) {
+    numbers.push(startIndex + i);
+    if ((startIndex + i) % 3 === 0 && (startIndex + i) % 5 === 0) {
+      numbers.push(threeCallback(startIndex + i));
+      numbers.push(fiveCallback(startIndex + i));
+    } else if ((startIndex + i) % 3 === 0) {
+      numbers.push(threeCallback(startIndex + i));
+    } else if ((startIndex + i) % 5 === 0) {
+      numbers.push(fiveCallback(startIndex + i));
```
yash-kapila: 👍 for logic. A couple of small suggestions though: 

1. If a number is divisible by both 3 and 5 then that number will also be divisible by 15, isn't it? 😉 Thus, you could also perhaps write the first if condition as `if ((startIndex + i) % 15 === 0) {}`
2. You need not complicate the `for` loop conditions. It could be as simple as `for(let i = startIndex; i <= stopIndex; i++) { // loop body }` 🙂

Week3/homework/step2-3.js
```diff
@@ -4,10 +4,12 @@
...
-
-  // Replace this comment and the next line with your code
-  console.log(str, num, result);
-
+  if (num > 0) {
+    for (num > 0; num--; ) {
+      result += str;
+    }
+    console.log(str, num, result);
+  }
```
yash-kapila: Couple things here:
1. The `for` loop conditions again seem a bit complicated than it should be. You could again simply write it as: `for(let i = num; i > 0; i--) { // loop body }`
2. The `console.log` need not be there.

Week3/homework/step2-3.js
```diff
@@ -17,10 +19,13 @@ console.log('for', repeatStringNumTimesWithFor('abc', 3));
...
-  // Replace this comment and the next line with your code
-  console.log(str, num, result);
-
+  if (num > 0) {
+    while (num > 0) {
+      result += str;
+      num--;
+    }
+    console.log(str, num, result);
+  }
```
yash-kapila: 👍

Week3/homework/step2-3.js
```diff
@@ -30,10 +35,14 @@ console.log('while', repeatStringNumTimesWithWhile('abc', 3));
...
-  console.log(str, num, result);
-
+  if (num > 0) {
+    do {
+      result += str;
+      num--;
+    } while (num > 0);
+
+    console.log(str, num, result);
+  }
```
yash-kapila: 👍

Week3/homework/step2-4.js
```diff
@@ -1,10 +1,14 @@
 'use strict';
 
 function Dog() {
-  // add your code here
+  this.name = 'Lesly';
+  this.color = 'Brown';
+  this.numLegs = 4;
```

```
  this.name = name;
  this.color = color;
  this.numLegs = legs;
}

// and so on
```

Week3/homework/step2-5.js
```diff
@@ -4,7 +4,11 @@ function multiplyAll(arr) {
   // eslint-disable-next-line
   let product = 1;
 
-  // Replace this comment and the next line with your code
+  for (let i = 0; i < arr.length; i++) {
+    for (let j = 0; j < arr[i].length; j++) {
+      product = product * arr[i][j];
```

Week3/homework/step2-6.js
```diff
@@ -2,15 +2,28 @@
...
-  console.log(arr);
+  for (let k = 0; k < arr.length; k++) {
+    for (let l = 0; l < arr[k].length; l++) {
+      for (let m = 0; m < arr[k][l].length; m++) {
+        flattenedArr3d.push(arr[k][l][m]);
+      }
+    }
+  }
+
+  return flattenedArr3d;
```
yash-kapila: Logic works fine but imagine if you have to do this for a 4-D or a 5-D array. What happens in that case? Suggestion: try reading about a programming concept called recursion and see if you can apply the same here. Also, look for `Array.prototype.flat` method on MDN. 🙂

Week3/homework/step2-7.js
```diff
@@ -18,6 +18,7 @@ function f2(val) {
 
 f2(y);
 
-console.log(y);
+console.log(y); // {x: 10}
 
-// Add your explanation as a comment here
+// x is a constant number. Function creates a block scoped x variable. X can chance in block scope.
+// y is an object. Function can change x key value. Object keys and values mutable.
```
yash-kapila: I don't think that is exactly what's happening here. As mentioned in the homework, look for call by value vs call by reference in JavaScript. 🙂

Week3/homework/step3-bonus.js
```diff
@@ -3,8 +3,8 @@
 const values = ['a', 'b', 'c', 'd', 'a', 'e', 'f', 'c'];
 
 function makeUnique(arr) {
-  // Replace this comment and the next line with your code
-  console.log(arr);
+  arr = [...new Set(arr)];
```
yash-kapila: 👏 for using Sets. But avoid mutating/changing values of function arguments. Instead, create a local variable in the function as:

```
const uniqueArr = [...new Set(arr)];
return uniqueArr;
```

Week3/homework/step3.js
```diff
@@ -1,8 +1,9 @@
 'use strict';
 
 function createBase(base) {
-  // Replace this comment and the next line with your code
-  console.log(base);
+  return function(num) {
+    return base + num;
```

Week3/homework/step2-2.js
ahmthsrvsmz: 1. Good suggestion. I will change it. 👍 😉
2. I defined "i" just for a loop because of started from 0 and finished (stopIndex- startIndex).  It is the same way but little complicated. I will change thanks

Week3/homework/step2-3.js
ahmthsrvsmz: 1. I defined "num" instead of "i" but not problem I can change.
2. console.log just for checking on the browser. I will delete it

Week3/homework/step2-6.js
ahmthsrvsmz: I will look flat method. 👍

Week3/homework/step2-7.js
ahmthsrvsmz: I will look again

Week3/homework/step2-6.js
ahmthsrvsmz: I will learn flat concept

Week3/homework/step3-bonus.js
ahmthsrvsmz: okay 👍

PR #485, opened 2019-09-11
submit js2 week3 solutions- Abdullah

abdullahkanatsiz7:
my homework js2week3

### DIFF
--- a/Week3/homework/step2-3.js
+++ b/Week3/homework/step2-3.js
@@ -4,10 +4,9 @@
 function repeatStringNumTimesWithFor(str, num) {
   // eslint-disable-next-line prefer-const
   let result = '';
-
-  // Replace this comment and the next line with your code
-  console.log(str, num, result);
-
+  for (let x = 0; x < num; x++) {
+    result += str;
+  }
   return result;
 }
 
@@ -18,8 +17,11 @@ function repeatStringNumTimesWithWhile(str, num) {
   // eslint-disable-next-line prefer-const
   let result = '';
 
-  // Replace this comment and the next line with your code
-  console.log(str, num, result);
+  let x = 0;
+  while (x < num) {
+    x++;
+    result += str;
+  }
 
   return result;
 }
@@ -30,9 +32,13 @@ console.log('while', repeatStringNumTimesWithWhile('abc', 3));
 function repeatStringNumTimesWithDoWhile(str, num) {
   // eslint-disable-next-line prefer-const
   let result = '';
+  if (num <= 0) return result;
+  let z = 0;
 
-  // Replace this comment and the next line with your code
-  console.log(str, num, result);
+  do {
+    z++;
+    result += str;
+  } while (z < num);
 
   return result;
 }
--- a/Week3/homework/step2-4.js
+++ b/Week3/homework/step2-4.js
@@ -1,7 +1,9 @@
 'use strict';
 
 function Dog() {
-  // add your code here
+  this.name = 'karabas';
+  this.color = 'black';
+  this.numLegs = 4;
 }
 
 const hound = new Dog();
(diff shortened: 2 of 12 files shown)


Week3/homework/step2-3.js
```diff
@@ -30,9 +32,13 @@ console.log('while', repeatStringNumTimesWithWhile('abc', 3));
...
   let result = '';
+  if (num <= 0) return result;
+  let z = 0;
 
-  // Replace this comment and the next line with your code
-  console.log(str, num, result);
+  do {
+    z++;
+    result += str;
+  } while (z < num);
```
yash-kapila: 👍 for logic but I wonder why did you not have the initial `if` condition inside `for` and `while` loop implementations? 🤔

Week3/homework/step2-4.js
```diff
@@ -1,7 +1,9 @@
 'use strict';
 
 function Dog() {
-  // add your code here
+  this.name = 'karabas';
+  this.color = 'black';
+  this.numLegs = 4;
 }
```

```
  this.name = name;
  this.color = color;
  this.numLegs = legs;
}

// and so on
```

Week3/homework/step2-5.js
```diff
@@ -3,8 +3,11 @@
 function multiplyAll(arr) {
   // eslint-disable-next-line
   let product = 1;
-
-  // Replace this comment and the next line with your code
+  for (let i = 0; i < arr.length; i++) {
+    for (let j = 0; j < arr[i].length; j++) {
+      product *= arr[i][j];
+    }
+  }
```

Week3/homework/step2-6.js
```diff
@@ -4,13 +4,15 @@ const arr2d = [[1, 2], [3, 4], [5, 6]];
 const arr3d = [[[1, 2], [3, 4]], [[5, 6], [7, 8]]];
 
 function flattenArray2d(arr) {
-  // Replace this comment and the next line with your code
-  console.log(arr);
+  return arr.reduce((acc, value) => acc.concat(value), []);
+  // another solution (spread): return arr.reduce((acc, value) => [...acc, ...value], [])
```

Week3/homework/step2-6.js
```diff
@@ -4,13 +4,15 @@ const arr2d = [[1, 2], [3, 4], [5, 6]];
...
 }
 
 function flattenArray3d(arr) {
-  // Replace this comment and the next line with your code
-  console.log(arr);
+  // eslint-disable-next-line arrow-body-style
+  return arr.reduce((acc, value) => {
+    return acc.concat(Array.isArray(value) ? flattenArray3d(value) : value);
+  }, []);
```
yash-kapila: Logic works fine but imagine if you have to do this for a 4-D or a 5-D array. What happens in that case? Suggestion: try reading about a programming concept called recursion and see if you can apply the same here. Also, try reading about the method `Array.prototype.flat` on MDN. 🙂

Week3/homework/step2-7.js
```diff
@@ -20,4 +20,13 @@ f2(y);
 
 console.log(y);
 
-// Add your explanation as a comment here
+/* In javascript it's always pass by reference.when a variable refers to an object which includes array,
```
yash-kapila: 👍 but I think there is a typo - In javascript it's always pass by **value**. When a variable refers to an object, the value is the reference to the object.

Week3/homework/step3-bonus.js
```diff
@@ -3,8 +3,7 @@
 const values = ['a', 'b', 'c', 'd', 'a', 'e', 'f', 'c'];
 
 function makeUnique(arr) {
-  // Replace this comment and the next line with your code
-  console.log(arr);
+  return [...new Set(arr)];
```
yash-kapila: 👍

Week3/homework/step3.js
```diff
@@ -1,8 +1,7 @@
 'use strict';
 
 function createBase(base) {
-  // Replace this comment and the next line with your code
-  console.log(base);
+  return add => add + base;
```

PR #486, opened 2019-09-11
Homework week3 msonmez

msnmz:


Week3/homework/step2-6.js
```diff
@@ -4,13 +4,14 @@ const arr2d = [[1, 2], [3, 4], [5, 6]];
...
+  return arr.reduce((flattenedArr, currentArr) => [...flattenedArr, ...currentArr], []);
 }
 
 function flattenArray3d(arr) {
-  // Replace this comment and the next line with your code
-  console.log(arr);
+  return arr.reduce(
+    (flattenedArr, current2dArr) => [...flattenedArr, ...flattenArray2d(current2dArr)],
+    [],
+  );
```
yash-kapila: 👍 for logic and using `flattenArray2d` inside `flattenArray3d`. However, I would recommend that you also read about a topic in programming called `recursion`.

Week3/homework/step2-7.js
```diff
@@ -21,3 +21,19 @@ f2(y);
...
+ *
+ * For function f2(y):
+ *  y is type of object, and it is passed by reference to the function f2.
+ *  The const declaration creates a read-only reference to a value.
+ *  It does not mean the value it holds is immutable, just that the variable identifier cannot be reassigned.
+ *  So, if the const declared variable is of type object, we can change the objects properties.
+ *  And y has a property named x. When we call function f2, and pass y as parameter, it is passed by reference.
+ *  And f2 function changes the property of the y, so there is no conflict with const declaration.
+ *  Because, y is passed by reference to the function f2, f2 changes the properties of actual y, not its copy.
+ */
```
If you want to further test this out, you could have a look at:
```
function changeStuff(a,b,c,d) {
  a = a * 10;
  b.item = "changed";
  c = { item: "changed" };
  d.push(40);
}

const a = 10;
const b = { item: "unchanged" };
const c = { item: "unchanged" };
const d = [10,20,30];

changeStuff(a,b,c,d);

console.log(a,b,c,d); // what do you think would be printed here? Test it out ;)
```

Week3/homework/step3-bonus.js
```diff
@@ -3,8 +3,11 @@
 const values = ['a', 'b', 'c', 'd', 'a', 'e', 'f', 'c'];
 
 function makeUnique(arr) {
-  // Replace this comment and the next line with your code
-  console.log(arr);
+  return arr.reduce(
+    (uniquesArr, currentElm) =>
+      uniquesArr.includes(currentElm) ? uniquesArr : [...uniquesArr, currentElm],
+    [],
+  );
```
yash-kapila: Would suggest that you had a look at a new topic in JavaScript called `Sets` 🙂

Week3/homework/step3.js
```diff
@@ -1,8 +1,7 @@
 'use strict';
 
 function createBase(base) {
-  // Replace this comment and the next line with your code
-  console.log(base);
+  return num => num + base;
```

Week3/homework/step2-3.js
```diff
@@ -31,8 +34,12 @@ function repeatStringNumTimesWithDoWhile(str, num) {
...
   let result = '';
 
-  // Replace this comment and the next line with your code
-  console.log(str, num, result);
+  if (num <= 0) return result;
+
+  do {
+    result += str;
+    --num;
+  } while (num > 0);
```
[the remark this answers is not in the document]
msnmz: Yes, I made it intentionally because, if `number < 0`, `for loop` and the `while loop `won't be executed at all. But `do-while loop` will be executed at least once, so I need to check the value before `do-while loop`, but others have this check internally.

Week3/homework/step2-2.js
```diff
@@ -3,8 +3,15 @@
...
-  console.log(startIndex, stopIndex, threeCallback, fiveCallback, numbers);
+  // Make array
+  for (let i = startIndex; i <= stopIndex; ++i) {
+    numbers.push(i);
+  }
+
+  numbers.forEach(number => {
+    if (number % 3 === 0) threeCallback(number);
+    if (number % 5 === 0) fiveCallback(number);
+  });
```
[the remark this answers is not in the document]
msnmz: Actually, I made the `if` statements separate and thought that if the `number is divisible by 15`, both will be evaluated as `true` and executed.
And for the second part, you are absolutely right, there are redundant codes, I fixed them.

Week3/homework/step2-4.js
```diff
@@ -1,7 +1,9 @@
 'use strict';
 
 function Dog() {
-  // add your code here
+  this.name = 'Dog';
+  this.color = 'yellow';
+  this.numLegs = 4;
 }
```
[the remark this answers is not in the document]
msnmz: This is a fatal mistake of me, really, I just realized that I hardcoded the values. They should be parameterized.

Week3/homework/step2-6.js
msnmz: I tried to make it recursive, I did it but there might be a better way because I didn't like my solution to be honest.

Week3/homework/step2-7.js
msnmz: The crucial part is, parameter c is not read, and after assigning a new value, it is not pointing to the value that was passed by reference, and it won't affect it.

Week3/homework/step3-bonus.js
msnmz: Great, learned a brand new subject.

msnmz:
Thank you very much @yash-kapila for your comments. I am very much appreciated. I really thank you!

PR #487, opened 2019-09-11
JS2 Week3 Homework

ihk2018:


Week3/homework/step2-2.js
```diff
@@ -3,18 +3,26 @@
...
+  }
+
+  numbers.forEach(number => {
+    if (number % 3 === 0) {
+      threeCallback(number);
+    }
+    if (number % 5 === 0) {
+      fiveCallback(number);
+    }
+  });
```
1. You can combine the two loop iterations for creating the numbers array and the `if` conditions.
2. I think you missed the requirement where you were also supposed to log on console if a number was divisible by both 3 and 5. Hint: a number which is divisible by both 3 and 5 will also be divisible by 15. 🙂

Week3/homework/step2-3.js
```diff
@@ -2,41 +2,36 @@
...
-
-  // Replace this comment and the next line with your code
-  console.log(str, num, result);
-
+  if (num > 0) {
+    do {
+      result += str;
+      num--;
+    } while (num > 0);
+  }
```
yash-kapila: Any reason why you added this if check here in the do-while loop and not for or while loop? 🙂

Week3/homework/step2-4.js
```diff
@@ -1,10 +1,13 @@
 'use strict';
 
 function Dog() {
-  // add your code here
+  this.name = 'Lassie';
+  this.color = 'white';
+  this.numLegs = 4;
 }
```

```
  this.name = name;
  this.color = color;
  this.numLegs = legs;
}

// and so on
```

Week3/homework/step2-5.js
```diff
@@ -1,12 +1,13 @@
...
 function multiplyAll(arr) {
-  // eslint-disable-next-line
   let product = 1;
 
-  // Replace this comment and the next line with your code
-  console.log(arr, product);
-
+  for (let i = 0; i < arr.length; i++) {
+    for (let x = 0; x < arr[i].length; x++) {
+      product = product * arr[i][x];
```

Week3/homework/step2-6.js
```diff
@@ -4,17 +4,19 @@ const arr2d = [[1, 2], [3, 4], [5, 6]];
...
+  arr.map(arrFirst => arrFirst.map(arrSecond => newArray.push(arrSecond)));
+  return newArray;
 }
 
 function flattenArray3d(arr) {
-  // Replace this comment and the next line with your code
-  console.log(arr);
+  const newArray = [];
+  arr.map(newArr1 => newArr1.map(newArr2 => newArr2.map(newArr3 => newArray.push(newArr3))));
+  return newArray;
```
yash-kapila: Logic works fine but imagine if you have to do this for a 4-D or a 5-D array. What happens in that case? Suggestion: try reading about a programming concept called recursion and see if you can apply the same here. 🙂 

Also, have a look at `Array.prototype.flat` method which is newly introduced in MDN.

Week3/homework/step2-2.js
ihk2018: function threeFive(startIndex, stopIndex, threeCallback, fiveCallback) {
  const numbers = [];

  for (let i = startIndex; i <= stopIndex; i++) {
    numbers.push(i);
  }

  numbers.forEach(number => {
    if (number % 15 === 0) {
      console.log(`${number}, 'can be divided by 3, 5 and 15'`);
    } else if (number % 3 === 0) {
      threeCallback(number);
    } else if (number % 5 === 0) {
      fiveCallback(number);
    }
  });
}

function sayThree(number) {
  console.log(`${number}, 'can be divided by 3'`);
}

function sayFive(number) {
  console.log(`${number}, 'can be divided by 5'`);
}

threeFive(10, 15, sayThree, sayFive);

ihk2018: i wrote something like that i works , but I am not sure if it's eneoug for the homework requirement

Week3/homework/step2-3.js
ihk2018: Because , for and while loops check the conditions before executing the code. In do - while loop, it executes the condtions atleast once, therefore I wanted to add if condition for preventing do execute the code.

Week3/homework/step2-4.js
ihk2018: Thank you Yash, i rewrite the code based on your recommodation

Week3/homework/step2-6.js
ihk2018: Thak you for the advice, i rewite the code

PR #488, opened 2019-09-11
Add Js2-week3 Homework

EsraZengn:
Here is my JavaScript2 Week3 homework.

### DIFF
--- a/Week3/homework/step2-1.js
+++ b/Week3/homework/step2-1.js
@@ -1,9 +1,7 @@
 'use strict';
 
 function foo(func) {
-  // What to do here?
-  // Replace this comment and the next line with your code
-  console.log(func);
+  func();
 }
 
 function bar() {
--- a/Week3/homework/step2-2.js
+++ b/Week3/homework/step2-2.js
@@ -3,18 +3,30 @@
 function threeFive(startIndex, stopIndex, threeCallback, fiveCallback) {
   const numbers = [];
 
-  // Replace this comment and the next line with your code
-  console.log(startIndex, stopIndex, threeCallback, fiveCallback, numbers);
+  const count = stopIndex - startIndex + 1;
+  for (let i = 0; i < count; i++) {
+    numbers.push(startIndex);
+    startIndex++;
+  }
+
+  for (const item of numbers) {
+    if (item % 3 === 0 && item % 5 === 0) {
+      threeCallback(item);
+      fiveCallback(item);
+    } else if (item % 3 === 0) {
+      threeCallback(item);
+    } else if (item % 5 === 0) {
+      fiveCallback(item);
+    }
+  }
 }
 
 function sayThree(number) {
-  // Replace this comment and the next line with your code
-  console.log(number);
+  console.log(`${number} is divisible by 3.`);
 }
 
 function sayFive(number) {
-  // Replace this comment and the next line with your code
-  console.log(number);
+  console.log(`${number} is divisible by 5.`);
 }
 
 threeFive(10, 15, sayThree, sayFive);
--- a/Week3/homework/step2-3.js
+++ b/Week3/homework/step2-3.js
@@ -5,8 +5,13 @@ function repeatStringNumTimesWithFor(str, num) {
   // eslint-disable-next-line prefer-const
   let result = '';
 
-  // Replace this comment and the next line with your code
-  console.log(str, num, result);
+  if (num > 0) {
+    for (let i = 0; i < num; i++) {
+      result = result + str;
+    }
+  } else {
+    result = '';
+  }
 
   return result;
 }
@@ -18,8 +23,15 @@ function repeatStringNumTimesWithWhile(str, num) {
   // eslint-disable-next-line prefer-const
   let result = '';
 
-  // Replace this comment and the next line with your code
-  console.log(str, num, result);
+  if (num > 0) {
+    let i = 0;
+    while (i < num) {
+      result = result + str;
+      i++;
+    }
+  } else {
+    result = '';
+  }
 
   return result;
 }
@@ -30,9 +42,15 @@ console.log('while', repeatStringNumTimesWithWhile('abc', 3));
 function repeatStringNumTimesWithDoWhile(str, num) {
   // eslint-disable-next-line prefer-const
   let result = '';
-
-  // Replace this comment and the next line with your code
-  console.log(str, num, result);
+  if (num > 0) {
+    let i = 0;
+    do {
+      result = result + str;
+      i++;
+    } while (i < num);
+  } else {
+    result = '';
+  }
 
   return result;
 }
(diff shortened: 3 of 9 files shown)


Week3/homework/step2-1.js
```diff
@@ -1,9 +1,7 @@
 'use strict';
 
 function foo(func) {
-  // What to do here?
-  // Replace this comment and the next line with your code
-  console.log(func);
+  func();
```

Week3/homework/step2-2.js
```diff
@@ -3,18 +3,30 @@
...
+  for (const item of numbers) {
+    if (item % 3 === 0 && item % 5 === 0) {
+      threeCallback(item);
+      fiveCallback(item);
+    } else if (item % 3 === 0) {
+      threeCallback(item);
+    } else if (item % 5 === 0) {
+      fiveCallback(item);
+    }
+  }
```
yash-kapila: Ah. I think you can improve the logic here a bit more. For example, if you think mathematically, a number which is divisible by both 3 and 5 will always be divisible by number 15, isn't it? Something like:

```
if (i % 15 === 0) {
  // call both three and five callbacks
} else if(i % 3 === 0) {
  // call three callbacks
} else if(i % 5 === 0) {
  // call five callbacks
}
```

Also, you need not have two separate loops for creating the `numbers` array and then for logging on console. You could do it in the same loop as well.

Also, you could rewrite your first loop as:

```
  for (let i = startIndex; i < stopIndex; i++) {
    numbers.push(i);
  }
``` 

🙂

Week3/homework/step2-3.js
```diff
@@ -5,8 +5,13 @@ function repeatStringNumTimesWithFor(str, num) {
...
   let result = '';
 
-  // Replace this comment and the next line with your code
-  console.log(str, num, result);
+  if (num > 0) {
+    for (let i = 0; i < num; i++) {
+      result = result + str;
+    }
+  } else {
+    result = '';
```
yash-kapila: 👏

Week3/homework/step2-3.js
```diff
@@ -18,8 +23,15 @@ function repeatStringNumTimesWithWhile(str, num) {
...
-  console.log(str, num, result);
+  if (num > 0) {
+    let i = 0;
+    while (i < num) {
+      result = result + str;
+      i++;
+    }
+  } else {
+    result = '';
+  }
```
yash-kapila: 👏

Week3/homework/step2-3.js
```diff
@@ -30,9 +42,15 @@ console.log('while', repeatStringNumTimesWithWhile('abc', 3));
...
-  console.log(str, num, result);
+  if (num > 0) {
+    let i = 0;
+    do {
+      result = result + str;
+      i++;
+    } while (i < num);
+  } else {
+    result = '';
+  }
```
yash-kapila: 👏

Week3/homework/step2-4.js
```diff
@@ -1,7 +1,9 @@
 'use strict';
 
 function Dog() {
-  // add your code here
+  this.name = 'Pachinko';
+  this.color = 'Gray';
+  this.numLegs = 4;
```

```
  this.name = name;
  this.color = color;
  this.numLegs = legs;
}

// and so on
```

Week3/homework/step2-5.js
```diff
@@ -4,9 +4,11 @@ function multiplyAll(arr) {
...
   let product = 1;
 
-  // Replace this comment and the next line with your code
-  console.log(arr, product);
-
+  for (let i = 0; i < arr.length; i++) {
+    for (let j = 0; j < arr[i].length; j++) {
+      product = product * arr[i][j];
+    }
+  }
```

Week3/homework/step2-6.js
```diff
@@ -4,13 +4,28 @@ const arr2d = [[1, 2], [3, 4], [5, 6]];
...
 
 function flattenArray2d(arr) {
-  // Replace this comment and the next line with your code
-  console.log(arr);
+  const flattenedArr2d = [];
+
+  for (let i = 0; i < arr.length; i++) {
+    for (let j = 0; j < arr[i].length; j++) {
+      flattenedArr2d.push(arr[i][j]);
+    }
```

Week3/homework/step2-6.js
```diff
@@ -4,13 +4,28 @@ const arr2d = [[1, 2], [3, 4], [5, 6]];
...
 
 function flattenArray3d(arr) {
   // Replace this comment and the next line with your code
-  console.log(arr);
+  const flattenedArr3d = [];
+
+  for (let i = 0; i < arr.length; i++) {
+    for (let j = 0; j < arr[i].length; j++) {
+      for (let k = 0; k < arr[i][j].length; k++) {
+        flattenedArr3d.push(arr[i][j][k]);
```
yash-kapila: Logic works fine but imagine if you have to do this for a 4-D or a 5-D array. What happens in that case? Suggestion: try reading about a programming concept called recursion and see if you can apply the same here. Also, look up for a new method in JavaScript named `Array.prototype.flat`.  🙂

Week3/homework/step2-7.js
```diff
@@ -20,4 +20,20 @@ f2(y);
 
 console.log(y);
 
-// Add your explanation as a comment here
+/* With the first console log, the value of the x will be printed as 9. Firstly, the value of x 
+is set to 9 which is a number and thus primitive data type. Primitive data types cannot be 
+reassigned when defined as const which is the case here. Even if x is defined as 'let', the value of 
+it would still be 9 since the value of x is passed into the function f1 and x is not assigned to a 
+new value. The value of x would have changed if we wrote it like this:
```
yash-kapila: What you have written for primitive is correct but the actual reason why value of `x` doesn't change after the function is because JavaScript as a language is call by value. This means that the value of `x` is passed to the function and not the variable itself. Which is why modifying the value inside the function wouldn't have an impact on the variable which was passed as argument.

As far as objects in JavaScript is concerned, what is passed to the function is not the value but the reference of the object. Thus, if we modify it inside the function, it modifies the variable itself which is also pointing to the same reference. Hope it makes sense. 🙂

Week3/homework/step3-bonus.js
```diff
@@ -3,8 +3,16 @@
...
+  const sortedArr = arr.sort();
+  const uniqueArr = [];
+
+  for (let i = 0; i < sortedArr.length; i++) {
+    if (sortedArr[i] !== sortedArr[i + 1]) {
+      uniqueArr.push(sortedArr[i]);
+    }
+  }
+
+  return uniqueArr;
```
yash-kapila: Aahaan. This is a lot of code, isn't it? I would suggest reading about **Sets** in JavaScript and see if it helps in reducing the code required 🙂

Week3/homework/step3.js
```diff
@@ -1,8 +1,7 @@
 'use strict';
 
 function createBase(base) {
-  // Replace this comment and the next line with your code
-  console.log(base);
+  return addend => base + addend;
```

PR #489, opened 2019-09-11
Js2w3 Homework

ersarue:


Week3/homework/step2-1.js
```diff
@@ -1,9 +1,7 @@
 'use strict';
 
 function foo(func) {
-  // What to do here?
-  // Replace this comment and the next line with your code
-  console.log(func);
+  return func();
```

Week3/homework/step2-2.js
```diff
@@ -2,21 +2,25 @@
...
+    if (i % 3 === 0 && i % 5 === 0) {
+      threeCallback(i);
+      fiveCallback(i);
+    }
+    if (i % 3 === 0 && i % 5 !== 0) {
+      threeCallback(i);
+    } else if (i % 5 === 0 && i % 3 !== 0) {
+      fiveCallback(i);
+    }
+  }
```
yash-kapila: I don't think you need to have the if checks as:
```
if (i % 3 === 0 && i % 5 !== 0) { }
```
and
```
else if (i % 5 === 0 && i % 3 !== 0) { }
```

Once you have a check for `i % 3 === 0 && i % 5 === 0` at the beginning, you need not check for `i % 5 !== 0` and `i % 3 !== 0` as it is now redundant. 

Also, a number which is divisible by 3 and 5 both will also be divisible by 15, right? 😉 Can you think of making your first if condition shorter then?

Week3/homework/step2-3.js
```diff
@@ -30,9 +28,13 @@ console.log('while', repeatStringNumTimesWithWhile('abc', 3));
 function repeatStringNumTimesWithDoWhile(str, num) {
   // eslint-disable-next-line prefer-const
   let result = '';
-
-  // Replace this comment and the next line with your code
-  console.log(str, num, result);
+  if (num <= 0) {
+    return result;
```
yash-kapila: You could also add this `if` condition to `for` and `while` loop, right?

Week3/homework/step2-4.js
```diff
@@ -1,10 +1,12 @@
 'use strict';
 
-function Dog() {
-  // add your code here
+function Dog(name, color) {
+  this.name = name;
+  this.color = color;
+  this.numLegs = 4;
```

```
  this.name = name;
  this.color = color;
  this.numLegs = legs;
}

// and so on
```

Week3/homework/step2-5.js
```diff
@@ -4,8 +4,11 @@ function multiplyAll(arr) {
   // eslint-disable-next-line
   let product = 1;
 
-  // Replace this comment and the next line with your code
-  console.log(arr, product);
+  for (let i = 0; i < arr.length; i++) {
+    for (let j = 0; j < arr[i].length; j++) {
+      product = product * arr[i][j];
```

Week3/homework/step2-6.js
```diff
@@ -4,13 +4,11 @@ const arr2d = [[1, 2], [3, 4], [5, 6]];
...
 function flattenArray2d(arr) {
-  // Replace this comment and the next line with your code
-  console.log(arr);
+  return arr.flat();
 }
 
 function flattenArray3d(arr) {
-  // Replace this comment and the next line with your code
-  console.log(arr);
+  return arr.flat(Infinity);
```
yash-kapila: 👍. But since this method is quite new and wouldn't be supported by all browsers, I would suggest that you read about a topic in programming named as recursion.

Week3/homework/step2-7.js
```diff
@@ -20,4 +20,4 @@ f2(y);
 
 console.log(y);
 
-// Add your explanation as a comment here
+// in the first part, the defined variable is primitive and can't be mutated. reassignment, on line 5, works but its effect is limited to  block scope, thus not effecting value. it is not the same in the second part. here, on line 15, it is mutating, not reassigning. it's being passed by reference and mutating is reflected in 'y'.
```

Week3/homework/step2-2.js
yash-kapila: Also, I am wondering if you did work together with Faruk on the logic for this particular assignment as we see here - https://github.com/HackYourFuture/JavaScript2/blob/d05c3e0119b49a062edd1609948a22fb957504ab/Week3/homework/step2-2.js? 🙂

ersarue: > I don't think you need to have the if checks as:
> 
> ```
> if (i % 3 === 0 && i % 5 !== 0) { }
> ```
> 
> and
> 
> ```
> else if (i % 5 === 0 && i % 3 !== 0) { }
> ```
> 
> Once you have a check for `i % 3 === 0 && i % 5 === 0` at the beginning, you need not check for `i % 5 !== 0` and `i % 3 !== 0` as it is now redundant.
> 
> Also, a number which is divisible by 3 and 5 both will also be divisible by 15, right? 😉 Can you think of making your first if condition shorter then?

yes that's right, no need the second if statements as the first one does what's needed to be done. when busy with bunch of assignments in a late time, some repetition exists and you miss that. I'll be paying more attention

ersarue: > Also, I am wondering if you did work together with Faruk on the logic for this particular assignment as we see here - https://github.com/HackYourFuture/JavaScript2/blob/d05c3e0119b49a062edd1609948a22fb957504ab/Week3/homework/step2-2.js? 🙂

True. Some time we group up for a part of homework and try to find solution together. And to speak frankly, we just thought yesterday that we better notify instructors when it is the case

Week3/homework/step2-3.js
ersarue: hmm, when thinking now I see that's possible. I'll look into it

Week3/homework/step2-4.js
ersarue: Now with your explanation I clearly see that my code would not allowing modify number of legs :) You're right, I am changing it

Week3/homework/step2-6.js
ersarue: I am on this recursion topic

Week3/homework/step2-2.js
yash-kapila: Got it. Yes, I think it will be quite helpful for the reviewers going through the pull requests. Thanks. 🙂

PR #490, opened 2019-09-11
JS2 Week3 Hw

FarukTercan:


Week3/homework/step2-3.js
```diff
@@ -17,10 +16,13 @@ console.log('for', repeatStringNumTimesWithFor('abc', 3));
 function repeatStringNumTimesWithWhile(str, num) {
   // eslint-disable-next-line prefer-const
   let result = '';
-
-  // Replace this comment and the next line with your code
-  console.log(str, num, result);
-
+  if (num <= 0) {
+    return result;
+  }
```
yash-kapila: How about adding the `if` logic to the for loop as well? 🙂

Week3/homework/step2-4.js
```diff
@@ -1,7 +1,9 @@
 'use strict';
 
 function Dog() {
-  // add your code here
+  this.name = 'Max';
+  this.color = 'white';
+  this.numLegs = 4;
```

```
  this.name = name;
  this.color = color;
  this.numLegs = legs;
}

// and so on
```

Week3/homework/step2-5.js
```diff
@@ -3,9 +3,11 @@
 function multiplyAll(arr) {
   // eslint-disable-next-line
   let product = 1;
-
-  // Replace this comment and the next line with your code
-  console.log(arr, product);
+  for (let i = 0; i < arr.length; i++) {
+    for (let j = 0; j < arr[i].length; j++) {
+      product = product * arr[i][j];
```

Week3/homework/step2-6.js
```diff
@@ -4,13 +4,11 @@ const arr2d = [[1, 2], [3, 4], [5, 6]];
...
 function flattenArray2d(arr) {
-  // Replace this comment and the next line with your code
-  console.log(arr);
+  return arr.flat(1);
 }
 
 function flattenArray3d(arr) {
-  // Replace this comment and the next line with your code
-  console.log(arr);
+  return arr.flat(2);
```
yash-kapila: 👍. But since this method is quite new and wouldn't be supported by all browsers, I would suggest that you read about a topic in programming named as recursion.

Week3/homework/step2-7.js
```diff
@@ -20,4 +20,9 @@ f2(y);
 
 console.log(y);
 
-// Add your explanation as a comment here
+/*  const x in line 5 is a number,primitive and immutable. So the changes of x in the f1(x) cannot pass the block scope and cannot
+ change the value of const x since it also has a different memory location.
+ */
+/* On the other hand, y has a object value which is complex data type and mutable. 
+The changes in the f2(y) can effect the value of y.
```

Week3/homework/step3-bonus.js
```diff
@@ -3,8 +3,7 @@
 const values = ['a', 'b', 'c', 'd', 'a', 'e', 'f', 'c'];
 
 function makeUnique(arr) {
-  // Replace this comment and the next line with your code
-  console.log(arr);
+  return [...new Set(arr)];
```
yash-kapila: 👏

Week3/homework/step3.js
```diff
@@ -1,8 +1,7 @@
 'use strict';
 
 function createBase(base) {
-  // Replace this comment and the next line with your code
-  console.log(base);
+  return num => num + base;
```

Week3/homework/step2-2.js
```diff
@@ -2,19 +2,28 @@
...
+    if (i % 3 === 0 && i % 5 === 0) {
+      threeCallback(i);
+      fiveCallback(i);
+    }
+    if (i % 3 === 0 && i % 5 !== 0) {
+      threeCallback(i);
+    } else if (i % 5 === 0 && i % 3 !== 0) {
+      fiveCallback(i);
+    }
+  }
```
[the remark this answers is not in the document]
FarukTercan: Hi Yash, thanks for the feedback. As you recognized, we have a small group :) first, each one tries to complete hw alone. Then, we discuss a problem, which we get stuck on, to find a solution.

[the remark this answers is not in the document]
yash-kapila: Ah okay. I understand. But could you please let the reviewer know the next time you work together with someone on a homework assignment? It would be easier for that person while reviewing. 🙂

PR #491, opened 2019-09-11
Js2 week 3 homework

haha1906:
JS2 Week 3 homework is ready for check

### DIFF
--- a/Week3/homework/step2-1.js
+++ b/Week3/homework/step2-1.js
@@ -1,9 +1,7 @@
 'use strict';
 
 function foo(func) {
-  // What to do here?
-  // Replace this comment and the next line with your code
-  console.log(func);
+  return func();
 }
 
 function bar() {
--- a/Week3/homework/step2-4.js
+++ b/Week3/homework/step2-4.js
@@ -1,10 +1,15 @@
 'use strict';
 
 function Dog() {
-  // add your code here
+  this.name = 'Puppy';
+  this.color = 'brown';
+  this.numLegs = 4;
 }
 
 const hound = new Dog();
+hound.name = 'Miami';
+hound.color = 'pink';
+console.log(hound);
 
 // Do not change or remove anything below this line
 module.exports = hound;
--- a/Week3/homework/step2-6.js
+++ b/Week3/homework/step2-6.js
@@ -4,13 +4,25 @@ const arr2d = [[1, 2], [3, 4], [5, 6]];
 const arr3d = [[[1, 2], [3, 4]], [[5, 6], [7, 8]]];
 
 function flattenArray2d(arr) {
-  // Replace this comment and the next line with your code
-  console.log(arr);
+  const flattenedArray = [];
+  for (let i = 0; i < arr.length; i++) {
+    for (let j = 0; j < arr[i].length; j++) {
+      flattenedArray.push(arr[i][j]);
+    }
+  }
+  return flattenedArray;
 }
 
 function flattenArray3d(arr) {
-  // Replace this comment and the next line with your code
-  console.log(arr);
+  const flattened = [];
+  for (let i = 0; i < arr.length; i++) {
+    for (let j = 0; j < arr[i].length; j++) {
+      for (let k = 0; k < arr[i][j].length; k++) {
+        flattened.push(arr[i][j][k]);
+      }
+    }
+  }
+  return flattened;
 }
 
 console.log(flattenArray2d(arr2d)); // -> [1, 2, 3, 4, 5, 6]
--- a/Week3/homework/step2-7.js
+++ b/Week3/homework/step2-7.js
@@ -20,4 +20,12 @@ f2(y);
 
 console.log(y);
 
-// Add your explanation as a comment here
+// The first function incriminate variable of val 1 number.
+// Then returns the number as a variable.
+// In the console it is written 9, because x was a const variable.
+// That means x never changes.
+
+// The second function incriminate x property of an object called val.
+// Then returns the number as an object.
+// In the console it is written 10, because y was a const object.
+// That means y doesn't change, but the content of y can change.
--- a/Week3/homework/step3-bonus.js
+++ b/Week3/homework/step3-bonus.js
@@ -3,8 +3,14 @@
 const values = ['a', 'b', 'c', 'd', 'a', 'e', 'f', 'c'];
 
 function makeUnique(arr) {
-  // Replace this comment and the next line with your code
-  console.log(arr);
+  for (let i = 0; i < arr.length; i++) {
+    for (let j = 0; j < i; j++) {
+      if (arr[i] === arr[j]) {
+        arr.splice(i, 1);
+      }
+    }
+  }
+  return arr;
 }
 
 const uniqueValues = makeUnique(values);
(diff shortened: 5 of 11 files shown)


Week3/homework/step2-6.js
```diff
@@ -4,13 +4,25 @@ const arr2d = [[1, 2], [3, 4], [5, 6]];
...
 }
 
 function flattenArray3d(arr) {
-  // Replace this comment and the next line with your code
-  console.log(arr);
+  const flattened = [];
+  for (let i = 0; i < arr.length; i++) {
+    for (let j = 0; j < arr[i].length; j++) {
+      for (let k = 0; k < arr[i][j].length; k++) {
+        flattened.push(arr[i][j][k]);
```
Meno-101: Excellent job, this ran fine which is doing the task.
But sometimes we don't only have to do the task, but to think of `what if my program have this situation`?
```javascript
// this will not be as you expected!
// but it will result with:  [ [ [ 1, 2 ], [ 3, 4 ] ], [ [ 5, 6 ], [ 7, 8 ] ] ] -> half way 😅
const arrXd = [[[[[1, 2], [3, 4]], [[5, 6], [7, 8]]]]];
```

you can scroll [here] to see how people solved this issue with many solutions, but for simplicity I would go for the first [result on google]

[here]: https://stackoverflow.com/questions/10865025/merge-flatten-an-array-of-arrays
[result on google]: https://www.google.com/search?sxsrf=ACYBGNRxR8JYJ8_YygPAGLO693S5jrVJhQ%3A1569024533993&ei=FWqFXYalPIfYwQKgk6SIBg&q=flatten+arrays+with+javascript+mdn&oq=flatten+arrays+with+javascript+mdn&gs_l=psy-ab.3..33i22i29i30.163590.164533..164837...0.2..0.88.316.4......0....1..gws-wiz.......0i71j0i22i30.qPtLYprnxbc&ved=0ahUKEwjGg_jvz-DkAhUHbFAKHaAJCWEQ4dUDCAs&uact=5

Week3/homework/step2-7.js
```diff
@@ -20,4 +20,12 @@ f2(y);
 
 console.log(y);
 
-// Add your explanation as a comment here
+// The first function incriminate variable of val 1 number.
+// Then returns the number as a variable.
+// In the console it is written 9, because x was a const variable.
+// That means x never changes.
```
Meno-101: Almost there, exactly as you said:
> The first function incriminate variable of val 1 number, Then returns the number as a variable.

And I will add from here on, that in the console it is written as 9, because what happened here that we passed a ver new value to the function as a `val` name, so the `val` parameter contains a very fresh value and
> what happened in Vegas stays in Vegas.

to explain the `const` thing, you can change it to be `let` or even `var`, that will not change it's value unless you explicitly specified:
```javascript
// assuming x declared using `var` or `let`
x = "something else"
```

then `x` is `something else`, in the `const` scenario this will throw an error, and that's the result of a constant, if you try to change it, it will complain and it will not change.

Week3/homework/step2-7.js
```diff
@@ -20,4 +20,12 @@ f2(y);
...
+
+// The second function incriminate x property of an object called val.
+// Then returns the number as an object.
+// In the console it is written 10, because y was a const object.
+// That means y doesn't change, but the content of y can change.
```
Meno-101: continuing the previous comment but this time you have hitted the spot.
> The second function incriminate x property of an object called val.
> Then returns the number as an object.
> In the console it is written 10, because y was a const object.
> That means y doesn't change, but the content of y can change.

to be more accurate, an object is just a reference to what it contains, that means the `y` is just somewhere else in our computer, and that makes it changes.

just by passing the reference to `val` that doesn't means `val` contains the exact thing as `x`, but all it contains is a reference that points to where `x` is located.

**And this is the difference between `passing by reference` and `passing by value`**
A value never change unless you change it directly by it's variable name
A reference points to it's content on the desk and once that changed somehow it changes everywhere.

Week3/homework/step3.js
```diff
@@ -1,8 +1,8 @@
 'use strict';
 
 function createBase(base) {
-  // Replace this comment and the next line with your code
-  console.log(base);
+  const number = base;
+  return x => x + number;
```
Meno-101: Excellent 💯 
But for shortness, you don't have to declare line number 4 but you only have to:
```javascript
function createBase(base) {
  return x => x + base;
}
```

Week3/homework/step3-bonus.js
```diff
@@ -3,8 +3,14 @@
 const values = ['a', 'b', 'c', 'd', 'a', 'e', 'f', 'c'];
 
 function makeUnique(arr) {
-  // Replace this comment and the next line with your code
-  console.log(arr);
+  for (let i = 0; i < arr.length; i++) {
+    for (let j = 0; j < i; j++) {
+      if (arr[i] === arr[j]) {
+        arr.splice(i, 1);
```
Meno-101: That's Smart! 🥇 

I think it's time to move to a bit more advanced level, what about reading:
https://gomakethings.com/removing-duplicates-from-an-array-with-vanilla-javascript/

It's super useful to learn new techniques, Skip what you find complicated!

Week3/homework/step2-1.js
```diff
@@ -1,9 +1,7 @@
 'use strict';
 
 function foo(func) {
-  // What to do here?
-  // Replace this comment and the next line with your code
-  console.log(func);
+  return func();
```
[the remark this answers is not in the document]
haha1906: Sorry, I still don't understand how should I make the branches. Currently it is written:
	

> haha1906 wants to merge 3 commits into `HackYourFuture:master` from `haha1906:JS2-Week-3-Homework`

So I think that I've made `JS2-Week-3-Homework` branch. And my source branch was `master` branch of Hack your future. Should I have taken Week-2 branch, instead of master? How should my commit on top must be seen?

Week3/homework/step2-4.js
```diff
@@ -1,10 +1,15 @@
...
 function Dog() {
-  // add your code here
+  this.name = 'Puppy';
+  this.color = 'brown';
+  this.numLegs = 4;
 }
 
 const hound = new Dog();
+hound.name = 'Miami';
+hound.color = 'pink';
```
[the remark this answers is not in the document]
haha1906: I understand, so intstead of writing too long code, I can make it shorter as:
```suggestion
function betterDog(name, color, numLegs) {
  this.name = name;
  this.color = color;
  this.numLegs = numLegs;
}
```
So that, I can call them in only one small function.
```suggestion
const betterHound = new betterDog('betterMiami', 'betterPink', 4);
```
This is what you said, right?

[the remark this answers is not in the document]
haha1906: Because I don't know how to make suggestion on GitHub, I marked only line of `hound.color = 'pink';`. But I wanted to make suggestions for line 3-7 and line 9-11.

Week3/homework/step2-6.js
haha1906: Now, i read the links and learned about `.isArray();` and `.flat(Infinity);`  functions. Thanks a lot

PR #495, opened 2019-09-14
NesibeInce JS2 HW3

NesibeInce:


Week3/homework/step2-1.js
```diff
@@ -1,13 +1,11 @@
 'use strict';
 
 function foo(func) {
-  // What to do here?
-  // Replace this comment and the next line with your code
-  console.log(func);
+  return `What is your name? ${func()}`;
```
Meno-101: Cool stuff!

Week3/homework/step2-2.js
```diff
@@ -2,19 +2,27 @@
...
 function threeFive(startIndex, stopIndex, threeCallback, fiveCallback) {
   const numbers = [];
-
-  // Replace this comment and the next line with your code
-  console.log(startIndex, stopIndex, threeCallback, fiveCallback, numbers);
+  for (let i = startIndex; i <= stopIndex; i++) {
+    numbers.push(i);
+  }
+  for (let i = 0; i < numbers.length; i++) {
+    if (numbers[i] % 3 === 0 && numbers[i] % 5 === 0) {
```
Meno-101: This is a correct solution, I believe it's as a 95% to 97% correct, looking back to the question it said:
> Both functions should be called if the array value is divisible by both 3 and 5.

and that's why it's correct.
the `if` block is protecting any code to be executed by the next `else if` statements,
that means we have
 - `if` it's divisible by 5 and 3 do this
 - `else if` it's divisible by 3 do that
 - `else if` it's divisible by 5 do the other

if we removed the else statements there is nothing preventing the function from executing the rest, what if we could focus on the result of executing both the functions on 3 and 5, so perhaps we only need two `if` statements in this case? 🙂

Week3/homework/step2-3.js
```diff
@@ -4,23 +4,28 @@
...
 
 // Use a 'while' loop
 function repeatStringNumTimesWithWhile(str, num) {
-  // eslint-disable-next-line prefer-const
   let result = '';
-
-  // Replace this comment and the next line with your code
-  console.log(str, num, result);
-
+  if (num <= 0) {
```
Meno-101: While this is a `while` loop and line 7 you have a `for` loop, checking for `num` if it's less than 0 would be a redundant step, the for/while loops will not loop through what you have unless the condition is correct, so:
if `i < num` and `i` is 0 the condition will never pass

the only one needs that condition is the `do while` loop because it's doing before it checks!

Week3/homework/step2-4.js
```diff
@@ -1,7 +1,9 @@
 'use strict';
 
 function Dog() {
-  // add your code here
+  this.name = 'Bull';
+  this.color = 'Brown';
+  this.numLegs = 4;
 }
 
 const hound = new Dog();
```
Meno-101: constructor functions used to be a way for creating objects in a more dynamic easy short way, yup all of these!

means if you have the **same** structure of data all time, as 
```javascript
{ name, age, favoritePlace, address1, address2, phoneNumber, mobileNumber, ...etc }
```
and they are with a different values each, imagine your self defining each of them in it's own line, as you are doing here on lines 4,5,6 , it's very short to have for a one user or two, but think of a million!

Instead of that we have constructors as:
```javascript
function User(name, age, favoritePlace, address1, address2, phoneNumber, mobileNumber) {
  this.name = name;
  this.age = age;
  this.favoritePlace = favoritePlace;
  this.address1 = address1;
  // ...etc
}

// And. we can have them all defined in a one line,
// as for now, let's make it simple and hard code them as required in the homework
const me = new user('Jawhar', 23, 'Home', 'some where' /*, ...etc */);
console.log(me) // this will give us the previous object with the specified details I wrote up above.
```

Same applies to the `Dog` Object constructor.

Week3/homework/step2-6.js
```diff
@@ -5,12 +5,25 @@ const arr3d = [[[1, 2], [3, 4]], [[5, 6], [7, 8]]];
...
+    for (let j = 0; j < arr[i].length; j++) {
+      for (let k = 0; k < arr[i][j][k]; k++) {
+        newArr.push(arr[i][j][k]);
+      }
+    }
+  }
+  return newArr;
 }
 
 console.log(flattenArray2d(arr2d)); // -> [1, 2, 3, 4, 5, 6]
```
Meno-101: Excellent job, this ran fine which is doing the task.
But sometimes we don't only have to do the task, but to think of `what if my program have this situation`?
```javascript
// this will not be as you expected!
// but it will result with:  [ [ [ 1, 2 ], [ 3, 4 ] ], [ [ 5, 6 ], [ 7, 8 ] ] ] -> half way 😅
const arrXd = [[[[[1, 2], [3, 4]], [[5, 6], [7, 8]]]]];
```

you can scroll [here] to see how people solved this issue with many solutions, but for simplicity I would go for the first [result on google]

[here]: https://stackoverflow.com/questions/10865025/merge-flatten-an-array-of-arrays
[result on google]: https://www.google.com/search?sxsrf=ACYBGNRxR8JYJ8_YygPAGLO693S5jrVJhQ%3A1569024533993&ei=FWqFXYalPIfYwQKgk6SIBg&q=flatten+arrays+with+javascript+mdn&oq=flatten+arrays+with+javascript+mdn&gs_l=psy-ab.3..33i22i29i30.163590.164533..164837...0.2..0.88.316.4......0....1..gws-wiz.......0i71j0i22i30.qPtLYprnxbc&ved=0ahUKEwjGg_jvz-DkAhUHbFAKHaAJCWEQ4dUDCAs&uact=5

Week3/homework/step2-7.js
```diff
@@ -21,3 +25,5 @@ f2(y);
 console.log(y);
 
 // Add your explanation as a comment here
+// f1(x) = 10 but console.log(x) is 9 here x is not changing val is changing
+// f2(y) = {x:10} and console.log(y) is {x:10} the change is effecting objects.
```
Meno-101: You're right, but further more explanation would be appreciated 🙂

Week3/homework/step3.js
```diff
@@ -1,8 +1,9 @@
...
 
 function createBase(base) {
-  // Replace this comment and the next line with your code
-  console.log(base);
+  return function(add) {
+    return base + add;
+  };
 }
 
 const addSix = createBase(6);
```
Meno-101: 💯

Week3/homework/step3-bonus.js
```diff
@@ -3,10 +3,11 @@
 const values = ['a', 'b', 'c', 'd', 'a', 'e', 'f', 'c'];
 
 function makeUnique(arr) {
-  // Replace this comment and the next line with your code
-  console.log(arr);
+  const uniqueArray = [...new Set(arr)];
```
Meno-101: 🎖

PR #528, opened 2020-03-01
week2 structure

gajduk:


gajduk:
@NoerGitKat  added the same structure from week3 to week2

PR #529, opened 2020-03-01
week1-homework-structure

gajduk:


gajduk:
@NoerGitKat  same structure for week1

